Issue #150, opened 2015-02-10
Incorporate GADM shapefiles and admin hierarchy for locations


nditada:
Replace current location model with GADM hierarchy and shapefiles.

nekron:
This was replaced by the location service

PR #152, opened 2015-02-18
Feature GADM locations

flbulgarelli:
Using GADM locations instead of harcoded locations in seeds, through geojson_import gem.


flbulgarelli:
Fixes  #150

PR #210, opened 2015-04-16
Feature regenerate mappings

### LINKED ISSUE #209: Create a document mapping template per device model
Instead of keeping a single monolithic template, keep a default one and one per device model.

spalladino:
Fixes #209


spalladino:
@nekron please review and merge

PR #350, opened 2015-09-23
Feature/make user model lockable

tripledub:


tripledub:
No real reason behind locking/reset strategy. The story is not so well defined, thought we could get started and get some feedback and refinements from stakeholders. I'm happy to make that change. Thanks for the link will take a look.

PR #351, opened 2015-09-23
Feature/session timeout

anthonyoleary:
Added a default timeout for session inactivity and a minimum password length of 8 characters.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,7 +3,7 @@ class User < ActiveRecord::Base
   # :timeoutable
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :trackable,
-         :validatable, :confirmable, :omniauthable,
+         :validatable, :confirmable, :omniauthable, :timeoutable,
          :lockable
 
   has_many :identities
@@ -18,6 +18,10 @@ class User < ActiveRecord::Base
 
   after_create :grant_implicit_policy
   attr_accessor :skip_implicit_policy
+  
+  def timeout_in
+    2.minutes
+  end
 
   def create(model)
     if model.respond_to?(:user=)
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -16,6 +16,10 @@ class User < ActiveRecord::Base
   has_many :granted_policies, class_name: "Policy", foreign_key: "granter_id"
 
   after_create :grant_implicit_policy
+  
+  def timeout_in
+    5.minutes
```
spalladino: A timeout of 5 min can be a bit of a hassle, especially in development mode. I'd suggest moving this value to a setting in `settings.yml`, and keeping a much larger value in `development.yml`.

nekron:
Would you mind making that change Anthony?

app/models/user.rb
anthonyoleary: Good point. I will do it.

spalladino: If you need more info on how `Settings` are computed, we are using the [rails_config](https://github.com/railsconfig/config) gem for this.

anthonyoleary:
I changed it to add different values for production and development settings. Hope it is ok now.

config/settings/development.yml
```diff
@@ -1,3 +1,4 @@
 nndd_url: "http://localhost:9000"
 location_service_url: "http://locations-stg.instedd.org"
+web_session_timeout: 120
```
nekron: Would you mind increasing this value to 3600, to simplify development and avoid having to log back in frequently?

If you want to develop and test the timeoutable behavior, you can always redefine it in a "development.local.yml" file that will be ignored by git.

cisaacs01:
Can we reactivate the timeout feature please

nditada:
It was never deactivated, it just has a long setting for development servers.

Issue #353, opened 2015-09-24
Change prefix to include resource


nekron:
Change cdxp:readLaboratory to laboratory:read

Remove cdxp prefix so resources are like:
laboratory?institution=3

spalladino:
Remove policies with resources that do not match their action (ie read lab on a device)

spalladino:
Fixed in PR #361

PR #361, opened 2015-09-25
Feature computed policies

### LINKED ISSUE #343: Precalculate permissions for each user based on policies
Recursively evaluating policies every time a permission needs to be checked is too expensive. Precalculate permissions for each user.

spalladino:
New version of policies: dropped effect and deny, and added except. 
Precompute policies via ComputedPolicy model.

Fixes #343 and #353.


app/policies/implicit.json.erb
```diff
@@ -0,0 +1,13 @@
+{
+  "statement": [
+    {
+      "action": ["institution:read"],
+      "resource": ["institution/0"]
+    },
```
nekron: Why is this read action needed?

spalladino: To satisfy the spec `allows a user to read institutions even if there are no institutions on the system`

nekron: That test existed because in order to be able to create an institution, you needed to access the index. Right now, a simple redirect is being done.

Please change the controller to avoid the forbidden and redirect the user to the institution creation screen. With that you don't need an implicit read access for a ghost institution.

spalladino: Fixed!

Issue #342, opened 2015-09-14
Sync Client remote log access


nekron:
Servers should be able to retrieve logs from sync-clients installed in lab computers to be able to diagnose issues and provide support without requiring the operator intervention.

Include 3 control methods:
- [x] If there is an exception on the client, post it to the server.
- [x] Allow the server to request the device logs. For that:
 - [x] Allow the server to let commands to the client in a specific URI (not only ask for logs)
 - [x] From the client side, frequently pull for updates (if possible, at irregular intervals)
- [x]  If there is a connectivity error, let the user "run a diagnostic" (try to get commands and latest version, and if possible some ssh request or rsync test), and "export the results" (including logs and configuration files) as a zip to send them to cdx tech support via email / pendrive / etc.

In CDX:
- [x] Show logs: allow to browse them in each device

PR #373, opened 2015-09-30
#342: sync client remote log access

asterite:
This implements the first two points of https://github.com/instedd/cdx/issues/342:

* Added these routes:
  * `POST /devices/:id/device_logs`: so that the client can submit logs there (auth is done with the device key)
  * `GET /devices/:device_id/device_logs`: to list the logs for a cdx user. Permission checked: "supportDevice"
  * `GET /devices/:device_id/device_logs/:id`: to download a log file. Logs are stored compressed in the `device_logs` table, in a binary column
  * `POST /devices/:id/request_client_logs`: in the device edition page a new link "Request client logs" is added, that ends up POSTing here. This ends up creating an entry in the `device_command` table with a name "send_logs". Once this entry is created, the "request client logs" link is replaced with a "Cilent logs requested. Please wait" text. Permission checked: "supportDevice"
  * `GET /devices/:device_id/device_commands`: returns a JSON with all queued commands for a device.  This is used by the cdx client to retrieve commands periodically.
  * `POST /devices/:device_id/device_commands/:id/reply": sends data to a command. In the case of "send_logs", the raw log data is POSTed by the cdx client. This ends up creating an entry in the `device_logs` table. The command is deleted afterwards.

Additionally, now when a cdx client is activated with an activation token, the device UUID and secret key are sent in the response. This allows the client to check commands and send logs.


asterite:
Client side: https://github.com/instedd/cdx-sync-client/pull/29

config/routes.rb
```diff
@@ -30,6 +31,12 @@
         post 'reprocess'
       end
     end
+    resources :device_logs
+    resources :device_commands do
```
nekron: You are missing a `only: [:index]` for `:device_commands` and `only: [:index, :show, :create]` for `:device_logs`.

PR #425, opened 2015-10-21
Feature/demodata

anthonyoleary:
This feature is to be able to generate demo test results from either the playground web page or to run automatically each day [run rake db:seed to generate the required user/device/model.].  This should never be used in production.  The goal is to have data for the alerts and customers to see some data. This demo data will need to be regularly updated over time with new test cases ,devices etc.

There are templates in seeds/manifests_demo_template for devices, such as  cepheid.json

{
  "event": {
    "result": "{result}",
    "assay_name": "MTB-RIF Assay G4_noQC1_2",
    "sample_ID": "{sample_id}",
    "system_user": "{system_user}",
    "device_serial_number": "123456",
    "end_time": "{end_datetime}",
    "start_time": "{start_datetime}",
    "event_id": "{event_id}",
    "test_type": "{test_type}"
  }
}

the '{...}' are values that will be replaced by random faker data, the advantage to to this tempalte approach is we do not need to fake every field so can paste in extra fields that do not need random data.  

Currently doing a wiki page to describe it, and writing some code test cases.


waj:
I like the idea of have a way to generate random demo data, but I don't think we have to mix that within the normal flow in the MessagesController.
Maybe a rake task that we execute manually from command line is good enough.

cisaacs01:
Having to manually run a task every day to generate data is not the intent. Needs to run automatically.

waj:
Use cron to execute it every day then

nekron:
I think what @waj was trying to say is that you shouldn't change the behavior in MessagesController to support this flow. This should remain as an external task or as a part of the playground. Otherwise you risk breaking the normal flow if any device accidentally triggers this behavior (https://github.com/instedd/cdx/pull/425/files#diff-7e1dadb924641a7911857a6ab553ef7aR18).

db/seeds.rb
```diff
@@ -6,3 +6,16 @@
 #   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
 #   Mayor.create(name: 'Emanuel', city: cities.first)
 
+if ENV['USE_DEMO_DATA'] || Settings.use_demo_data || Rails.env.test?
```
nekron: This shouldn't be part of the seeds, as is not needed for a regular environment to work.

The seeds shouldn't be needed for the test environment. It should be the test responsibility to create the objects that it needs in order to run.

Please extract this part of the seeds into a rake task that does the same, and into the setup part of the tests that need this data.

anthonyoleary: The reason it is part of the MessagesController and playground is that I was asked to get it working in the playground first and then to automate it. What do you think of having it in the playground?  we will have to regularly change/test the demo data as we add new data-fields/devices and the playground is a handy place to do it.

The daily demo data job is then triggered by a sidekiq cron addon ,the advantage being to use the sidekiq GUI to see if jobs run/fail.

I can maybe move the seeds/manifests_demo_template files to app/jobs/manifests_demo_template  [prefer to keep them as individual files rather than the setup of a  task/job?

nekron: - It's ok to have it in the playground, but that shouldn't imply any modifications to the MessagesController. The playground should remain isolated enough to be easily removed in the future without having to change anything else.

- I like the idea of using sidekiq. But add a rake task to allow any user to create a batch of tests on demand without having to setup the daily job. It could just trigger the same sidekiq job. Accept the number of tests to be generated as a parameter.

- I don't have any preferences regarding if you move it to an individual file or to the task setup.

anthonyoleary: It's difficult to remove the demodata flow from MessagesController if we still want the demodata functionality on the playground page [we are relying on the user being familiar with what the page is supposed to do - or maybe put in some safety such as if repeat_demo >0 and the data does not contain {..} ignore this repeat_demo parameter or similiar..], or maybe we create a new api/demoplayground web page or something [but thats maybe duplicating code/effort]. what do you think?

-just creating a rake task now as from the comments and moving the template seed data files.

nekron: I see what you mean.

In order to avoid code duplication with an additional endpoint, as you say, I'd suggest to have a different endpoint in the PlaygroundController that handles the demo data generation, and returns an array of json or xml demo data to the playground page, which will then send it back to the MessagesController one by one as usual.

If the number of tests generated is not too big, it won't be a problem. And on the other hand, if you want to generate a considerable number of tests then you should run the rake task, otherwise you will hang the server and it won't matter the way we implemented the playground generation.

Issue #371, opened 2015-09-29
Support single-tenant deployments


nditada:
We foresee the likelihood of countries wanting to have their own server in-country. A CDx server instance should be able to be configured as 'single-tenant', meaning basically that users cannot self-register. Multiple institutions can be created, but they will have to be created by the server super-admin

nekron:
We need an stg single tenant server to test this deploy.

cisaacs01:
Does single tenant deploy remove link to Google sign in on login page? If so #1005 is redundant.

mmuller:
Now users can self-register (version 0.11-pre15)

nekron:
Single tenant deploy does not remove the link to the google sign in on the login page. #1005 should be implemented that way.

nekron:
Splitted into #1038.

PR #426, opened 2015-10-21
**REVIEW ONLY - NOT READY TO MERGE ** Feature/user requests access

tripledub:
So my thinking is when a prospective user 'Prospect' requests access to the system they get stored as a prospect. An admin can view the prospects and if approved a user invite is sent out for the prospect to accept when they get redirected to set their password etc.

I've inlined some comments on areas I need clarification on which is mainly around setting of policies.

### DIFF
--- a/app/controllers/prospects_controller.rb
+++ b/app/controllers/prospects_controller.rb
@@ -10,14 +10,6 @@ class ProspectsController < ApplicationController
     end
   end
 
-  def index
-    @prospects = Prospect.pending
-  end
-
-  def new
-    @prospect = UserRequest.new
-  end
-
   def create
     @prospect = UserRequest.new(prospect_params)
     if @prospect.valid?
@@ -28,7 +20,26 @@ class ProspectsController < ApplicationController
     end
   end
 
+  def index
+    @prospects = Prospect.pending
+  end
+
+  def new
+    @prospect = UserRequest.new
+  end
+
+  def reject
+    Prospect.where(uuid: params[:id]).tap do |prospects|
+      deny(prospects.first) if prospects.any?
+      redirect_to prospects_path
+    end
+  end
+
   private
+
+  def deny(prospect)
+    prospect.update_attribute(:uuid, nil)
+  end
 
   def invite(prospect)
     User.invite!(email: prospect.email).tap do |user|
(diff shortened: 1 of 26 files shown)


app/controllers/prospects_controller.rb
```diff
@@ -0,0 +1,65 @@
+class ProspectsController < ApplicationController
+  skip_before_action :authenticate_user!
```
tripledub: This feels like a hack, I added these lines so I could progress and make prospects page. I need to know more about policies.

app/controllers/prospects_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  private
+
+  def deny(prospect)
+    prospect.update_attribute(:uuid, nil)
+  end
+
+  def invite(prospect)
+    User.invite!(email: prospect.email).tap do |user|
+      prospect.update_attribute(:uuid, nil)
```
tripledub: So after this line I need to create a policy to add the invited user to an institution. I guess I'll need to change the UI when I figure out how this works, but some help here for now would be appreciated.

app/controllers/prospects_controller.rb
nekron: First things first: Any user should be able to create an account in cdx.
You are seeing cdx as a closed product where no one can create an account without invitation or proper authorization, but for most of the cases that's not true.

Having said that, there is an ongoing development related to this: https://github.com/instedd/cdx/issues/371 and https://github.com/instedd/cdx/tree/feature/371-single_tenant_deployment

The idea behind this is that in a single tenant deploy the user would be able to create an account but he won't be able to do anything else. Then an administrator would then give him access through a policy and from there he'll be able to access the application.

The "request access" feature would then be a part of the new institution screen, with an autocomplete in the name, the user will be able to send a request to the admin of that institution. If there is no institution left to send requests, an empty page will be displayed saying that his request is waiting for approval.

The invite screen will be a part of the create policy screen, from there you can choose an existing user to assign the policy. The idea is that besides choosing an existing user, you have a text box where you can type an email and the invite to that user will be sent.

cisaacs01: This is not true. THIS VERSION of the application will mostly be deployed as a closed product and user invites and administration will be managed similar to an enterprise model. Users may be able to request access from a central admin but otherwise the admin will be sending out invites to users individually and in groups. Please mail me for further clarification. Dont forget this is a FIND version of the application and we will get what we want and not what has previously been anticipated.

nekron: That's ok. My goal is that the closed version coexists with the open version in a single code base.

PR #449, opened 2015-10-28
Fixed #339: support OAuth

### LINKED ISSUE #339: Support oauth
- [ ] Allow queries using oauth
- [ ] Allow subscribers creation using oauth
- [ ] Update hub to use oauth to connect to cdx.
- [ ] Remove the broken "Request API Token" link from the institutions screen.
- [ ] Allow the user to request an API bearer token from his settings.

asterite:
* Added [doorkeeper gem](https://github.com/doorkeeper-gem/doorkeeper#other-customizations) which adds support for OAuth2. Very little is used for this, mainly the apps and access tokens generation logic, and the authentication logic.
* In ApiController use `current_user` if available (which also supports basic auth) and if not tries to use the OAuth token (which can be provided in a header or query string and is handled by doorkeeper)
* Added an "API tokens" menu option in the right UI navbar which lets you manage your API tokens. For now you can click "+" and it creates a token that never expires. From that same list you can delete that token. In the future we might want to add an expiration date and maybe add a name to each token, or other options.
* Removed the "Request API token" link and related logic


asterite:
Once this is merged I'll change hub to support oauth authentication for cdx subscribers.

PR #451, opened 2015-10-28
Feature/371 single tenant deployment

### LINKED ISSUE #371: Support single-tenant deployments

nekron:
Fixes #371.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -72,12 +72,12 @@ class ApplicationController < ActionController::Base
   # filters/authorize @institutions by action. Assign calls resource.institution= if only one institution was left
   def prepare_for_institution_and_authorize(resource, action)
     @institutions = authorize_resource(@institutions, action)
-    if @institutions.one?
-      resource.institution = @institutions.first
-      @institutions
-    elsif @institutions.blank?
+    if @institutions.blank?
       head :forbidden
       nil
+    elsif @institutions.one?
+      resource.institution = @institutions.first
+      @institutions
     else
       @institutions
     end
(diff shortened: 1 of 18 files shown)


app/models/user.rb
```diff
@@ -30,9 +30,15 @@ def create(model)
   end
 
   def implicit_policies
-    self.institutions.pluck(:id, :kind).map do |institution_id, kind|
+    policies = self.institutions.pluck(:id, :kind).map do |institution_id, kind|
       Policy.owner(self, institution_id, kind)
-    end + [Policy.implicit(self)].compact
```
spalladino: This change can be implemented in this block:
```ruby
self.institutions.pluck(:id, :kind).map do |institution_id, kind|
  Policy.owner(self, institution_id, kind)
end + [Policy.implicit(self) unless Settings.single_tenant].compact
```

app/controllers/application_controller.rb
```diff
@@ -61,19 +54,23 @@ def authorize_resource(resource, action)
   def check_no_institution!
     return if current_user && current_user.need_change_password?
     if current_user && current_user.institutions.empty? && current_user.policies.empty?
-      redirect_to new_institution_path
+      if has_access?(Institution, CREATE_INSTITUTION)
+        redirect_to new_institution_path
+      else
+        @hide_nav_bar = true
```
spalladino: Should the user be redirected somewhere else in this case?

nekron: Eventually, when the user has the ability to request access to an institution he will be redirected to that page.

nekron: I'll create a "contact your administrator" page for now.

PR #452, opened 2015-10-28
fixes #441 for release/0.6 by adding a validation in the model

### LINKED ISSUE #441: Devices should not be associated to Sites from another Institution
Found in version: 
0.6-pre3

Prerequisites:
There are two Institutions ('Foo', and 'Bar')
The Site 'Foo' belongs to the Institution 'Foo'

Steps:
Go to create a new Device 
Fill 'Name' with 'Bar Device'
Select the Institution 'Bar'
Select the option 'Site Foo' from the Sites listing
Fill the form and save it

Obtained:
The new Device ('Bar Device') is associated to a Site (Site 'Foo') from another Institution ('Foo')

Expected:
Devices should not be associated to Sites from another Institution

bcardiff:
fixes #441 for release/0.6 by adding a validation in the model that is forwarded to the ui

### DIFF
--- a/app/models/device.rb
+++ b/app/models/device.rb
@@ -26,6 +26,7 @@ class Device < ActiveRecord::Base
   validates_presence_of :device_model
 
   validate :unpublished_device_model_from_institution
+  validate :same_institution_of_site
 
   before_create :set_uuid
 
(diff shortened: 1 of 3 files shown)


bcardiff:
@nekron after merging to release/0.6 it should be also merged to master.
Then I will implement the feature to update the list of sites based on the institution.

PR #457, opened 2015-10-29
Feature/admin panel superuser

anthonyoleary:
Added the mange user policy.

### DIFF
--- a/app/assets/stylesheets/active_admin.scss
+++ b/app/assets/stylesheets/active_admin.scss
@@ -1,0 +1,51 @@
+// SASS variable overrides must be declared before loading up Active Admin's styles.
+//
+// To view the variables that Active Admin provides, take a look at
+// `app/assets/stylesheets/active_admin/mixins/_variables.css.scss` in the
+// Active Admin source.
+//
+// For example, to change the sidebar width:
+// $sidebar-width: 242px;
+
+$primary-color: #000000;
+$table-stripe-color: rgba(0,0,0,.4);
+
+h1 { font-size: 14px !Important; line-height: 1; margin-bottom: 0.5em;  }
+.header-item { font-size: 14px !Important;}
+.title_bar { color: #ffffff !Important; background-color: #ffffff !Important;}
+.sidebar { color: #ffffff !Important; background-color: #ffffff !Important;}
+
+.sidebar_section { color: #ffffff !Important; background-color: #ffffff !Important;}
+
+.even { color: #ffffff !Important; background-color: #ffffff !Important;}
+
+.yes {  background-color: #808080 !Important;}
+
+.col {  background-color: #ffffff !Important;}
+
+h3 {  background-color: #ffffff !Important;}
+
+input[type="submit"]{ background: #ff0000 !Important; } 
+
+select {
+  background: #ffffff;
+  color: #000;
+}
+
+.panel_contents {  background-color: #ffffff !Important;}
+
+
+
+
+
+// Active Admin's got SASS!
+@import "active_admin/mixins";
+@import "active_admin/base";
+
+// Overriding any non-variable SASS must be done after the fact.
+// For example, to change the default status-tag color:
+//
+//   .status_tag { background: #6090DB; }
+
+
+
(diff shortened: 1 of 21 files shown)


app/controllers/application_controller.rb
```diff
@@ -53,7 +66,7 @@ def authorize_resource(resource, action)
 
   def check_no_institution!
     return if current_user && current_user.need_change_password?
-    if current_user && current_user.institutions.empty? && current_user.policies.empty?
+    if current_user && current_user.institutions.empty? && current_user.policies.empty? && !is_super_admin?
```
nekron: This check is redundant: In order to be a "super admin" the user needs to have a policy, so the previous check would fail and this last part will never be executed.

app/controllers/application_controller.rb
```diff
@@ -62,6 +75,14 @@ def check_no_institution!
     end
   end
 
+  def after_sign_in_path_for(resource)
```
nekron: Please do not change the after sign in path for admins, let them access the administration menu by adding it to the top level navigation.

app/controllers/application_controller.rb
```diff
@@ -19,6 +19,19 @@ class ApplicationController < ActionController::Base
...
+  def is_super_admin?  
+    if current_user && !current_user.policies.empty? && Policy.can?(MANAGE_USER, User, current_user, @current_user_policies)
+      true
+    else
+      false
+    end
+  end
+  
+  def authenticate_admin_user!
+    redirect_to new_user_session_path unless is_super_admin?
```
nekron: If a non-admin user tries to access the admin dashboard it gets a redirect loop instead of an "unauthorized" message.

app/assets/stylesheets/active_admin.scss
```diff
@@ -0,0 +1,51 @@
...
+
+// Active Admin's got SASS!
+@import "active_admin/mixins";
+@import "active_admin/base";
+
+// Overriding any non-variable SASS must be done after the fact.
+// For example, to change the default status-tag color:
+//
+//   .status_tag { background: #6090DB; }
+
```
nekron: Having added active admin forces you to keep the style of the application in sync with the style of activeadmin's dasbhard. And it doesn't quite seems as part of the same application.

db/schema.rb
```diff
@@ -362,10 +395,19 @@
     t.datetime "locked_at"
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.boolean  "isActive",                                   default: true
```
nekron: You have the columns duplicated, maybe because you renamed them in the migration without first rolling back.

db/schema.rb
```diff
@@ -362,10 +395,19 @@
...
+    t.boolean  "isArchived"
+    t.boolean  "isAdmin",                                    default: false
+    t.boolean  "is_admin",                                   default: false
```
nekron: You still have the is_admin boolean in the users. This is not needed as you already have the MANAGE_USER policy.

db/schema.rb
```diff
@@ -32,6 +32,39 @@
...
+    t.string   "author_type",   limit: 255
+    t.datetime "created_at"
+    t.datetime "updated_at"
+  end
+
+  add_index "active_admin_comments", ["author_type", "author_id"], name: "index_active_admin_comments_on_author_type_and_author_id", using: :btree
+  add_index "active_admin_comments", ["namespace"], name: "index_active_admin_comments_on_namespace", using: :btree
+  add_index "active_admin_comments", ["resource_type", "resource_id"], name: "index_active_admin_comments_on_resource_type_and_resource_id", using: :btree
+
+  create_table "admin_users", force: :cascade do |t|
```
nekron: Why is this `admin_users` table needed?

db/schema.rb
```diff
@@ -32,6 +32,39 @@
 
   add_index "activations", ["activation_token_id"], name: "index_activations_on_activation_token_id", unique: true, using: :btree
 
+  create_table "active_admin_comments", force: :cascade do |t|
```
nekron: If the only thing that can be administered is users, and the only thing that can be made on them is disabling their accounts, is it really necessary to have an extra table just to store a comment about why it was disabled. It seems that this could be solved with a "disabled_reason" string variable in the user model.

Issue #543, opened 2015-11-23
Allow the user to upload a pdf with instructions for a device


nekron:

pmallol:
After trying to upload a .pdf with instructions the user is redirected to a blank page reading:
`An unhandled lowlevel error occurred. The application logs may have details.`

spalladino:
@waj should we configure paperclip to [store all attachments in S3](https://github.com/thoughtbot/paperclip/wiki/Paperclip-with-Amazon-S3)? The same may apply as well to device model images.

pmallol:
Ok in Version: 0.7-pre10

PR #547, opened 2015-11-23
#543 allow manufacturer to upload pdf instruction file.

bcardiff:
allow user to download pdf instruction file from setup instructions
shrink model used by DeviceSetup react component

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151119133838) do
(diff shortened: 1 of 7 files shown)


db/schema.rb
```diff
@@ -406,6 +410,7 @@
     t.integer  "invited_by_id",                  limit: 4
     t.string   "invited_by_type",                limit: 255
     t.integer  "invitations_count",              limit: 4,   default: 0
+    t.string   "last_navigation_context",        limit: 255
```
nekron: It seems that this came from the site navigation branch.

PR #589, opened 2015-12-01
Feature/587 scope sites with navigation context

### LINKED ISSUE #569: When creating a new site, automatically select the current scope as the parent site.
Let the user change it, but only to the child nodes of the current scope.

### LINKED ISSUE #587: Scope sites with navigation context

bcardiff:
fixes #587 
fixes #569 

There is a slightly duplicate logic between devices and sites since both are filterable within the hierarchy, but using different field names.

### DIFF
--- a/app/views/sites/index.html.haml
+++ b/app/views/sites/index.html.haml
@@ -5,11 +5,9 @@
       - t.thead do
         %tr
           %th Name
-          %th Institution
           %th Location
       - t.tbody do
         - @sites.each do |site|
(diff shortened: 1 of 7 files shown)


app/views/sites/index.html.haml
```diff
@@ -5,13 +5,11 @@
       - t.thead do
         %tr
           %th Name
-          - unless @institutions.one?
-            %th Institution
+          %th Institution
```
nekron: It shouldn't be necessary to display the institution at all.

PR #618, opened 2015-12-04
Feature/admin blocks user

tripledub:


app/controllers/settings_controller.rb
```diff
@@ -0,0 +1,27 @@
+class SettingsController < ApplicationController
+  before_filter :load_locales
+
```
spalladino: I'd add a `skip_before_filter :ensure_context` here, since settings are independent from navigation

app/controllers/users_controller.rb
```diff
@@ -1,29 +1,24 @@
 class UsersController < ApplicationController
-  LOCALES = [
-              ["English", "en"],
-            ]
-
-  def settings
-    load_locales
+  def edit
```
spalladino: Add a check to ensure that the current user can actually edit this user. We might need a new `user` resource and its corresponding `user:update` action. Same applies to the `update` action in this controller.

app/controllers/users_controller.rb
```diff
@@ -1,29 +1,24 @@
...
-    if current_user.update(params)
-      redirect_to settings_path, notice: "Settings updated"
-    else
-      load_locales
-      render :settings
+  def update
+    @user = User.find(params[:id])
+
+    if can_edit? && @user.update(admin_update_params)
+      @_message = "User updated"
```
spalladino: Why do you need to store this in an instance variable?

app/controllers/users_controller.rb
```diff
@@ -1,29 +1,24 @@
...
-  def user_params
-    params.require(:user).permit(:locale, :time_zone, :timestamps_in_device_time_zone)
+  def admin_update_params
+    params.require(:user).permit(:is_active)
   end
 
-  def load_locales
-    @locales = LOCALES
+  def can_edit?
+    current_user.id != params[:id].to_i
```
spalladino: I'd check on the already loaded user rather than the params, in case we start using a different parameter (uuid?) in the future. So you can do `current_user != @user` directly.

app/views/users/_form.html.haml
```diff
@@ -0,0 +1,35 @@
...
+        .col.pe-2
+          = f.label :is_active, "Status"
+        .col
+          = f.check_box :is_active
+          = f.label :is_active, "&nbsp; Can access Cdx?".html_safe
+      .row
+        .col.pe-2
+          = f.label :last_sign_in_at, "Activity"
+        .col
+          = last_activity(f.object)
```
spalladino: Add the `.value` class here as well

config/routes.rb
```diff
@@ -128,9 +128,10 @@
   end
 
   scope :user do
-    get 'settings' => "users#settings"
-    patch 'settings' => "users#update_settings"
+    get 'settings' => "settings#edit"
+    patch 'settings' => "settings#update"
   end
```
spalladino: A shorter alternative is to use a [singular resource](http://guides.rubyonrails.org/routing.html#singular-resources) here, with only the edit and update actions

features/login.feature
```diff
@@ -15,3 +15,10 @@ Feature: Logging into system
     Then he should see "You have one more attempt before your account is locked"
     When he attempts to log-in 3 times with incorrect password
     Then he should see "Your account is locked"
+
+  Scenario: Admin disables account
+    Given the user 'foouser@example.com' has an account
+    When his account has been disabled
+    And he attempts to log-in with correct details
+    Then he should see "Access to your account has been suspended, please refer to your supervisor"
+
```
spalladino: Though I still haven't made up my mind regarding Cucumber, I have to say that this spec looks fantastic :-)

app/controllers/users_controller.rb
tripledub: Actually that was a topic I wanted to bring up at stand-ups before we got distracted with staging site. I was looking to drop something into Policy model but there was a few things I thought it was worth talking about first.

features/login.feature
tripledub: Thanks for the feedback on Cucumber. Believe it or not I'm not 100% sure about it either ^_^. I am however a test-first zealot and when I'm adding functionality via 3rd party gem this seemed perfect way to at least test something.

In all my years in ruby I've been lucky enough to work in some pretty diverse teams. I spent the first 4 years 100% anti-cucumber. Last year I got to work in a place where the QA team was very technical - they were all 'developers in test'. They convinced me the  [Testing Triangle](https://www.mountaingoatsoftware.com/blog/the-forgotten-layer-of-the-test-automation-pyramid) is the way to work effectively.

My preferred workflow is to write a simple feature test that touches all components. I'm guilty of not writing enough 'service' tests (RSpec features in this case) but I do make up with good unit tests I hope. I think the main objection to cucumber is the work required to set up tests but now I'm more experienced I don't think this has to be the case.

app/controllers/users_controller.rb
```diff
@@ -1,29 +1,30 @@
...
+
+  def can_edit?
+    current_user.id != @user.try(:id)
   end
 
-  def load_locales
-    @locales = LOCALES
+  def load_resource
+    @user ||= User.find(params[:id])
+    @_message = 'Unable to load user' unless @user
```
spalladino: `find` raises an `ActiveRecord::RecordNotFound` error, which is automatically handled and rendered as a 404, so there is no need to handle the `@user.nil?` case here

app/controllers/users_controller.rb
```diff
@@ -1,29 +1,30 @@
...
-    if current_user.update(params)
-      redirect_to settings_path, notice: "Settings updated"
-    else
-      load_locales
-      render :settings
+  def update
+    return unless authorize_resource(@user, UPDATE_USER)
+    if can_edit? && @user.update(user_params)
+      @_message = 'User updated'
     end
```
spalladino: We should split these cases:
- If cannot edit, then `head :json` and return, like in [authorize_resource](https://github.com/instedd/cdx/blob/feature/568-map_site/app/controllers/application_controller.rb#L44)
- If cannot update, due to a validation error, then `render edit` with the current model
- Only if it was successful redirect to `edit_user_path`

Note that with these changes there is no need for `@_message` to be an instance var

app/models/user.rb
```diff
@@ -79,4 +81,20 @@ def create_api_token
...
+    self.is_active? ? super : I18n.t('devise.failure.suspended')
+  end
+
+  scope :within, -> (institution_or_site) {
+    if institution_or_site.is_a?(Institution)
+      where(institution: institution_or_site)
+    else
+      where("prefix LIKE concat(?, '%')", institution_or_site.prefix)
+    end
+  }
```
spalladino: This scope is failing for both institutions and sites. Note that user does not directly have an institution or prefix fields, this should probably be resolved via the `role_users` relationship, checking the associated site or institution of all the related roles.

app/views/users/_form.html.haml
```diff
@@ -0,0 +1,35 @@
...
+        .col.pe-2
+          = f.label :email
+        .col
+          .value= f.object.email
+    .col.pe-6
+      .row
+        .col.pe-2
+          .value= f.label :is_active, "Status"
+        .col
+          = f.check_box :is_active
```
spalladino: We should not allow users without `user:update` rights to toggle the `is_active` check, they should just see an unmodifiable check here. We have two options:
1- Split this view into an edit and a show view, and when the current user accesses a user, redirect him/her to edit or show based on permissions (see device_models for an example)
2- Use this view for both edit and show, but set the checkbox to be not-editable and hide the Save button.

The former is the cleanest option, while the latter is quicker to implement now

app/controllers/sites_controller.rb
```diff
@@ -106,6 +106,7 @@ def users
     return unless authorize_resource(@site, READ_SITE)
 
     @users = ComputedPolicy.authorized_users(Policy::Actions::READ_SITE, @site)
```
spalladino: We need to define which users we'll see here. We can either use this criteria, or list all users who belong to a role for this site. @nekron ?

app/controllers/sites_controller.rb
```diff
@@ -106,6 +106,7 @@ def users
...
+    check_access(User.within(@navigation_context.entity), READ_SITE)
```
spalladino: `check_access` will return the list of resources for which the current user can perform the chosen action. So this should be something along the lines of:
```ruby
@users = check_access(@users, READ_USER)
```

Which will keep in `@users` all the users that can be READ by the currently logged in user.

app/controllers/sites_controller.rb
nekron: This was an initial approach that was only necessary because we didn't had roles at that time. We should change this to the list of users who belongs to a role for this site.

app/controllers/users_controller.rb
tripledub: Fair comment, I guess what I was thinking was I could explicitly communicate the user was not found

app/views/users/_form.html.haml
tripledub: If we did that then maybe could remove `can_edit?` from controller which I don't really like anyway?

spalladino: I guess we could; still, we'll need to ensure that a user doesn't block him/herself

PR #653, opened 2015-12-21
Feature/645 admin blocks user

### LINKED ISSUE #645: Admin should be able to block a user
Permission for this is granted only to superadmins. Refer to https://github.com/instedd/cdx/pull/618.

spalladino:
Allow an admin to block a user. Fixes #645.

- Merged https://github.com/finddx/cdx/tree/feature/admin_blocks_user branch from @tripledub
- Removed users view from sites show page
- Enforce site:readUsers and institution:readUsers permissions when listing users
- Changed devise registrations controller to act as settings controller and removed duplicated functionality in users and settings controllers


app/models/policy.rb
```diff
@@ -63,6 +65,8 @@ module Actions
     DELETE_ROLE = "role:delete"
     ASSIGN_USER_ROLE = "role:assignUser"
     REMOVE_USER_ROLE = "role:removeUser"
+
+    UPDATE_USER = 'user:edit'
```
nekron: Can we change this to "user:update"? Just for consistency with all the other actions.

PR #688, opened 2016-01-07
Create patients

### LINKED ISSUE #601: Allow the user to register a new patient
Include:
- Name
- Birthday
- Location
- Address
- Email
- Phone

Country and State are not PII.

All fields should be stored within the core fields.

Location should behave like site location.

### LINKED ISSUE #608: [hiereorgs] Scope patients crud

bcardiff:
Note: Created to start an early review cc: @nekron @spalladino 

* fixes #601 adds patients related permissions
* fixes #608 using only the institution of the navigation context
* DRY validation_errors
* DRY pagination controls
* Use current user locale in requests
* basic patient listing, just paging.

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -1,0 +1,36 @@
+class PatientsController < ApplicationController
+  def index
+    @can_create = true # TODO change permission
+    # @patients = check_access(Patient.where(institution: @navigation_context.institution).order(updated_at: :desc), READ_INSTITUTION) # TODO change permission
+    @patients = Patient.where(institution: @navigation_context.institution).order(updated_at: :desc)
+  end
+
+  def new
+    @patient = PatientForm.new
+    prepare_for_institution_and_authorize(@patient, READ_INSTITUTION) # TODO change permission
+  end
+
+  def edit
+  end
+
+  def create
+    @institution = @navigation_context.institution
+    return unless authorize_resource(@institution, READ_INSTITUTION) # TODO change permission
+
+    @patient = PatientForm.new(patient_params)
+    @patient.institution = @institution
+
+
+    if @patient.save
+      redirect_to patients_path, notice: 'Patient was successfully created.'
+    else
+      render action: 'new'
+    end
+  end
+
+  private
+
+  def patient_params
+    params.require(:patient).permit(:name, :dob_text, :lat, :lng, :location_geoid, :address, :email, :phone)
+  end
+end
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -2,6 +2,11 @@ class PatientsController < ApplicationController
   def index
     @can_create = has_access?(@navigation_context.institution, CREATE_INSTITUTION_PATIENT)
     @patients = check_access(Patient.where(institution: @navigation_context.institution), READ_PATIENT).order(updated_at: :desc)
+
+    @page_size = (params["page_size"] || 10).to_i
+    @page = (params["page"] || 1).to_i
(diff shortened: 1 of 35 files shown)


app/controllers/patients_controller.rb
```diff
@@ -0,0 +1,66 @@
+class PatientsController < ApplicationController
+  def index
+    @can_create = has_access?(@navigation_context.institution, CREATE_INSTITUTION_PATIENT)
+    @patients = check_access(Patient.where(institution: @navigation_context.institution), READ_PATIENT).order(updated_at: :desc)
+  end
```
spalladino: If patients don't have a site, then the only way for a user to browse patients is to have permissions over the whole institution. Should we add something more related to site?

app/models/patient.rb
```diff
@@ -22,4 +23,37 @@ def has_entity_id?
...
+          else
+            self.core_fields[arg.to_s] = value
+          end
+        end
+      end
+    end
+  end
+
+  attribute_field :name, :gender, :dob, :email, :phone
+
```
spalladino: Looks nice. Should we move this to Entity?

app/models/patient_form.rb
```diff
@@ -0,0 +1,104 @@
...
+    save
+  end
+
+  def save
+    self.class.assign_attributes(patient, self)
+    self.valid? && patient.save
+  end
+
+  validates_presence_of :name
+  GENDER_VALUES = %w(male female other)
```
spalladino: `Cdx::Fields.entities.core_field_scopes.find{|s| s.name == 'patient'}.fields.find{|f| f.name == 'gender'}.options`

app/models/patient_form.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  def dob_text_placeholder
+    date_format[:placeholder]
+  end
+
+  def dob_is_a_date
+    return if dob.blank?
+    errors.add(:dob_text, "should be a date in #{dob_text_placeholder}") unless dob.is_a?(Time)
+  end
+  # end dob
```
spalladino: What about keeping a single field "dob" which is typecasted to datetime when read, like:
```ruby
def available_at
  return super.to_datetime if super.respond_to?(:to_datetime)
  DateTime.parse(super)
end
```

Then validating the format client-side, with a `pattern` attribute, and just the presence here?

app/views/device_models/_form.haml
```diff
@@ -10,13 +10,7 @@
...
   });
 
-- if @device_model.errors.present?
-  .alert-container
-    .alert.alert-error
-      %ul
-        - if @device_model.errors.present?
-          - @device_model.errors.full_messages.each do |msg|
-            %li= msg
+= validation_errors @device_model
```
spalladino: :smiley:

app/models/patient.rb
bcardiff: We can, yet is only used in Patient at the moment. Also we need to add a way to copy a field to an AR attribute.  Maybe a `copy: true` as an options. That will pollute a bit the above code most likely.

app/models/patient_form.rb
bcardiff: `Patient.entity_fields.detect { |f| f.name == 'gender' }.options`, but yeah +1

app/models/patient_form.rb
bcardiff: :-1: regarding "just the presence". The client-side pattern is not enough.

But I could try to keep it just in a single property. I make the two due to previous intent of wrapping the formatting in patient model directly with alias_method_chain and the getter/setter property just got to ugly.

Related technical debt: dob is stored/read as text from Entity.

PR #749, opened 2016-02-05
Feature/506 move site

### LINKED ISSUE #506: [hierorgs] Move site to another parent site
As a user with site management permission, I want to move an existing site from one parent site to another.

The user has to have site management permissions for both parent sites in order to do that.

bcardiff:
fixes #506 by: 
* creating a new site
* soft deleting the original
* moving the device
* removing original sites roles
* leaving test results where they are

Pending

- [x] when sending the same message (or an update an update to a message) a device error log is created with a "Validation failed: Site can't be blank” message.
The test_results when processing the messages should be scoped by the current site of the device.
- [x] when accessing a test result from an old parent site, a policy.can? is failing, probably because the site is not been found. Have in mind that the old roles are removed from the system. So, the test should be accessible due to a more general policy
  - [x] if the policy applies to other than the device, the test result should be displayed
  - [ ] if the policy applies to the device (in the new site), the test result should NOT be displayed. this requires changes in the computed policy module. (a pending test is left in `$/spec/controllers/test_results_controller_spec.rb:252`)
- device status graphs and logs are not scoped by the current site, and then the whole history is displayed despite the messages and messages_logs do not belongs to the new site. **this also happens currently when moving device from device to device**
 - [x] device status graphs (depends on test results)
 - [x] messages and ~logs~ (depends on messages and ~device_logs~). Note: only device_messages are shown in the UI. device_logs where not changed.

### DIFF
--- a/app/controllers/device_messages_controller.rb
+++ b/app/controllers/device_messages_controller.rb
@@ -4,6 +4,7 @@ class DeviceMessagesController < ApplicationController
   def index
     device_ids = check_access(Device, SUPPORT_DEVICE).pluck(:id)
     @messages = DeviceMessage.where("device_id IN (?)", device_ids).joins(device: :device_model)
+      .where('devices.site_id = device_messages.site_id')
     apply_filters
 
     @date_options = date_options_for_filter
--- a/app/controllers/devices_controller.rb
+++ b/app/controllers/devices_controller.rb
@@ -346,7 +346,10 @@ class DevicesController < ApplicationController
     {
       "since" => (Date.today - 1.year).iso8601,
       "device.uuid" => @device.uuid
-    }
+    }.tap do |h|
+      # display only test results of the current site of the device
+      h["site.uuid"] = @device.site.uuid if @device.site
+    end
   end
 
 end
--- a/app/controllers/sites_controller.rb
+++ b/app/controllers/sites_controller.rb
@@ -72,6 +72,7 @@ class SitesController < ApplicationController
   def update
     @site = Site.find params[:id]
     institution = @site.institution
+    redirect_options = {}
 
     return unless authorize_resource(@site, UPDATE_SITE)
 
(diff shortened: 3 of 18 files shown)


bcardiff:
@spalladino I think the last checkbox is yours

If the resource to authorize is a TestResult (SiteContained maybe (?) ) then only the policies scoped to the device of the site of the test result should apply.

Right now, the user does not have a way to discover the test_result id/uuid to gain access to the show action of that test, but still, has the (wrong) rights to access to it.

spalladino:
@nekron after discussing with @bcardiff, we actually reached the conclusion that a policy on a *device* should allow the user to see all tests from that device, regardless of in which site the device is. The offending test was [marked as pending](https://github.com/instedd/cdx/pull/749/files#diff-0be372cd2ebc8908cfac400466f35326R252) in the PR, and it was changed to match this behaviour in  #755.

bcardiff:
Despite that the permision per device is kept, we agree that the device
status page and device logs should show only information based on current
site, right? Otherwise we still need to work on this.

Maybe we can add a link about older status/logs information if multiple
site information is present. But that should handle the current user
permissions.
On jue, 11 feb 2016 at 17:45 Santiago Palladino <notifications@github.com>
wrote:

> @nekron <https://github.com/nekron> after discussing with @bcardiff
> <https://github.com/bcardiff>, we actually reached the conclusion that a
> policy on a *device* should allow the user to see all tests from that
> device, regardless of in which site the device is. The offending test was marked
> as pending
> <https://github.com/instedd/cdx/pull/749/files#diff-0be372cd2ebc8908cfac400466f35326R252>
> in the PR, and it was changed to match this behaviour in #755
> <https://github.com/instedd/cdx/pull/755>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/instedd/cdx/pull/749#issuecomment-183054193>.
>
-- 
Brian J. Cardiff

Issue #876, opened 2016-02-22
Specify minimum allowed characters for AD-HOC RECIPIENT fields in Alerts


pmallol:
As for Version: 0.10-pre4, an error will prevent saving an "Ad-hoc recipient" in an Alert Group if the field 'first name' from AD-HOC RECIPIENT doesn't contain a minimum amount of characters.

It would be useful if a description appear under the field indicating the minimum amount of characters needed to validate the field.

![screen shot 2016-02-22 at 3 57 07 pm](https://cloud.githubusercontent.com/assets/9214460/13230564/9ff51dee-d984-11e5-8f08-9a7a556c1296.png)

anthonyoleary:
Anthony: I will look at this.

anthonyoleary:
if no first name or last name is entered for adhoc userss , a placeholder displays enters a min. length.

mmuller:
Ok in version 0.11-pre15

PR #1031, opened 2016-04-14
if no first name or last name is entered for adhoc userss , a placeho…

anthonyoleary:
…lder displays enters a min. length


nekron:
This fixes #876

PR #1088, opened 2021-04-30
Add LaboratorySample Create

ftarulla:
Depends on https://github.com/instedd/cdx/pull/1087

### DIFF
--- a/app/controllers/laboratory_samples_controller.rb
+++ b/app/controllers/laboratory_samples_controller.rb
@@ -9,4 +9,26 @@ class LaboratorySamplesController < ApplicationController
     # paginate samples
     @samples = perform_pagination(@samples)
   end
+
+  def new
+    @sample = LaboratorySample.new
+  end
+
+  def create
+
+    @sample = LaboratorySample.new({:uuid => uuid_param})
+
+    if @sample.save
+      next_url = laboratory_samples_path(@sample)
+      redirect_to next_url, notice: 'Sample was successfully created.'
+    else
+      render action: 'new'
+    end
+  end
+
+  private
+
+  def uuid_param
+    params.require(:uuid)
+  end
 end
--- a/app/controllers/laboratory_samples_controller.rb
+++ b/app/controllers/laboratory_samples_controller.rb
@@ -34,10 +34,4 @@ class LaboratorySamplesController < ApplicationController
       render action: 'new'
     end
   end
-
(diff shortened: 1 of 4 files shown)


app/controllers/laboratory_samples_controller.rb
```diff
@@ -9,4 +9,26 @@ def index
...
     @samples = perform_pagination(@samples)
   end
+
+  def new
+    @sample = LaboratorySample.new
+  end
+
+  def create
+
+    @sample = LaboratorySample.new({:uuid => uuid_param})
```
matiasgarciaisaia, line 19: Never trust user input!

If the UUID can't be created by users (ie, the system will auto-generate them, as in this case), let's not read them from a user parameter.

In this case, a user can edit the DOM of the `/new` page to change the system-defined `uuid` to a custom one, and the app will accept that new string as a UUID - which will probably break many assumptions through the app.

Since the user shouldn't be able to edit the UUID, let's not allow them to set it. Either `save` the sample as soon as the user hits the "New sample" button, or, if we want to ask the user for extra info, let's ask that _without showing_ the UUID, and redirect users to the `show` page after creating the new Sample.

app/views/laboratory_samples/_form.html.haml
```diff
@@ -0,0 +1,22 @@
+= form_for(@sample) do |f|
+  -#.row
+  -#  .col.pe-2
+  -#    = f.label :institution
+  -#  .col
+  -#    .value= f.object.institution
+
+  = hidden_field_tag :uuid, f.object.uuid
```
matiasgarciaisaia, line 8: Don't send this hidden field. We must not depend on it for the apps workings.

matiasgarciaisaia (changes requested):
We have to change how we create samples.

app/controllers/laboratory_samples_controller.rb
```diff
@@ -2,11 +2,42 @@ class LaboratorySamplesController < ApplicationController
...
+    else
+      render action: 'new'
+    end
+  end
+
+  private
+
+  def uuid_param
+    params.require(:uuid)
+  end
```
matiasgarciaisaia, line 42:
```suggestion
```

matiasgarciaisaia (approved):
:+1: There's a leftover from the previous implementation, but the PR is OK

PR #1089, opened 2021-04-30
Add LaboratorySample Type

ftarulla:
Depends on https://github.com/instedd/cdx/pull/1088

In this PR we add the attribute `type` to model `LaboratorySample`. 
It can be selected when creating a new `LaboratorySample` and then filtered by this attrib in List view

### DIFF
--- a/app/views/components/_cdx_select.html.haml
+++ b/app/views/components/_cdx_select.html.haml
@@ -1,3 +1,3 @@
 - field_name = cdx_select[:form] ? ActionView::Helpers::Tags::TextField.new(cdx_select[:form].object_name, cdx_select[:name], cdx_select[:form], {}).send(:tag_name) : cdx_select[:name]
 - field_value = cdx_select.fetch(:value) { cdx_select[:form] ? cdx_select[:form].object.send(cdx_select[:name]) : nil }.to_s
-= react_component('CdxSelect', name: field_name, items: cdx_select[:items] || [], value: field_value, className: cdx_select[:class], multi: cdx_select[:multi])
+= react_component('CdxSelect', name: field_name, items: cdx_select[:items] || [], value: field_value, className: cdx_select[:class], multi: cdx_select[:multi], searchable: cdx_select[:searchable].nil? || cdx_select[:searchable])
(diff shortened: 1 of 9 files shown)


app/views/components/_cdx_select.html.haml
```diff
@@ -1,3 +1,3 @@
 - field_name = cdx_select[:form] ? ActionView::Helpers::Tags::TextField.new(cdx_select[:form].object_name, cdx_select[:name], cdx_select[:form], {}).send(:tag_name) : cdx_select[:name]
 - field_value = cdx_select.fetch(:value) { cdx_select[:form] ? cdx_select[:form].object.send(cdx_select[:name]) : nil }.to_s
-= react_component('CdxSelect', name: field_name, items: cdx_select[:items] || [], value: field_value, className: cdx_select[:class], multi: cdx_select[:multi])
+= react_component('CdxSelect', name: field_name, items: cdx_select[:items] || [], value: field_value, className: cdx_select[:class], multi: cdx_select[:multi], searchable: cdx_select[:searchable].nil? || cdx_select[:searchable])
```
matiasgarciaisaia, line 3: This will make the component `searchable` by default (ie, if `cdx_select[:searchable]` is nil, the `CdxSelect` has `searchable: true`. Are we sure that's what we want?

app/views/laboratory_samples/_form.html.haml
```diff
@@ -11,6 +11,19 @@
...
 
+  .row
+    .col.pe-2
+      = f.label :type
+    -#.col
+    -#  = f.collection_radio_buttons :type, LaboratorySample.types, :second, :first do |b|
+    -#    .row
+    -#      .col
+    -#        = b.radio_button
+    -#        = b.label
```
matiasgarciaisaia, line 22:
```suggestion
```

I guess we leaned towards having a select instead of a radio button

matiasgarciaisaia (approved):
Great one! 👌

matiasgarciaisaia:
Afterthought: we're missing a check to make the `sample_type` required. As is, the user can avoid selecting a type, and the app will create the sample anyways. That's not OK.

PR #1090, opened 2021-04-30
Add LaboratorySample Edit view

ftarulla:
Depends on https://github.com/instedd/cdx/pull/1089

Adds Edit action to `LaboratorySample`

### DIFF
--- a/app/controllers/laboratory_samples_controller.rb
+++ b/app/controllers/laboratory_samples_controller.rb
@@ -36,9 +36,28 @@ class LaboratorySamplesController < ApplicationController
     end
   end
 
+  def edit
+    @sample = LaboratorySample.find(params[:id])
+    @can_delete = true
+  end
+
+  def update
+    @sample = LaboratorySample.find(params[:id])
+
+    if @sample.update(update_sample_params)
+      redirect_to laboratory_samples_path, notice: 'Sample was successfully updated.'
+    else
+      render action: 'edit'
+    end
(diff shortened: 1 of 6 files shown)


app/controllers/laboratory_samples_controller.rb
```diff
@@ -36,9 +36,28 @@ def create
...
+
   private
 
   def sample_params
     params.require(:laboratory_sample).permit(:sample_type)
   end
+
+  def update_sample_params
+    params.require(:laboratory_sample).permit(:sample_type)
+  end
```
matiasgarciaisaia, line 62: Do we need to have two different sets of params for creation vs update?

app/controllers/laboratory_samples_controller.rb
```diff
@@ -36,9 +36,28 @@ def create
     end
   end
 
+  def edit
+    @sample = LaboratorySample.find(params[:id])
+    @can_delete = true
```
matiasgarciaisaia, line 41: Let's avoid adding options in the UI before allowing the user to perform that action.

If, as of this commit, we don't support sample deletion, let's delay the `@can_delete = true` to whenever we have implemented the deletion.

We're merging all these related PRs in a row, but commits should be independent from each other - if we merge this PR without merging any other (for whatever reason), things should not fail.

matiasgarciaisaia (approved):
🚀 🍟

PR #1091, opened 2021-04-30
Add Delete to LaboratorySample

ftarulla:
Depends on https://github.com/instedd/cdx/pull/1090

In this PR we add the action Delete to `LaboratorySample`

### DIFF
--- a/app/views/laboratory_samples/_filters.haml
+++ b/app/views/laboratory_samples/_filters.haml
@@ -18,4 +18,3 @@
             = cdx_select name: :sample_type, value: params["sample_type"], searchable: false do |select|
               - select.item "", "Show all"
               - select.items LaboratorySample.sample_types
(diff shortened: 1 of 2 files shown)


app/views/laboratory_samples/_filters.haml
```diff
@@ -18,4 +18,3 @@
             = cdx_select name: :sample_type, value: params["sample_type"], searchable: false do |select|
               - select.item "", "Show all"
               - select.items LaboratorySample.sample_types
-
```
matiasgarciaisaia, line 21: Let's pay attention to this whitespace changes. If the change isn't meaningful, let's not add them to the commit.

matiasgarciaisaia (approved):
We're deliberately delaying permissions checks for later PRs.

PR #1094, opened 2021-05-05
Fix LaboratorySample Type presence

ftarulla:
When creating a LaboratorySample, the Type is required.
Also fix the redirect URL when creating a LaboratorySample

### DIFF
--- a/app/assets/stylesheets/_errors.scss
+++ b/app/assets/stylesheets/_errors.scss
@@ -2,3 +2,12 @@
   color: $orange;
   font-style: italic;
 }
+
+form {
+  .error {
+    font-size: 12px;
+    font-style: italic;
+    color: $red;
+    margin-top: 5px;
+  }
+}
--- a/app/controllers/laboratory_samples_controller.rb
+++ b/app/controllers/laboratory_samples_controller.rb
@@ -27,13 +27,12 @@ class LaboratorySamplesController < ApplicationController
 
     @sample = LaboratorySample.new(sample_params.merge(uuid: uuid))
 
-    @sample.save!
-    session.delete(:creating_sample_uuid)
-    redirect_to laboratory_samples_path, notice: 'Sample was successfully created.'
-
-  rescue ActiveRecord::RecordInvalid => ex
-    @sample.errors.add :sample_type, ex.message
-    render action: 'new'
+    if @sample.save
+      session.delete(:creating_sample_uuid)
+      redirect_to edit_laboratory_sample_path(@sample), notice: 'Sample was successfully created.'
+    else
+      render action: 'new'
+    end
   end
 
   def print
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/_errors.scss
```diff
@@ -2,3 +2,12 @@
   color: $orange;
   font-style: italic;
 }
+
+form {
+  .error {
```
matiasgarciaisaia, line 7: I don't think we managed to develop platform for so long without implementing error styles. Isn't this somewhere else?

app/controllers/laboratory_samples_controller.rb
```diff
@@ -27,14 +27,13 @@ def create
...
+    redirect_to laboratory_samples_path, notice: 'Sample was successfully created.'
 
-      next_url = laboratory_samples_path(@sample)
-      redirect_to next_url, notice: 'Sample was successfully created.'
-    else
-      render action: 'new'
-    end
+  rescue ActiveRecord::RecordInvalid => ex
+    @sample.errors.add :sample_type, ex.message
+    render action: 'new'
```
matiasgarciaisaia, line 36: [Using exceptions for normal control flow is usually not a good idea](https://wiki.c2.com/?DontUseExceptionsForFlowControl).

`ActiveRecord::Model#save` already populates the `.errors` array. So I understand this change is not necessary at all.

This code should do:

```
@sample = LaboratorySample.new(sample_params.merge(uuid: uuid))

    if @sample.save
      session.delete(:creating_sample_uuid)
      redirect_to laboratory_samples_path, notice: 'Sample was successfully created.'
    else
      render action: 'new'
    end
```

Why would you change that into a `save!`?

app/views/laboratory_samples/_form.html.haml
```diff
@@ -18,6 +18,8 @@
         .col
           = cdx_select form: f, name: :sample_type, searchable: false do |select|
             - select.items LaboratorySample.sample_types
+          - unless @sample.errors[:sample_type].blank?
+            .error= "Please choose a sample's type"
```
matiasgarciaisaia, line 22:
```suggestion
            .error= "Please choose a sample type"
```

app/controllers/laboratory_samples_controller.rb
matiasgarciaisaia: I'd also redirect users to `edit_laboratory_sample_path(@sample)` on success, so they can see & print the barcode.

matiasgarciaisaia (changes requested):
Let's get rid of the exception throwing & catching.

app/assets/stylesheets/_errors.scss
ftarulla: I thought the same thing but I didn't find it.

The pattern used, is to list the errors at the top of the form. For example in `Patients`:

```
= form_for(@patient) do |f|
  - unless params[:next_url].blank?
    = hidden_field_tag :next_url, params[:next_url]

  = validation_errors @patient
```

Then `validation_errors` renders `partial: "shared/validation_errors"` but without any style (it shows as simple text)

I'm fixing it to this solution and then we may add some style to it.

matiasgarciaisaia (approved):
Awesome! 👌

PR #1097, opened 2021-05-13
Add print selected LaboratorySamples in list

ftarulla:
We add checkboxes to the LaboratorySamples list so that the user may select samples. Then the selected samples may be printed in a bulk action.

### DIFF
--- a/app/models/concerns/multipage_pdf_renderer.rb
+++ b/app/models/concerns/multipage_pdf_renderer.rb
@@ -1,0 +1,35 @@
+require 'open3'
+
+# https://github.com/mileszs/wicked_pdf/issues/339
+class MultipagePdfRenderer
+  def self.combine(documents, options)
+    outfile = ::WickedPdf::WickedPdfTempfile.new('multipage_pdf_renderer.pdf')
+
+    tempfiles = documents.each_with_index.map do |doc, index|
+      file = ::WickedPdf::WickedPdfTempfile.new("multipage_pdf_doc_#{index}.html")
+      file.binmode
+      file.write(doc)
+      file.rewind
+      file
+    end
+
+    filepaths = tempfiles.map{ |tf| tf.path.to_s }
+
+    wickedPdf = WickedPdf.new
+    binary = wickedPdf.send(:find_wkhtmltopdf_binary_path)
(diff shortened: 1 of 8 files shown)


app/models/concerns/multipage_pdf_renderer.rb
```diff
@@ -0,0 +1,35 @@
+require 'open3'
```
matiasgarciaisaia, line 1: Oh my.

app/models/concerns/multipage_pdf_renderer.rb
```diff
@@ -0,0 +1,35 @@
...
+      file.write(doc)
+      file.rewind
+      file
+    end
+
+    filepaths = tempfiles.map{ |tf| tf.path.to_s }
+
+    wickedPdf = WickedPdf.new
+    binary = wickedPdf.send(:find_wkhtmltopdf_binary_path)
+    options = wickedPdf.send(:parse_options, options)
```
matiasgarciaisaia, line 20: Is this Ruby's `send` to send the `find_wkhtmltopdf_binary_path` message? Is there any reason to not directly invoke it (ie, `wickedPdf.find_wkhtmltopdf_binary_path`)?

app/models/concerns/multipage_pdf_renderer.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    command = [binary, '-q']
+    command += options
+    filepaths.each { |fp| command << fp }
+    command << outfile.path.to_s
+    err = Open3.popen3(*command) do |stdin, stdout, stderr|
+      stderr.read
+    end
+
+    raise "Problem generating multipage pdf: #{err}" if err.present?
```
matiasgarciaisaia, line 30: Are we handling this exception somewhere?

matiasgarciaisaia (approved):
Great one!

app/models/concerns/multipage_pdf_renderer.rb
ftarulla: It's necessary to invoke private messages.

matiasgarciaisaia (approved):
🚀 🍟

PR #1102, opened 2021-05-15
Add delete selected LaboratorySamples in list

ftarulla:
The user may select LaboratorySamples in the List view and delete the samples in a bulk action. 
A confirmation dialog is shown to proceed.


devduarte (approved):
Looks good! 🚀

PR #1112, opened 2021-05-21
Fix LaboratorySample to be a CDx Entity

ftarulla:
Related https://github.com/instedd/cdx/issues/1110

Now LaboratorySample uses Entity concern


devduarte (approved):
🚀

Issue #1115, opened 2021-05-27
List Batches


ftarulla:
List batches and allow the user to filter by isolate name

**Empty view**
https://projects.invisionapp.com/d/main#/console/4103589/452219777/preview

**List view**
https://projects.invisionapp.com/d/main#/console/4103589/452219776/preview

PR #1116, opened 2021-05-27
List Batches and filter by isolate name

ftarulla:
Related #1115

### DIFF
--- a/app/controllers/batches_controller.rb
+++ b/app/controllers/batches_controller.rb
@@ -1,0 +1,10 @@
+class BatchesController < ApplicationController
+
+  def index
+    @batches = Batch.all.order('created_at DESC')
+
+    #paginate batches
+    @batches = perform_pagination(@batches)
+  end
+
+end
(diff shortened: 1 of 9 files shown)


db/migrate/20210527024130_add_isolate_name_to_batches.rb
```diff
@@ -0,0 +1,5 @@
+class AddIsolateNameToBatches < ActiveRecord::Migration
```
devduarte, line 1: I think it would be better to add this column in the create table migration file and delete this file at all. Is there any specific reason of why this column is being created in a different file other than its table creation?

devduarte (changes requested):

db/migrate/20210527024130_add_isolate_name_to_batches.rb
ftarulla: yeah!! Great!!

devduarte (approved):
🚀

PR #1128, opened 2021-06-01
Fixes: Batch volume unit, samples_quantity validation

ftarulla:
Related #1129 

Fixes:
- Batch Volume can be greater than 100
- Fix UI: show volume unit
- Fix samples quantity validation
- Fix create LaboratorySamples when Batch created


devduarte (approved):
🚀

PR #1216, opened 2021-06-22
Fix file-dropzone out of view when scrolling file-container

ftarulla:
The file-dropzone was being ​scrolled by the container.


devduarte (approved):
🚀

PR #1222, opened 2021-06-23
Add remove recent assays

ftarulla:
When the user adds new assays a button is shown with the option to remove the new added assays.
[FileList API](https://developer.mozilla.org/en-US/docs/Web/API/FileList) is readonly and files cannot be removed once added to the input that's why this solution removes all the recent added assays.


devduarte (approved):
🚀

Issue #1209, opened 2021-06-16
Edit QC-Sample: Notes


diegoliberman:
When editing a Quality Control batches, the user can add notes.

[Invision](https://projects.invisionapp.com/d/main#/console/4103589/113463974/preview)

### Compromise solution
- The user can add or delete notes.  Editing was left behind for now.
- The delete action was added as an icon.  The 3-dots menu wasn't implemented for now.  
- Only the author of the note can delete it.

PR #1230, opened 2021-06-24
Add Notes to LaboratorySamples

devduarte:
Related: #1209

### DIFF
--- a/app/controllers/laboratory_samples_controller.rb
+++ b/app/controllers/laboratory_samples_controller.rb
@@ -100,15 +100,19 @@ class LaboratorySamplesController < ApplicationController
     # TODO:
     # return unless authorize_resource(patient, UPDATE_PATIENT)
     params = sample_params
-    if @sample.is_quality_control and not params[:notes].empty?
-      note = Note.new
-      note.description = params[:notes]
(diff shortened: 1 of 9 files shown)


app/controllers/laboratory_samples_controller.rb
```diff
@@ -86,6 +86,7 @@ def bulk_print
   def edit
     @sample = LaboratorySample.find(params[:id])
     @sample.test_qc_result = TestQcResult.new if @sample.test_qc_result.nil?
+    @notes = @sample.notes
```
ftarulla, line 89: I think `@notes` is not being used in the view.

app/controllers/laboratory_samples_controller.rb
```diff
@@ -98,7 +99,9 @@ def update
     @sample = LaboratorySample.find(params[:id])
     # TODO:
     # return unless authorize_resource(patient, UPDATE_PATIENT)
-    if @sample.update(sample_params)
+    params = sample_params
```
ftarulla, line 102: We may fix this line along with the branch conflict

ftarulla (approved):

Issue #1231, opened 2021-06-25
Allow to delete laboratory sample notes


devduarte:
Any user should be able to delete its own laboratory sample notes. 
design: https://projects.invisionapp.com/d/main#/console/4103589/452291628/inspect

- [x] Notes can be deleted
- [x] Notes can be deleted only by their Authors.

PR #1245, opened 2021-07-02
Allow user to delete own notes

devduarte:
-related: #1231 

The logged user should only be able to delete its own notes.


ftarulla (changes requested):
The validation is implemented only in the view (to decide whether to show the delete option or not depending on the author of the note). The same validation should be done later on at the time of saving the `sample`

ftarulla (approved):

Issue #1228, opened 2021-06-24
Fields to include when creating a Batch


diegoliberman:
When creating a Batch, it should have the same fields than the Sample with 4 differences in respect to the batch:
1. An extra field: Batch ID
2. Quantity of Samples
3. User shouldn't be able to attach Assays
4. User shouldn't be able to add notes

Issue #1201, opened 2021-06-09
Add batch ID to batch creation


gbatiston:
A piece of feedback we got from the users after this week's demo is that they'd find it useful to have a batch number assigned to the batch when created. 

@jkicillof this is a piece of information we should add to the details when creating a batch, along with Volume, n of samples, technician name, etc

![image](https://user-images.githubusercontent.com/25181052/124316458-0272af80-db4c-11eb-84b7-5752a65b52c5.png)

jkicillof:
[New batch - Empty](https://projects.invisionapp.com/d/main#/console/4103589/452219759/preview)
[New batch - Complete](https://projects.invisionapp.com/d/main#/console/4103589/452219758/preview)
[Edit batch](https://projects.invisionapp.com/d/main#/console/4103589/452219986/preview)
[Edit batch - Selected](https://projects.invisionapp.com/d/main#/console/4103589/452291697/preview)

devduarte:
Just for QA purposes, the combination of `isolate name` and `batch_number` should be unique. There should not be two different batches with the same `isolate_name` and `batch_number`.

mmuller:
@devduarte the validation of the combination of `isolate name` and `batch_number` is case-sensitive. 
Batch `B1` could have `1234` as `batch_number` and `foo` as `isolate_name`.
Batch `B2` could have `1234` as `batch_number` and `FOO` as `isolate_name`.

PR #1250, opened 2021-07-12
Reorder batch fields in batch form to be the same as laboratory samples

devduarte:
-Related: #1228 

Since the creation of laboratory samples was fixed as part of #1248, this issue only reorders input fields the same as it is in Laboratory Samples form. 
The user is not allowed to create notes or either attach assays when creating a batch, it has to be saved first. Batch_id was created in #1201.


ftarulla:
Fix in #1252

PR #1257, opened 2021-07-16
Fix: date-produced format in Samples list

ftarulla:
Date produced column in Samples list was displayed without format


devduarte (approved):
🚀

PR #1271, opened 2021-07-19
Fix Samples navigation depending on start: Samples or Batch edit

ftarulla:
Samples edit returns to Samples-list or Batch-edit depending on where the edit-action started


devduarte (approved):
🚀 🍟

Issue #1227, opened 2021-06-23
Samples - Assays - Allow to upload any file type and allow to download


diegoliberman:
1. We shouldn't restrict the type of file that can be uploaded.
2. Allow to download the files that are attached in the Assays.

PR #1281, opened 2021-07-22
Allow to upload any type of files for assays and download assays

ftarulla:
Related #1227

### DIFF
--- a/app/views/samples/_form_js.haml
+++ b/app/views/samples/_form_js.haml
@@ -9,11 +9,14 @@
         }
 
         $('.assay > .upload-new-file').append(
-          $('<div>', { class: 'picture-container new-file' })
+          $('<div>', { class: 'picture-container new-file assay-file' })
             .append((src == "") ? $('<div>', { class: 'any-filetype icon-document' }) : $('<img>', { class: 'uploaded', src: src }))
             .append($('<div>', { class: 'picture-title', title: file.name })
               .append($('<div>').text(file.name)))
         );
+
+        // render Tooltip and Add-assays action
+        toggleTooltipAction();
       }
     }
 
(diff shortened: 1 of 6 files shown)


app/views/samples/_form_js.haml
```diff
@@ -1,13 +1,22 @@
...
+        var filetype = file.type;
+        if (filetype == 'image/png' || filetype == 'image/jpeg' || filetype == 'image/gif') {
+          src = event.target.result;
+        }
 
-        $('.assay > .upload-new-file')
-          .append(fileDiv);
+        $('.assay > .upload-new-file').append(
+          $('<div>', { class: 'picture-container new-file assay-file' })
+            .append((src == "") ? $('<div>', { class: 'any-filetype icon-document' }) : $('<img>', { class: 'uploaded', src: src }))
```
ggiraldez, line 13: The usage of the `src` variable here is weird. You can abstract this to make it more readable. Something like this:
```js
function fileThumbnail(file, url) {
  if (file.type.startsWith('image/')) {
    return $('<img>', { class: 'uploaded', src: url });
  } else {
    return $('<div>', { class: 'any-filetype icon-document' });
  }
}
```

app/views/samples/_form_js.haml
```diff
@@ -77,3 +90,14 @@
       $('.choose-picture').removeClass('on');
     }
   }, false);
+
+  function toggleTooltipAction() {
```
ggiraldez, line 94: This name is not entirely correct, it's not a toggle. I'd go with `updateAddAssaysControls`.

ggiraldez (approved):
Looks good AFAICS. Minor suggestions in the comments.

Issue #1278, opened 2021-07-20
Mandatory fields in Samples


diegoliberman:
The only mandatory field in Samples is the Production Date.

PR #1296, opened 2021-08-02
Mandatory fields samples

devduarte:
Make production date the only required field for samples

-Related: #1278


ftarulla:
I think we should create the branch from `main` and only push the form validations (i.e. `delete required field validations except for production date` commit)

ftarulla (approved):

PR #1298, opened 2021-08-04
Support attachments in non-default S3 regions

matiasgarciaisaia:
Uploading attachments crashed with an exception due to an issue with S3.

```
AWS::S3::Errors::PermanentRedirect (The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.)
```

We now allow to specify the region-specific endpoint of the S3 bucket to avoid this.

### DIFF
--- a/config/initializers/paperclip_defaults.rb
+++ b/config/initializers/paperclip_defaults.rb
@@ -4,8 +4,12 @@ Paperclip::Attachment.default_options.update({
 })
 
 if s3_bucket = ENV['PAPERCLIP_S3_BUCKET'].presence
-  Rails.application.config.paperclip_defaults = {
+  paperclip_defaults = {
     storage: :s3,
     bucket: s3_bucket
   }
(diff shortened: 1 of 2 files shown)


config/initializers/paperclip_defaults.rb
```diff
@@ -4,8 +4,12 @@
...
 
 if s3_bucket = ENV['PAPERCLIP_S3_BUCKET'].presence
-  Rails.application.config.paperclip_defaults = {
+  paperclip_defaults = {
     storage: :s3,
     bucket: s3_bucket
   }
+  if s3_host_name = ENV['PAPERCLIP_S3_HOST_NAME'].presence
+    paperclip_defaults[:s3_host_name] = s3_host_name
+  end
```
ftarulla, line 13: I think we could use:
`paperclip_defaults[:s3_host_name] =  ENV['PAPERCLIP_S3_HOST_NAME'].presence`

ftarulla (approved):

Issue #1312, opened 2021-08-24
Bug: specimen role shouldn't allow not listed values


diegoliberman:
When creating a batch or a sample, the user shouldn't be able to save the record with a value that is not part of the list of options.

Issue #1311, opened 2021-08-24
Mandatory fields in Batches


diegoliberman:
All fields should be mandatory, except Specimen Role.

PR #1320, opened 2021-09-03
Add SpecimenRole validation to only permit allowed values

devduarte:
-Related #1312 

Only allow to save permitted values for Samples and Batches models. It should only allow options in `specimen_roles.yml` and should allow blanks for Batches since it isn't a required field as described in #1311 . It should allow blanks for Samples as well, described in #1278 

Screen Shot 2021-09-03 at 15 59 52


ysbaddaden:
@devduarte Batch now requires a valid specimen value, which broke spec factories.

ftarulla (approved):

Issue #1313, opened 2021-08-24
Improve specimen role UX


diegoliberman:
1- When the text of the selected option doesn't fit in the box, we should allow the user to see the complete text.  Options: Tooltip, ellipsyis, etc.
2- Clicking the value of the field should open the options box.

PR #1323, opened 2021-09-10
Improve specimen role ux

devduarte:
- Change current specimen role autocomplete component for the custom autocomplete component being used un Policies.
- Store only the code for the specimen_role field in the database.
- Build a script to substring the current saved data in specimen_role field and save only the first letter (which is the code) to make already saved batches and samples compliant with the new requirement.

New autocomplete component fixes: 
- clicking on the field will display the available options.

Current autocomplete component being used.

Screen Shot 2021-09-10 at 17 50 18

New autocomplete component added in these fixes:

Screen Shot 2021-09-10 at 17 52 03

Related: #1313

### DIFF
--- a/lib/tasks/batches.rake
+++ b/lib/tasks/batches.rake
@@ -2,13 +2,12 @@ namespace :batches do
   desc "normalize specimen_role values in Batches to be one of the options listed in fields.yml"
   task normalize_specimen_roles: :environment do
     Batch.find_each do |batch|
-      unless batch.specimen_role.nil?
-        batch.specimen_role = batch.specimen_role[0].downcase
-        if not Batch::SPECIMEN_ROLES_IDS.include? batch.specimen_role
-          batch.specimen_role = nil
-        end
-        batch.save
-      end
+      next if batch.specimen_role.nil?
+
+      specimen_role_id = batch.specimen_role[0].downcase
+      batch.specimen_role = (Batch::SPECIMEN_ROLES_IDS.include? specimen_role_id) ? specimen_role_id : nil
+
+      batch.save
     end
     puts "All Batches processed"
   end
--- a/lib/tasks/batches.rake
+++ b/lib/tasks/batches.rake
@@ -1,0 +1,15 @@
+namespace :batches do
+  desc "normalize specimen_role values in Batches to be one of the options listed in fields.yml"
+  task normalize_specimen_roles: :environment do
+    Batch.find_each do |batch|
+      unless batch.specimen_role.nil?
+        batch.specimen_role = batch.specimen_role[0].downcase
+        if not Batch::SPECIMEN_ROLES_IDS.include? batch.specimen_role
+          batch.specimen_role = nil
+        end
+        batch.save
+      end
+    end
+    puts "All Batches processed"
+  end
+end
--- a/lib/tasks/samples.rake
+++ b/lib/tasks/samples.rake
@@ -2,13 +2,12 @@ namespace :samples do
   desc "normalize specimen_role values in Samples to be one of the options listed in fields.yml"
   task normalize_specimen_roles: :environment do
     Sample.find_each do |sample|
-      unless sample.specimen_role.nil?
-        sample.specimen_role = sample.specimen_role[0].downcase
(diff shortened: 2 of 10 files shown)


ysbaddaden (commented):
Looks good.

One duplication that we could maybe avoid with a concern?

```ruby
module SpecimenRoles
  extend ActiveSupport::Concern

  included do
    validates_inclusion_of :specimen_role,
      in: specimen_roles_ids,
      allow_blank: true,
      message: "is not within valid options (should be one of #{specimen_roles_ids.map(&:upcase).join(', ')})"
  end

  class_methods do
    def specimen_role_ids
      @specimen_role_ids ||= entity_fields.detect { |f| f.name == 'specimen_role' }.options
    end

    def specimen_role_descriptions
      YAML.load_file(File.expand_path("../config/specimen_roles.yml", __FILE__)["specimen_roles"]
    end

    def specimen_roles
      @specimen_roles ||= specimen_role_ids.map do |id|
        { id: id, description: "#{id.upcase} - #{specimen_role_descriptions[id]}" }
      end
    end
  end
end

class Batch
  include SpecimenRoles
end

class Sample
  include SpecimenRoles
end
```

ysbaddaden:
@devduarte also CI is unhappy:

> Validation failed: Core fields Q - Control specimen is not within batch.specimen_role valid options (should be one of b, c, e, f, g, l, o, p, q, r, v), Specimen role is not within valid options (should be one of B, C, E, F, G, L, O, P, Q, R, V)

I believe the factories are now invalid, or something? Or maybe something with `upcase` (B) vs `downcase` (b)?

ysbaddaden:
@devduarte can you integrate #1326 as well?

lib/tasks/batches.rake
```diff
@@ -0,0 +1,15 @@
+namespace :batches do
+  desc "normalize specimen_role values in Batches to be one of the options listed in fields.yml"
+  task normalize_specimen_roles: :environment do
+    Batch.find_each do |batch|
+      unless batch.specimen_role.nil?
```
ftarulla, line 5: I think we could write this block in a clearer way, like this:
```
Batch.find_each do |batch|
   next if batch.specimen_role.nil?

   specimen_role_id = batch.specimen_role[0].downcase
   batch.specimen_role = (Batch::SPECIMEN_ROLES_IDS.include? specimen_role_id) ? specimen_role_id : nil

   batch.save
end
```

lib/tasks/samples.rake
```diff
@@ -0,0 +1,15 @@
+namespace :samples do
+  desc "normalize specimen_role values in Samples to be one of the options listed in fields.yml"
+  task normalize_specimen_roles: :environment do
+    Sample.find_each do |sample|
```
ftarulla, line 4: Same as `Batch`

ftarulla (commented):
There are Pros and Cons using `Tasks` for data migration. [This is a good post](https://www.ombulabs.com/blog/rails/data-migrations/three-useful-data-migrations-patterns-in-rails.html) that list different approaches with pros and cons.
I mention this just to keep it in mind in future data migrations

lib/tasks/batches.rake
ysbaddaden: There are other alternatives to avoid issues around migrations to also migrate data. For example you can regularly replace old migrations with a single one, copying the current `db/schema.rb`  (never seen that one used), or you can migrate data using SQL and static data, which is often possible, and which I'm starting to use more and more. For example:

```sql
UPDATE batches
SET specimen_role = LOWER(LEFT(specimen_role, 1))
WHERE specimen_role REGEXP '^[bcefglopqrv]';

UPDATE batches
SET specimen_role = NULL
WHERE specimen_role NOT IN ('b', 'c', 'e', 'f', 'g', 'l', 'o', 'p', 'q', 'r', 'v');
```

A benefit is that it's incredibly faster, especially on large datasets, so migrations are smooth :boat: 
A drawback is that if the data changes before the release, we must remember to manually update the list here :imp:

ysbaddaden: Also `Batch.where.not(specimen_role: nil).find_each` to only load batches with a specimen_role.

ftarulla (approved):
Great!! 🍟 🚀

PR #1330, opened 2021-09-14
Integrate Sentry to monitor server errors

### LINKED ISSUE #1328: Monitor server errors (Sentry)
Integrate CDx into InSTEDD's Sentry instance to monitor server errors in both staging and production.

We must take care to santitize and filter any sensitive data from reports.

ysbaddaden:
First draft at the integration.

- Uses the legacy `sentry-raven` gem since the new `sentry-rails` gem doesn't support Rails 4.2 (only Rails 5.0+); not the latest release either, since it requires `faraday` 1.0 that won't be installed;
- Only enabled for the `staging` and `production` environments;
- Uses ActiveJob (i.e. Sidekiq) to send the bug reports asynchronously;
- Filters a bunch of words to avoid logging sensitive data, personal data as well as test results; I probably filter too much, but better be safe than sorry :no_good: 

Having an API token configured as per https://github.com/getsentry/action-release would allow to declare releases to Sentry when we create a release tag, possibly allowing to track an error to a release and suspected commits on GitHub if the Sentry project is connected to the GitHub repository.

closes #1328


ysbaddaden:
@matiasgarciaisaia I just tried the integration with the `instedd/cdx-stg` project on Sentry locally.

All that's needed is to configure the `SENTRY_DSN` environment variable to enable reporting to Sentry —or leave it undefined in environments where we don't want to report.

Someone should review the list of sanitized terms. Any param or argument that contains any of the words will be sanitized to *** in Sentry, so we don't leak any sensitive or personal data.

I don't push extra information about the current user or production site. Maybe the role (admin, ...) and production site? We can decide later if we'd like this, or need any of it.

ysbaddaden:
Note: I wrote the list of terms to sanitize by reviewing database columns, and marking suspicious looking names.

diegoliberman:
I think that these 2 fields don't contain any confidential data:
1. isolate_name
2. loinc_code_id

config/initializers/sentry.rb
```diff
@@ -0,0 +1,48 @@
+require "sentry-raven"
+
+Raven.configure do |config|
+  config.dsn = ENV["SENTRY_DSN"]
+  config.environments = %w[staging production]
+  config.current_environment = Rails.env.to_s
```
matiasgarciaisaia, line 6: Can't we declare the `current_environment` (probably from an env variable) without previously listing every available environment?

I'd like to setup a new instance (with a new environment) without having to update every other instance out there to reflect the change.

I'd rather do `config.current_environment = ENV['SENTRY_CURRENT_ENVIRONMENT']` (probably with a `.present?`) than these two lines, but I'm not sure if the `.environments` is required.

config/initializers/sentry.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  if File.exist?(Rails.root.join("VERSION"))
+    config.release = File.read(Rails.root.join("VERSION")).strip
+  end
```
matiasgarciaisaia, line 10: Don't we have a `version` helper somewhere? It sounds odd to have all this logic here instead of doing `config.release = App.version` or something like that.

matiasgarciaisaia (changes requested):
Thanks, @ysbaddaden!

I think the ability to configure the current environment from an environment variable is a must  to merge this.

Let me know if I can help somehow.

config/initializers/sentry.rb
ysbaddaden: The `environments` array is here so we only report in the  `development`. It could be omitted so it'd only report when SENTRY_DSN is set.

ysbaddaden:
@diegoliberman I removed the terms.

@matiasgarciaisaia I removed the `environments` so only the presence of SENTRY_DSN enables reports to Sentry (it's absence disables it). I also extracted a `Cdp.version` helper, this is nicer.

diegoliberman:
@matiasgarciaisaia would you like to take a quick look to the fixes that Julien made?  Thanks!

config/application.rb
```diff
@@ -37,4 +37,10 @@ class Application < Rails::Application
 
     config.assets.paths << Rails.root.join("app", "assets", "fonts")
   end
+
+  def self.version
+    if File.exist?(Rails.root.join("VERSION"))
```
matiasgarciaisaia, line 42: We usually have an `else` statement defaulting to `development` (or `development-#{commit-short-hash}` if we know `git` is available).

See [`instedd/guisso`](https://github.com/instedd/guisso/blob/7015ba06216772214981a63af3c887ffb30fc3d2/config/initializers/version.rb) for an example.

config/initializers/sentry.rb
```diff
@@ -0,0 +1,42 @@
+require "sentry-raven"
+
+Raven.configure do |config|
+  config.dsn = ENV["SENTRY_DSN"]
+  config.current_environment = Rails.env.to_s
```
matiasgarciaisaia, line 5:
```suggestion
  config.current_environment = ENV['SENTRY_CDX_ENVIRONMENT'] || Rails.env.to_s
```

Let's allow to tell different productive environments apart if needed.

matiasgarciaisaia (approved):
Awesome! 👌

ysbaddaden:
@matiasgarciaisaia thank you, I applied the changes.

@devduarte @ftarulla I believe the pull request is ready. Please merge when you feel like it :shipit:

ysbaddaden:
Sigh. CI is broken because of HTTPS issue.

ysbaddaden:
@ftarulla it was a stupid typo: I wrote `Env` instead of `ENV` :facepalm: Thanks CI.

Issue #1332, opened 2021-09-15
Move specimen_role related methods to a concern


devduarte:
As suggested by @ysbaddaden [here](https://github.com/instedd/cdx/pull/1323#pullrequestreview-753656230), we should move specimen_role methods to a concern in order to avoid duplicated code.

devduarte:
## QA guidelines
 Basically the specimen_role field should be working the same way it was woking before making this fix.
- Save values with valid options.
- Check if values are saved correctly.
- Try to save invalid values for the selector.
- Test if the field is showing the correct values when typing.
- Check if the options are displayed when focusing on the field.
- Check what happens if you manually type one the valid options (like 'c') without selecting one from the selector and try to save.

I'll update the issue with more guidelines if any other comes up!

PR #1333, opened 2021-09-17
Specimen Role concern

devduarte:
-Related: #1332


ysbaddaden:
@devduarte @ftarulla Thanks, I reproduced the error by moving `include SpecimenRole` with other includes and more importantly _before_ the `def self.entity_scope` which must be defined. I understand the "hack" was moving the include down?

I understand the bad feeling, since it becomes order dependent, which is meh :unamused: (yet acceptable). I think we can delay the call of `.specimen_role_ids` by using a lambda in the validation, so it's called on demand, not when loading the class:

  ```ruby
  validates_inclusion_of :specimen_roles, in: -> { specimen_role_ids }
  ```

The lambda is usually for when we want to return data that depends on the current record, so it's still kinda hack-ish but we can leave a note/comment about it, and we don't have to remember that the include is order-dependent.

ysbaddaden (approved):

ysbaddaden (approved):

PR #1337, opened 2021-09-23
Add Site to Samples and Batches

ftarulla:
Samples and Batches now belongs to a Site

To test this feature you can create 2 subsites depending from the original Institution. For example:
```
InstitutionFoo
|- SiteA
|_ SiteB
```

Then select `SiteA` and create a new Sample/Batch. Do the same with `SiteB`.
When selecting `SiteA` you should only see its samples but not `SiteB`'s ones. The same selecting `SiteB`.

On the other hand, when selecting `InstitutionFoo` you should see all samples, but unchecking `Selection includes all subsites` you should see only samples created when `InstitutionFoo` was selected.


devduarte (approved):
🚀

Issue #1305, opened 2021-08-20
Validate mandatory fields for Assays


diegoliberman:
When creating an Assay we should validate that it has a RESULT or an ATTACHMENT.  

- [x] Change the Add Assay behaviour.  Today, it's presenting the upload attachment dialog.  Instead, it should add an empty Assay box.
- [x] When the form is saved, validate for all the Assays, that it passes the validation of having a RESULT or an ATTACHMENT.

### Invision showing the error message
https://projects.invisionapp.com/d/main#/console/4103589/113463974/preview

Issue #1331, opened 2021-09-14
Refactor the Assays/Attachment feature


diegoliberman:

straight-shoota:
Could someone (@devduarte ?) fill in the reasons for splitting `AssayFile` as a separate model from `AssayAttachment`? It's just a single-field data type with a 1:1 relation to `AssayAttachment`. As far as I can see, there's no apparent reason why this couldn't have stayed all in a single model.

devduarte:
It was just a technical decision to make our lives a bit easier with this feature. The firsts versions of `AssayAttachment` included the file attribute, as we implemented Dropzone we needed to append the files data to the form but that implied more (unnecessary) work at that time since we needed to create the whole form a Dropzone and make other tweaks to make it work. Also, since we were working with only one file input (can't recall why, now) when user wanted to upload files in two different events, the files from the first event got overwritten. Handling the files async, as Dropzone allowed us to do that, felt like the answer we were looking for. 
Having two separate models is a consecuence of the async file handling. If the user uploaded a file, we would create an `AssayAttachment` but if the browser tab got closed the AssayAttachment would have been created without a parent Sample entity. So we created `AssayFile` to handle this behaviour separately from `AssayAttachment` and `Sample`.

ysbaddaden:
Yes, the same reasons that ActiveStorage uses distinct tables and models!

straight-shoota:
Oooh, I completely forgot that Paperclip *doesn't* 😆 

But I suppose this means we can drop `AssayFile` when we migrate to `ActiveStorage`.

PR #1343, opened 2021-10-18
Validate required fields for assays and improve assays attachs UX 

devduarte:
-Related: #1305 

## Before this fix

- Assays where created only when the user uploads a file
- Assays did not have any validation about having a result 
- Files were stored in a single rails file input

## After this fix

- DropzoneJS is added to upload files via ajax
- Every file input instance ( a dropzoneJS instance) can be managed separately
- Assays validation is added to either have a value for result input or a file attached
- if not, an error message is shown to the user for the assay that failed

Screen Shot 2021-10-18 at 09 24 25

### DIFF
--- a/app/assets/stylesheets/_sample_assays.scss
+++ b/app/assets/stylesheets/_sample_assays.scss
@@ -201,4 +201,6 @@
       z-index: 0;
     }
   }
+
+  .dz-default.dz-message { display: none !important;}
 }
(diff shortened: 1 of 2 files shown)


diegoliberman:
Also includes #1331

ftarulla:
> Tweak the code to check if dropzone can upload multiple files with only one instance.

Indeed! As we talk earlier, let's try to implement this functionality using only one dropzone instance. 
Then we have two options to continue:
- if dropzone can handle the files individually without uploading them, then we only need to handle the files (and each assay) on the server side once the user hits `save`.
- on the other hand, if dropzone cannot handle the files individually then we need to upload the files once the user drops or selects the file for the assay. In this case we also need to implement some kind of `garbage collector task` to remove files that (after some days) are not related to any assay (this could happen if the user selects a file and then cancels the `sample` update)

Issue #1342, opened 2021-10-13
Internal server error in sample creation


mmuller:
**Found in version:** 349fc59

**Steps to reproduce:**

1. Go to create a `Sample`.
2. Fill the `Production date` field.
3. Add a `Note`.
4. Press `Save`. This validation error is shown: `Specimen role is not within valid options`
5. Press `Save` again. 

**Obtained:**

- An internal server error was thrown.

**Error log:**
```{"type":"logentry","body":{"@message":"\nActionView::Template::Error (undefined method `full_name' for nil:NilClass):\n     5:     = f.fields_for :notes, @sample_form.notes do |note|\n     6:       .note\n     7:         .info\n     8:           = \"by #{note.object.user.full_name} #{@localization_helper.format_datetime(note.object.updated_at)} \"\n     9:           - if note.object.user.id == current_user.id\n    10:             .clear-note\n    11:               = note.check_box '_destroy', { id: \"destroy_note_#{note.object.id}\" }\n  app/views/samples/_form_notes.haml:8:in `block in _app_views_samples__form_notes_haml__1210552172823337360_70004225019440'\n  app/views/samples/_form_notes.haml:5:in `_app_views_samples__form_notes_haml__1210552172823337360_70004225019440'\n  app/views/samples/_form.haml:61:in `block in _app_views_samples__form_haml__884746759992460260_52278700'\n  app/views/samples/_form.haml:1:in `_app_views_samples__form_haml__884746759992460260_52278700'\n  app/views/samples/new.haml:11:in `_app_views_samples_new_haml___3722956548224725190_5240","@tags":[],"@pid":"14.55171040","@level":"fatal","@source":"cdx","@timestamp":"2021-10-13T00:10:03.530599Z","@activity":"90c4365c-59e2-c47b-c2dc-49e6fe308dc8"}}```

PR #1346, opened 2021-10-18
Fix notes error when sample form has validation errors

devduarte:
- Related: #1342 

## Before this fix

-Samples where throwing an error because rails treated new notes as saved notes when the form had error validations. This caused nil references to a "saved" object that did not exist.

## After this fix
- Recreate the notes as new ones, when the form has error validations, so nil references are not longer created.

### DIFF
--- a/app/views/samples/_form_notes.haml
+++ b/app/views/samples/_form_notes.haml
@@ -3,6 +3,8 @@
     = f.label :notes
   #notes.col
     = f.fields_for :notes, @sample_form.notes do |note|
+      #note-id-info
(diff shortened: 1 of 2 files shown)


ftarulla (changes requested):
I've found a bug:
- Edit a sample
- Clear `Production Date` so that the server rejects the update (in this case saying _Date produced should be a date in MM/DD/YYYY_)
- Add a new note
- Click on Save
At this point the server returns the error mentioned before
- Click Save again 
Server returns: 
_NoMethodError in Samples#update_ 
_undefined method `full_name' for nil:NilClass_

devduarte:
This issue is resolved in https://github.com/instedd/cdx/pull/1368!

Issue #1335, opened 2021-09-21
Move Inactivation Method related methods to a concern


ftarulla:
Create a concern with the `Inactivation Method` logic. 
Use it in `Samples` and `Batches`

Related #1332

ysbaddaden:
I can't remember who said this, but it struck with me:

> copy it once and _wince_
> copy it twice and _refactor_

A nice article about that: https://wcarss.ca/log/2015/01/write-once-copy-once-refactor-once/

PR #1347, opened 2021-10-18
Move inactivation methods to a concern

devduarte:
-Related #1335 

Unified inclusion validation of inactivation methods in a single concern and include it in batches and samples


ysbaddaden:
@devduarte test factories seem to be broken

PR #1351, opened 2021-10-28
Assays attachment validations

devduarte:
-Related: #1305 and #1331

## Before this fix

- Assays where created only when the user uploads a file
- Assays did not have any validation about having a result 
- Files were stored in a single rails file input

## After this fix

- DropzoneJS is added to upload files via ajax
- Every file input is managed by the same Dropzone instance.
- Created new model for attached files and created association with AssayAttachments
- Assays validation is added to either have a value for result input or a file attached
- if not, an error message is shown to the user for the assay that failed

Screen Shot 2021-10-18 at 09 24 25

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ RUN sed -i '/^mozilla\/DST_Root_CA_X3/s/^/!/' /etc/ca-certificates.conf && updat
 RUN \
   apt-get update && \
   apt-get install -y \
+    cron \
     # wkhtmltopdf dependencies \
     xfonts-75dpi \
     xfonts-base \
@@ -42,6 +43,9 @@ ADD . /app
 
 # Precompile assets
 RUN bundle exec rake assets:precompile RAILS_ENV=production
+
+# Update cron jobs used in whenever gem
+RUN whenever --update-crontab
 
 # Download NNDD
 RUN \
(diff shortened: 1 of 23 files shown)


ftarulla:
> My only hiccup with this PR is the manual building the DOM fragments with jquery —I prefer all HTML to be in the HTML, hence the `<template>` element 😍— but I assume this is how it's done in this project, and we already spent quite some time on this feature, so it's fine.
> 
> Also the JS is broken in phantomjs, which is probably to blame since it's a far too old webkit:
> 
> ```
> ReferenceError: Can't find variable: Symbol
> at http://127.0.0.1:43485/assets/application.js:87781
> at http://127.0.0.1:43485/assets/application.js:87807
> at http://127.0.0.1:43485/assets/application.js:88419
> at http://127.0.0.1:43485/assets/application.js:88384 in init (Capybara::Poltergeist::JavascriptError)
> ```

I think we can create a fix-issue to replace `jquery` with `<template>` I wouldn't use (just for now) another framework like `Stimulus` or `htmx` as the project already have React (it is used in some of the components)
Maybe after upgrading we can discuss to unify all future components to use a framework 🤓

ysbaddaden:
> You have 1 pending migration:
>  20211104135226 MigrateAssayAttachmentPicturesToAssayFile
> Run `rake db:migrate` to update your database then try again.

The `db/schema.rb` file isn't up to date. You must run `rake db:migrate` locally, that will update `db/schema.rb` which you can commit, and it will fix the issue.

ftarulla (approved):
`Gemfile.lock` should be commited to the repo. Then, it is good to merge 🍟 🚀

Issue #1336, opened 2021-09-21
Code refactor date_produced in Sample/SampleForm and Batch/BatchForm


ftarulla:
Move this logic to a concern and the validation should be executed at the `model` level

devduarte:
QA Guidelines:
- Batches and Samples should work as expected.
- Test valid dates for both, Batches and Samples.
- Test saving non valid dates for both batches and samples.
- Test saving blank dates for both batches and samples.
- Test saving alphanumeric characters for both batches and samples.

PR #1364, opened 2021-11-09
Date_produced Concern

devduarte:
-Related #1336 

-Move shared methods to date_produced concern
-Remove non used code
-Corrected views to reference the correct date methods 
-Unifiy date validation usage in batches and samples instead of having it mixed between samples, batches, batch_form and sample_form

### DIFF
--- a/app/models/batch_form.rb
+++ b/app/models/batch_form.rb
@@ -31,6 +31,8 @@ class BatchForm
   attr_accessor :samples_quantity
   delegate :id, :new_record?, :persisted?, to: :batch
 
+  validates_numericality_of :samples_quantity, greater_than: 0, message: "value must be greater than 0", if: :creating_batch?
+
   def self.for(batch)
     new.tap do |form|
       form.batch = batch
@@ -48,7 +50,7 @@ class BatchForm
     self.date_produced = if @batch.date_produced.is_a?(Time)
                            @batch.date_produced
                          else
-                           Time.strptime(@batch.date_produced, date_format[:pattern]) rescue @batch.date_produced
+                           Time.strptime(@batch.date_produced, Batch.date_format[:pattern]) rescue @batch.date_produced
                          end
   end
 
@@ -108,8 +110,6 @@ class BatchForm
     return false
   end
 
-  validates_numericality_of :samples_quantity, greater_than: 0, message: "value must be greater than 0", if: :creating_batch?
-
   def creating_batch?
     self.batch.id.nil?
   end
@@ -119,15 +119,11 @@ class BatchForm
   # BatchForm#date_produced will return always a string ready to be used by the user input with the user locale
   # BatchForm#date_produced= will accept either String or Time. The String will be converted if possible to a Time using the user locale
 
-  def date_format
-    { pattern: I18n.t('date.input_format.pattern'), placeholder: I18n.t('date.input_format.placeholder') }
-  end
-
   def date_produced
     value = @date_produced
 
     if value.is_a?(Time)
-      return value.strftime(date_format[:pattern])
+      return value.strftime(Batch.date_format[:pattern])
     end
 
     value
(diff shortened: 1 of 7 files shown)


app/models/concerns/date_produced.rb
```diff
@@ -0,0 +1,30 @@
+module DateProduced
+  extend ActiveSupport::Concern
+
+  included do
+    validate :date_produced_is_a_date
+
+    def date_produced_is_a_date
```
ftarulla, line 7: Although I think it doesn't make any difference in the end result, instance methods should be declare outside `included` block

app/models/concerns/date_produced.rb
```diff
@@ -0,0 +1,30 @@
...
+      return if date_produced.blank?
+      errors.add(:date_produced, "should be a date in #{self.class.date_produced_placeholder}") unless date_produced.is_a?(Time)
+    end
+
+    def date_produced_description
```
ftarulla, line 12: Same here: instance method inside `included` block

app/models/batch_form.rb
```diff
@@ -67,7 +70,7 @@ def create_sample
       institution: self.institution,
       site: self.site,
       sample_identifiers: [SampleIdentifier.new],
-      date_produced: self.date_produced,
+      date_produced: @date_produced,
```
ftarulla, line 73: This change shouldn't be necessary 🤔  since`date_produced` is a method declared in this class. 
I tried replacing it with the previous `self.date_produced` and something isn't working as before. 
I think we should find out what's the problem (given that the method itself had some implementation changes)

devduarte: `date_produced` is a field in `fields.yml` spcified as a `date`. Doing `self.date_produced` returns a string, so when we created a `Sample` with `self.date_produced` we were assigning a string type of value to the attribute. Trying `Sample.last.date_produced = Time.now` or `Sample.last.date_produced = '11/11/2021'` in a rails console results in a successful save operation, which means that the `Sample` model can be saved with either a `String` or a `Time` types. 
Also, a very importante detail: the date validation was being made in the `SampleForm` model, so when we were saving a `String` type to `date_produced` of the `Sample` model the validation was not being triggered. 

As of now, adding the validation to the `Sample` model triggers the validation and it fails since the value we are trying to save is a `String` type instead of a `Time` type. So we can't save `String` types to the `date_produced` field anymore. To comply with that I had two options: use `@date_produced` value or convert `self.date_produced` to Time (something like `Time.strptime(self.date_produced, Batch.date_format[:pattern])`).
Just to be cleaner, I used `@date_produced`, but any other suggestion is welcome!

app/models/concerns/date_produced.rb
```diff
@@ -0,0 +1,31 @@
+module DateProduced
+  extend ActiveSupport::Concern
+
+  included do
+    validate :date_produced_is_a_date
+  end
+
+  def date_produced_is_a_date
+    return if date_produced.blank?
+    errors.add(:date_produced, "should be a date in #{self.class.date_produced_placeholder}") unless date_produced.is_a?(Time)
```
ysbaddaden, line 10: @devduarte @ftarulla Maybe the validator could allow the String form? Something like:

```ruby
case date_produced?
when Time
  # ok
when String
  # error unless matches regexp (or that it parses as a date)
else
  # error
end
```

I think it would permit to continue using the `#date_produced` accessor instead of the ivar?

ysbaddaden (commented):
Great!

I'm just wondering why all the dates in specs are now the same exact date? I suspect a copy-paste, but I liked the diversity there used to be :smiley:

ysbaddaden:
@ftarulla just curious: why did you need to rollback?

ftarulla:
> @ftarulla just curious: why did you need to rollback?

We need the previous implementation just to be consistent with the logic used in `@dateProduced`, and the methods `dateProduced` and `dateProduced=`

I'm going to open an tech-debt issue to refactor this logic (but we need to refactor `Sample`, `Batch`, `SampleForm` and `BatchForm` to remove repeated logic

ftarulla (approved):
🍟 🚀

Issue #1360, opened 2021-11-08
Samples: QC sample missing distinctive icon


gbatiston:
When a sample Specimen Role value is changed from any value to `Q - Control Specimen`, a distinctive icon is displayed in the sample list within the batch details to indicate that sample as the Quality control. This is currently not the case in PRD. 

Steps to reproduce:
1.- Create a new batch with at least 2 samples
2.- Set the Specimen Role to any value BUT Q
3.- Navigate to the batch details, once created
4.- Select any of the Samples from the list
5.- Update the Specimen Role to its current value to `Q - Control Specimen`
6.- Navigate to the Batch details view

Expected: the updated sample has an icon that indicates its Specimen Role as Q
Actual: the updated sample does not have any indicator of its Specimen Role

Issue #1344, opened 2021-10-18
Specimen role default value bug


devduarte:
Saving a Sample with the required fields fails because Specimen Role default value is not setted correctly

![Screen Shot 2021-10-18 at 09 27 42](https://user-images.githubusercontent.com/23437283/137730243-01cbc02d-0c95-400d-b5ff-fd79ce449d93.png)

PR #1366, opened 2021-11-09
Fix Sample QC missing icon

ftarulla:
Related #1360 

When editing a Batch, the list of samples inside that batch is shown. The QC samples should have a distinctive icon but this icon was not showing. This PR fixes this bug.

Also in this PR:
- SpecimenRole Concern: Code refactor. Declare private some class methods that should not be part of the class API.

**Specimen Role** select component
Fix Samples and Batches form: specimen_role select component always had the QC option selected (when creating and when **updating** a `sample` or a `batch`)

Related to #1344 (issue) and #1345 (PR). `specimen_role` is optional in `Samples` and `Batches` models. But if a default `specimen_role` needs to be shown when creating a `sample` or a `batch` (ask: @diegoliberman @devduarte) then the fix should be applied to the model used in the create form (and not to the views/forms)

### DIFF
--- a/app/models/concerns/specimen_role.rb
+++ b/app/models/concerns/specimen_role.rb
@@ -9,19 +9,33 @@ module SpecimenRole
 
   end
 
+  def is_quality_control
+    specimen_role == 'q'
+  end
+
   class_methods do
+    def specimen_roles
+      @@specimen_roles ||= build_specimen_roles
+    end
+
+    private
+
+    def build_specimen_roles
+      specimen_role_ids.map do |id|
+        { id: id, description: "#{id.upcase} - #{specimen_role_descriptions[id]}" }
+      end
+    end
+
     def specimen_role_ids
-      @specimen_role_ids ||= entity_fields.detect { |f| f.name == 'specimen_role' }.options
+      @@specimen_role_ids ||= entity_fields.detect { |f| f.name == 'specimen_role' }.options
     end
 
     def specimen_role_descriptions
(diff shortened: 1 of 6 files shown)


devduarte (approved):
🚀

app/models/concerns/specimen_role.rb
```diff
@@ -9,19 +9,33 @@ module SpecimenRole
 
   end
 
+  def is_quality_control
+    specimen_role == 'q'
+  end
+
   class_methods do
+    def specimen_roles
+      @@specimen_roles ||= build_specimen_roles
```
ysbaddaden, line 18: Be careful with class variables `@@` (vs class instance variables `@`) in concerns. It can quickly bite back. Yes, classes themselves are instances (of `Class`) and they can have both class variables (shared with descendant classes) _and_ instance variables :exploding_head: (private to the class).

I think it should work here, since its within a`class_methods` block that should inject the methods right into the class (so the variables shouldn't be shared by the module to different classes), but... I'm always wary of them.

ysbaddaden: But maybe the block merely creates a `ClassMethods` module and extends the class with it, in which case the class variables will bite:

once specimen roles are loaded for Sample or Batch they will be used for the other one :skull:

ftarulla: Great suggestion! I'll fix it 👍

ysbaddaden (approved):

devduarte (approved):

Issue #1362, opened 2021-11-09
Improve information displayed on Sample list overview


gbatiston:
The UCSD team feedback mentioned that they would like to have more information displayed on the Sample list overview. Currently, we are only displaying
| Sample ID | Lab Technician | Production Date |

The team has explained they would find it more useful if they can have the following fields
| Sample ID | Isolate | Inactivation | Specimen Role | Production Date |

 In this case, the Lab Technician column would be removed from view.

PR #1367, opened 2021-11-12
Reorder Sample list columns

devduarte:
-Related #1362

## Before this fix
Screen Shot 2021-11-12 at 14 00 00

## After this fix
Screen Shot 2021-11-12 at 14 01 06

### DIFF
--- a/app/models/concerns/specimen_role.rb
+++ b/app/models/concerns/specimen_role.rb
@@ -12,9 +12,17 @@ module SpecimenRole
     specimen_role == 'q'
   end
 
+  def specimen_role_description
+    self.class.specimen_role_description(specimen_role)
+  end
+
   class_methods do
     def specimen_roles
       @specimen_roles ||= build_specimen_roles
+    end
+
+    def specimen_role_description(id)
+      "#{id.upcase} - #{specimen_role_descriptions[id]}" if id
     end
 
     private
--- a/app/views/samples/index.haml
+++ b/app/views/samples/index.haml
@@ -17,14 +17,18 @@
               .icon-trash.btn-icon
           - t.columns do
             %col{:width => "5%"}
-            %col{:width => "40%"}
-            %col{:width => "30%"}
-            %col{:width => "25%"}
+            %col{:width => "20%"}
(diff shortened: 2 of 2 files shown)


app/models/concerns/specimen_role.rb
```diff
@@ -7,6 +7,11 @@ module SpecimenRole
                            allow_blank: true,
                            message: "is not within valid options"
 
+    def specimen_role_description(id)
+      specimen_role_id = id || ""
+      self.class.specimen_role_description(specimen_role_id)
+    end
```
ysbaddaden, line 13:
```suggestion
    def specimen_role_description
      self.class.specimen_role_description(specimen_role)
    end
```

We don't need to pass the specimen role id to the method, we already have access to the `#specimen_role` attribute. Also the class method will already check if the value is nil.

This allows us to use `sample.specimen_role_description` and `batch.specimen_role_description` directly :raised_hands:

app/models/concerns/specimen_role.rb
```diff
@@ -18,9 +23,14 @@ def specimen_role_descriptions
       YAML.load_file(File.join("app", "models", "config", "specimen_roles.yml"))["specimen_roles"]
     end
 
+    def specimen_role_description(id)
+      specimen_role_id = id || ""
+      "#{specimen_role_id.upcase} - #{specimen_role_descriptions[specimen_role_id] || ""}"
+    end
```
ysbaddaden, line 29: I understand that the `specimen_role` attribute is optional? Then maybe we should return nil or an empty string in that case? For example:

```suggestion
    def specimen_role_description(id)
      "#{id.upcase} - #{specimen_role_descriptions[id]}" if id
    end
```

ysbaddaden (approved):
Just a few suggestions to make the feature even nicer :)

app/models/concerns/specimen_role.rb
ftarulla: Also keep in mind that changes were made to this concern in #1366 (that should be merged before this PR)

ysbaddaden (approved):
LGTM!

Issue #1372, opened 2021-11-22
[BUG] More than one file cannot be added in assay without result


mmuller:
**Steps to reproduce:**

1. Go to create a sample.
2. Add an assay.
3. Upload a file to the assay.
4. Add other assay.
5. Upload a file to the second assay.
6. Try to save the sample.

**Obtained result:**
* The sample cannot be created.
* An error (422) is thrown.

PR #1374, opened 2021-11-26
Fix AssayFile filename and filesize bugs

devduarte:
-Related #1372 

-Fix filename problems when they include special characters.
-Change thumbnail size to allow dropzone to generate thumbnails largar than 10MB so it generates the assay card. We are not allowing files larger than 10mb to be saved, but we generate the card to force the validation to fail when it gets saved and show the correct error message to the user.
-Change error message when the AssayAttachment validation fails.

## Before this fix
Screen Shot 2021-11-26 at 13 57 38

## After this fix
Screen Shot 2021-11-26 at 13 59 39

### DIFF
--- a/app/views/samples/_form_assays_js.haml
+++ b/app/views/samples/_form_assays_js.haml
@@ -107,7 +107,7 @@
   }, false);
 
   function updateAddAssaysControls() {
-    var assays = $('.assay-file');
+    var assays = $('.card-container');
     if (assays.length > 0) {
       $('.assay > .upload-new-file > .tooltip').addClass('nodisplay');
     } else {
--- a/app/views/samples/_form_assays_js.haml
+++ b/app/views/samples/_form_assays_js.haml
@@ -9,7 +9,7 @@
       uploadMultiple: true,
       renameFile: function (file) {
         var completeFileName = file.name;
-        var sanitizedName = completeFileName.replace(/[^a-z^A-Z^0-9\^_.-]/g, '_');
+        var sanitizedName = completeFileName.replace(/[^\p{Letter}\p{Number}_.-]/gu, '_');
         var fileName = sanitizedName.split('.')[0];
         return sanitizedName.replace(fileName, fileName + Date.now());
       },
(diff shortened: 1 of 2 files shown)


app/views/samples/_form_assays_js.haml
```diff
@@ -3,19 +3,28 @@
     dropzoneOptions = {
       paramName: "assay_files",
       autoProcessQueue: true,
+      maxThumbnailFilesize: 20,
       parallelUploads: 10,
       url: '/assay_files/create',
       uploadMultiple: true,
+      renameFile: function (file) {
+        var completeFileName = file.name;
+        var sanitizedName = completeFileName.replace(/[^a-z^A-Z^0-9\^_.-]/g, '_');
```
ysbaddaden, line 12: This is _very_ restrictive, just a few ASCII characters. As long as CDx operates only in English speaking countries that should be acceptable, but it will be a problem in other languages, like Spanish or [Sinhala](https://en.wikipedia.org/wiki/Sinhala_language) (Sri Lanka).

From my experience we don't need to sanitize much characters (see this [wikipedia article](https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words)). I think the following would be nicer internationally. It should accepts all unicode letters & numbers, whatever the language (I didn't know `\p{}` before, it's _very_ nice):

```js
/[^\p{Alpha}\p{Numeric}_.-]/gu
```

Also, it's better to store the file with a completely different name (e.g. as a UUID, like `c7b78085-33b4-4a93-929c-9358ab1b2952.pdf`) then set the `Content-Disposition` header when downloading to set the original filename (using the `filename*=UTF-8''` form). Then you wouldn't have to bother sanitizing the filename.

app/views/samples/_form_assays_js.haml
```diff
@@ -3,19 +3,28 @@
...
+        var fileName = sanitizedName.split('.')[0];
+        return sanitizedName.replace(fileName, fileName + Date.now());
```
ysbaddaden, line 14: There may be dots inside the filename (e.g. `file.name.pdf)`), so you'll want to split and skip the last entry:

```js
var fileName = sanitizedName.split(".").slice(0, -1).join(".")
```

Thought it's possible that it may not even be the extension, which means that you won't sanitize the end of the filename, which could be a problem. So if you need to inject a timestamp:

```js
var parts = sanitizedFilename.split(".")
var extname = parts.pop()
parts.push(Date.now())
parts.push(extname)
return filename
```

But again: storing the files as a mere UUID would be a better solution.

ysbaddaden (approved):

Issue #1363, opened 2021-11-09
Provide more filtering options in the Sample view


gbatiston:
The UCSD team has mentioned that being able to use filters to search for specific batches and samples is very useful, but that they sometimes miss being able to filter using other fields despite the provided ones (sample ID, Batch ID, Isolate name)

They have made two specific requests:
- To be able to filter by Specimen Role in the Sample View
- To be able to filter by Date, in the Batch view.  

### Notes on Date filter
The idea is to start with the easiest solution we can think of.  Two fields for picking start/end dates and done.  I think we don't have a calendar picker, hence let's add fields with a format hint, as it was done with the `Production Date` field in the `Sample` form.

jkicillof:
[Samples](https://projects.invisionapp.com/d/main#/console/4103589/453801737/preview)
[Batches](https://projects.invisionapp.com/d/main#/console/4103589/453801740/preview)

PR #1377, opened 2021-12-01
Add spcimen_role filter for samples

devduarte:
-Related #1363 

## Current filters
Screen Shot 2021-12-01 at 20 16 50

## Filters added within this PR
Screen Shot 2021-12-01 at 20 18 37


ysbaddaden (approved):
I understand that specimen_role was an "entity" and is now also an actual column, so we can use the column to search?

devduarte:
> I understand that specimen_role was an "entity" and is now also an actual column, so we can use the column to search?

Yup! we were saving the value of `specimen_role` in the `core_fields` column (which is defined in `Entity`). For filtering we need the actual column in the DB so that is why this line `attribute_field :specimen_role, copy: true` is added.

Issue #1382, opened 2021-12-09
Parse String saved values into Time for Batches and Samples


devduarte:
Before #1336  Samples that were being created along with batches were being saved as a string . After we refactored the date_produced, the date validations were placed correctly in both models (samples and batches ). The already saved data were not valid anymore so the migrations we created for the filters issues (#1363 and #1376 ) did not saved successfully the invalid models.

diegoliberman:
Testing guidelines:
play with the date produced field, using pre-existent samples, and new samples.  Within the form itself and within the search filters.

Same for specimen role.

Thanks!

PR #1383, opened 2021-12-09
parse string saved dates into time for batches and samples

devduarte:
-Fixes: #1382 

Samples and Batches prior to #1363 and #1372 were not being considered by the new filters (specimen_role for Samples and date_produced for Batches).
-Parse String saved dates into Time for Samples.
-Fix a bug in the code that was preventing batches from being correctly saved. Replace `Time.strptime(value, Batch.date_format[:pattern]) rescue batch.date_produced` for `Time.strptime(batch.date_produced, Batch.date_format[:pattern]) rescue batch.date_produced`


ysbaddaden (approved):
LGTM

Issue #1393, opened 2021-12-17
delete and move samples.rake and batches.rake to migrations files


devduarte:
In order to get rid of manual rake tasks to normalize specimen_role data that is causing filter problems in production, move them to migrations file.

diegoliberman:
Can't be tested in Staging.

diegoliberman:
This had already been released in 0.17.1

Issue #1376, opened 2021-12-01
Provide more filtering options in the Batch view


devduarte:
Splited #1363 into two issues. This issue is for the Batches view.

The UCSD team has mentioned that being able to use filters to search for specific batches and samples is very useful, but that they sometimes miss being able to filter using other fields despite the provided ones (sample ID, Batch ID, Isolate name)

They have made this request:

To be able to filter by Date, in the Batch view.

Notes on Date filter
The idea is to start with the easiest solution we can think of. Two fields for picking start/end dates and done. I think we don't have a calendar picker, hence let's add fields with a format hint, as it was done with the Production Date field in the Sample form.

Design: [Batches](https://projects.invisionapp.com/d/main#/console/4103589/453801740/preview)

ysbaddaden:
@diegoliberman @devduarte There is a `date` input type that displays the date in a local way (e.g. DD/MM/YYYY in France) but formats the value as per ISO8601 (YYYY-MM-DD) + presents a calendar picker.

See https://developer.mozilla.org/fr/docs/Web/HTML/Element/Input/date

diegoliberman:
I didn't know that! Looks awesome and a low-hanging fruit.  Do you want to give it a try @devduarte !

devduarte:
@diegoliberman @ysbaddaden Sure! I'll create the PR as it is now and set it as a draft  PR so I can implement the `date` input.

PR #1394, opened 2021-12-17
Normalize specimen_role and date_produced data in migrations

devduarte:
-Fixes: #1393 

- We want to normalize `specimen_role` old data to be compliant with the samples validations in order to be able to successfully save them and get the filters working for batches and samples.

Motivation:
After deploying #1376 , #1363 and #1382 to production, filters were not working for some samples. After a research, we found out that there were samples with and old specimen_role data format(they were stored with their description concatenated, i.e.: "Q - Control specimen", "C - Calibrator", "E - Electronic QC", etc) that were preventing some samples to be saved successfully (and make the filters include them) as this value makes those models invalid.


db/migrate/20211217152414_normalize_specimen_role_and_date_produced_samples.rb
```diff
@@ -0,0 +1,27 @@
+class NormalizeSpecimenRoleAndDateProducedSamples < ActiveRecord::Migration
+  def up
+    Sample.find_each do |sample|
+      unless sample.specimen_role.blank?
+        specimen_role_id = sample.specimen_role[0].downcase
+        sample.specimen_role = (Sample.specimen_role_ids_list.include? specimen_role_id) ? specimen_role_id : sample.specimen_role
```
matiasgarciaisaia, line 6: I think we're relying too much on Models for these migrations, and [that's an issue](https://guides.rubyonrails.org/v3.2/migrations.html#using-models-in-your-migrations).

Specifically, this `Sample.specimen_role_ids_list` method may be renamed in the code after we merge this migration, so if we want to run the migration in an instance later - the migration will break.

For this case, I think I'd hard-code the list of known `specimen_role_ids` in the migration itself.

app/models/concerns/specimen_role.rb
```diff
@@ -25,6 +25,10 @@ def specimen_role_description(id)
       "#{id.upcase} - #{specimen_role_descriptions[id]}" if id
     end
 
+    def specimen_role_ids_list
+      specimen_role_ids
+    end
```
matiasgarciaisaia, line 30: Wouldn't it be better to just publish the private method?

db/migrate/20211217160437_normalize_specimen_role_and_date_produced_batches.rb
```diff
@@ -0,0 +1,27 @@
+class NormalizeSpecimenRoleAndDateProducedBatches < ActiveRecord::Migration
+  def up
+    Batch.find_each do |batch|
+      unless batch.specimen_role.blank?
+        specimen_role_id = batch.specimen_role[0].downcase
+        batch.specimen_role = (Batch.specimen_role_ids_list.include? specimen_role_id) ? specimen_role_id : batch.specimen_role
```
matiasgarciaisaia, line 6: Ditto

matiasgarciaisaia (changes requested):
I'd also try to reword the commit message a bit, focusing on _what we're trying to do_ instead of detailing _how we're doing it_.

matiasgarciaisaia (approved):
👍 Good one!

PR #1402, opened 2021-12-24
Fix typo in AssayAttachment to AssayFile data migration

### LINKED ISSUE #1397: Bug in attachments/assays migration
An old migration had a typo, which caused some failed migrations when we changed the Attachments approach, for the Assays.  

This caused some attachments to remain broken.

1. Confirm that the attachments are associated with the institution `Prod DiegoL`.
2. If #1 is confirmed, delete/clean-up the data/attachments.
3. Analyze if we need to fix the typo in the migration, for rolling back purposes, and if it's worth doing.

devduarte:
-Fixes #1397 

This PR fixes a typo that was preventing a migration to successfully create AssayFiles from AssayAttachments

### DIFF
--- a/db/migrate/20211104135226_migrate_assay_attachment_pictures_to_assay_file.rb
+++ b/db/migrate/20211104135226_migrate_assay_attachment_pictures_to_assay_file.rb
@@ -7,10 +7,9 @@ class MigrateAssayAttachmentPicturesToAssayFile < ActiveRecord::Migration
       assay_file.picture_file_size = assay_attachment.picture_file_size
       assay_file.picture_updated_at = assay_attachment.picture_updated_at
       assay_file.assay_attachment = assay_attachment
-      if assay_file.save!
-        assay_attachment.assay_file = assay_file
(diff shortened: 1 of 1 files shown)


db/migrate/20211104135226_migrate_assay_attachment_pictures_to_assay_file.rb
```diff
@@ -7,9 +7,9 @@ def up
       assay_file.picture_file_size = assay_attachment.picture_file_size
       assay_file.picture_updated_at = assay_attachment.picture_updated_at
       assay_file.assay_attachment = assay_attachment
-      if assay_file.save
-        assay_attachment.assay_File = assay_file
-        assay_attachment.save
+      if assay_file.save!
```
ftarulla, line 10: `save!` raises an exception if failed (it does not return a boolean)

ftarulla (approved):

diegoliberman:
@devduarte we can merge this PR, right?

PR #1384, opened 2021-12-16
Invite User or Institution

devduarte:
Fixes: #1356 

-Create frontend flow to allow the user to choose to invite an institution or a user.

## Before this PR
Screen Shot 2021-12-16 at 09 28 10

## After this PR
Screen Shot 2021-12-16 at 09 28 35

Screen Shot 2021-12-16 at 09 28 52

Screen Shot 2021-12-16 at 09 29 06

Screen Shot 2021-12-16 at 09 29 28

PR #1403, opened 2021-12-27
Create Institution Invites

### LINKED ISSUE #1357: Untitled story 5724

devduarte:
-Fixes #1357 

- Pass data between the invitations forms to be able to send it to the backend.
- Add Pending Institutions Invite model and table.
- Add endpoint to create a new user from the Institution Invite flow.
- Create a new `PendingInstitutionInvite` accordingly.

(There are no screenshots since this was mostly backend code)

### DIFF
--- a/db/migrate/20211216160348_create_pending_institution_invites.rb
+++ b/db/migrate/20211216160348_create_pending_institution_invites.rb
@@ -1,0 +1,10 @@
+class CreatePendingInstitutionInvites < ActiveRecord::Migration
+  def change
+    create_table :pending_institution_invites do |t|
+      t.belongs_to :user, index: true
+      t.string :institution_name
+      t.string :institution_kind, default: 'institution'
+      t.timestamps null: false
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


ftarulla (approved):
I think we should merge this to a feature branch (for example `feature/institution-invite`) following #1384

db/migrate/20211216160348_create_pending_institution_invites.rb
```diff
@@ -0,0 +1,11 @@
+class CreatePendingInstitutionInvites < ActiveRecord::Migration
+  def change
+    create_table :pending_institution_invites do |t|
```
ftarulla, line 3: Do we need a new table for pending institutions? 
Can we use the current `institutions` table and add a new column to represent the invitation state? (for example `invitation_token` , `invitation_created_at`, invitation_sent_at`, etc. in the `users` table)

ysbaddaden: I have mixed feelings about that. On one side I don't like duplicating a table, its columns, validations, but on the other side I like having a distinct place to store pending data (so we don't have to filter institutions _by default_ => bugs), we can have the columns that we need, can have a distinct set of validations, so...?

PR #1408, opened 2022-01-05
Redirect institution creation after email confirmation

devduarte:
Fixes: #1359 

-Deleted repeated validation 
-Created new endpoint to redirect user to new institution form with saved data from pending invites

## New institution screen with already filled data
Screen Shot 2022-01-04 at 22 02 45

### DIFF
--- a/app/controllers/institutions_controller.rb
+++ b/app/controllers/institutions_controller.rb
@@ -32,7 +32,7 @@ class InstitutionsController < ApplicationController
   end
 
   def create
-    unless PendingInstitutionInvite.where(invited_user_id: current_user,  status: 'pending').count > 0
+    if PendingInstitutionInvite.where(invited_user_id: current_user,  status: 'accepted').count > 0
       no_data_allowed
       return
     end
--- a/app/controllers/institutions_controller.rb
+++ b/app/controllers/institutions_controller.rb
@@ -32,6 +32,10 @@ class InstitutionsController < ApplicationController
   end
 
   def create
+    unless PendingInstitutionInvite.where(invited_user_id: current_user,  status: 'pending').count > 0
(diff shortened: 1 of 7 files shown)


ftarulla (approved):

app/controllers/institutions_controller.rb
```diff
@@ -32,7 +32,7 @@ def new
   end
 
   def create
-    unless PendingInstitutionInvite.where(invited_user_id: current_user,  status: 'pending').count > 0
+    if PendingInstitutionInvite.where(invited_user_id: current_user,  status: 'accepted').count > 0
```
ftarulla, line 35: **Important** 
This means: a user that already was invited to create a new institution (and accepted it) cannot be invited again to create another institution

ftarulla (approved):
We need to tackle the following situation:
When a user that already exists is invited to create a new institution, an invitation email is sent (not needed because the user is already created). 
This leads to another issue: the user cannot create a second institution (see last comment about `if PendingInstitutionInvite.where(invited_user_id: current_user, status: 'accepted').count > 0`)

A possible solution:
If the user already exists then the email should have a link to create the institution (bypassing the user creation)

We can implement this in a following issue

devduarte:
> We need to tackle the following situation: When a user that already exists is invited to create a new institution, an invitation email is sent (not needed because the user is already created). This leads to another issue: the user cannot create a second institution (see last comment about `if PendingInstitutionInvite.where(invited_user_id: current_user, status: 'accepted').count > 0`)
> 
> A possible solution: If the user already exists then the email should have a link to create the institution (bypassing the user creation)
> 
> We can implement this in a following issue

Issue #1414 created for fixing this bug

Issue #1308, opened 2021-08-24
Access restrictions for Batches & Samples by Site


diegoliberman:
For context, read #1301.

PR #1409, opened 2022-01-13
Fix Samples and Batches read access: Add view for only reading

ftarulla:
Related [#1308 comment](https://github.com/instedd/cdx/issues/1308#issuecomment-949739840)

# The Problem
When the user has READ access (for `Samples` or `Batches`) but not UPDATE access, it throws a 403 error because we always render the `edit` view for those entities (since there isn't any step between `view` and `edit` actions)

# Solution
When the list of `Samples` and `Batches` are being rendered, the links are generated taking into account if the user can UPDATE or only READ.


app/controllers/batches_controller.rb
```diff
@@ -18,6 +18,12 @@ def index
     @batches = perform_pagination(@batches)
   end
 
+  def show
+    batch = Batch.find(params[:id])
+    @batch_form = BatchForm.for(batch)
+    return unless authorize_resource(batch, READ_BATCH)
```
ysbaddaden, line 24: There must be something missing here, because whatever the authorize result, Rails will still render the `show` template? unless `#authorize_resource` raises? But then the `return` becomes superfluous.

app/controllers/samples_controller.rb
```diff
@@ -16,6 +16,17 @@ def index
     @samples = perform_pagination(@samples)
   end
 
+  def show
+    sample = Sample.find(params[:id])
+    @sample_form = SampleForm.for(sample)
+    return unless authorize_resource(sample, READ_SAMPLE)
```
ysbaddaden, line 22: Same here: the template will still be rendered? It will crash because `@view_helper` is undefined.

ysbaddaden (commented):
Adding the `show` action is great, but maybe we could reuse the `_form.haml` partials by passing a `readonly` local variable? That would avoid duplicating their contents which adds to maintenance cost (and potential bugs).

Maybe add some tests to verify access rights to the `show` actions?

app/controllers/samples_controller.rb
ftarulla: If not authorized the server returns a 403

ftarulla:
> Adding the `show` action is great, but maybe we could reuse the `_form.haml` partials by passing a `readonly` local variable? That would avoid duplicating their contents which adds to maintenance cost (and potential bugs).
> 
> Maybe add some tests to verify access rights to the `show` actions?

Yeah, that was my first choice but it resulted in a mess when trying to parameterized the `assays` form. Maybe I can try to parameterized `_form.haml` but keep `_show_assays.haml` and `_show_notes.haml`

ysbaddaden (approved):
That's great! No more form duplication.

It would be great to have a few tests, thought, to verify the behavior is indeed correct, and avoid future regressions :)

devduarte (approved):

Issue #1350, opened 2021-10-28
Invites - Users can create an Institution via a received invitation


gbatiston:
### Use case
 I want to create my Institution by accepting an invitation pre-filled with my Institution details

### Details
This feature is the second step on a series of changes that we'll be working on in the coming weeks/months. Please note that Institution (capital I) refers to the entity (regardless of the type), and institution (lower case i) refers to the specific type.

The UCSD will invite awardees to join CDx, but as each awardee represents a different institution (there are 49 of them), it does not make sense to have these users joining UCSD's profile. Instead, what we want is to give each of these institutions the possibility of joining CDx so that we can, at a later stage, allow UCSD to assign them ownership of samples and batches, and also allow awardees to upload the results of their tests and link their devices to our platform.

Once UCSD invites an Institution to join CDx, the new Institution should receive an invite where they can validate the associated email, provide a new password for their account, validate provided details for the Institution and confirm the creation of the account. 

### Behaviour
- The awardee receives an email invite containing
- - a welcome message
- - the details of the invitation
- - a link to accept the invite 
- the link in the email validates the email address 
- - this step today is done by sending an email to the user after they create their account and provide their password
- once the awardee clicks on the link to validate the email, they have to provide a password to complete the account creation step
- after a password is provided and confirmed, the pre-filled Institution details are displayed
- users can edit the details of the Institution before confirming
- after confirmation, the Institution is created and the awardee is redirected to CDx homepage

### Designs, Flow, other details
![Screen Shot 2021-10-28 at 12 13 10 PM](https://user-images.githubusercontent.com/2902379/139236052-c33ce433-4d1c-4fdd-9f54-9c47fe404c2a.png)

### Questions

devduarte:
We added an Institutions invitation feature. This means that you can invite a User (new or already existing) that will create a new Institution with pre-filled data.

Corner cases:
-Send multiple invitations to a user, the pre-filled data for the new Institution form should match the data filled in the invitation modal.
-Accept invitations in a different order in which the invitations were sent.
-Accept multiple times the same invitation for an Institution creation. The same Institution should not be created more than one time.
-Try going back with the back button in the URL bar in the browser after an Institution is created. The browser will lead the user in the creation form, but after hitting the save button the platform should not create the same Institution.
-Complete, if possible, the invitation modal without providing an Institution name or a User email. 
[...]

PR #1410, opened 2022-01-13
Create Institutions from recieved invites

devduarte:
Fixes: #1350 

-This is only a feature branch.


devduarte:
There was a mistake for this PR, the correct one is #1411

Issue #1414, opened 2022-01-15
Change email invitation flow for a persisted user


devduarte:
When a user that already exists is invited to create a new institution, an invitation email is sent (not needed because the user is already created).
This leads to another issue: the user cannot create a second institution due to `if PendingInstitutionInvite.where(invited_user_id: current_user, status: 'accepted').count > 0` logic

ysbaddaden:
If the email exists then we can skip the user creation _but_ we must authenticate the user to create institution.

PR #1415, opened 2022-01-18
Change invitation flow for persisted users

devduarte:
Fixes: #1414 

-Add a different email when inviting an existing user
-Change user reference from user_id to user_email in PendingInstitutionsInvite so we don't need a persisted user to be a valid invite
-Send the PendingInstitutionInvite id to the emails so we can validate later on with the actual invite to avoid multiple institutions creation
-Added some dummy validations to prevent the user from sending problematic data and to give some UI feedback for UX
-Changed the migration version so we don't cause any trouble with table/column dependency

## Before this fix
-The same email was being sent to any kind of user (persisted or not).
Screen Shot 2022-01-18 at 08 38 29

## After this fix
-Two types of mails, with different flows, are being sent accordingly. One for New Users
Screen Shot 2022-01-18 at 08 38 29

And the other one for persisted users
Screen Shot 2022-01-18 at 08 42 07

### DIFF
--- a/app/controllers/institutions_controller.rb
+++ b/app/controllers/institutions_controller.rb
@@ -32,9 +32,10 @@ class InstitutionsController < ApplicationController
 
   def create
     inst_params = institution_params
+    pending_invite_id = get_pending_invite_id_from_params(inst_params)
 
-    if inst_params.has_key?(:pending_institution_invite_id)
-      @pending_invite = PendingInstitutionInvite.find_by(id: inst_params["pending_institution_invite_id"])
+    if pending_invite_id
+      @pending_invite = PendingInstitutionInvite.find_by(id: pending_invite_id )
 
       unless @pending_invite and @pending_invite.status == 'pending'
         no_data_allowed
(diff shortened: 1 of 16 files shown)


app/controllers/institutions_controller.rb
```diff
@@ -32,11 +31,18 @@ def new
   end
 
   def create
-    if PendingInstitutionInvite.where(invited_user_id: current_user,  status: 'accepted').count > 0
-      no_data_allowed
-      return
+    inst_params = institution_params
+
+    if inst_params.has_key?(:pending_institution_invite_id)
```
ftarulla, line 36: We should extract the "get param" logic so we can write:
```
id = get_pending_invite_id
```
Then `id` could be `nil`

app/controllers/institutions_controller.rb
```diff
@@ -104,22 +113,24 @@ def no_data_allowed
   end
 
   def new_from_invite_data
-    pending_invite = PendingInstitutionInvite.where(invited_user_id: current_user,  status: 'pending').last
+    pending_institution_invite_id = params["pending_institution_invite_id"]
```
ftarulla, line 116: idem "get param" logic

app/controllers/users_controller.rb
```diff
@@ -139,13 +143,14 @@ def mark_as_invited(user)
   end
 
   def create_institution_invite
-    invited_user = User.find_by(email: @user_email)
+    invited_user_email = params[:user_invite_data]["email"]
     pending_invite = PendingInstitutionInvite.new
-    pending_invite.invited_user = invited_user unless invited_user.nil?
+    pending_invite.invited_user_email = invited_user_email unless invited_user_email.nil?
```
ftarulla, line 148: Should we send the `pending invite` if `invited_user_email.nil?`?

ftarulla (changes requested):

ftarulla (approved):

PR #1416, opened 2022-01-18
Remove client_max_body_size

devduarte:
-Remove client_max_body_size.conf from the repo and from the `dockerfile`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,4 +57,3 @@ RUN mkdir -p /app/tmp /app/log && chown -R app:app /app/tmp /app/log
 
 # Add config files
 ADD docker/web-run /etc/service/web/run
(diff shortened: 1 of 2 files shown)


ysbaddaden:
I'm not sure we want to _remove_ the max body size, it should just be large enough to fit the largest files that must uploaded. Having a 10MB or 20MB limit seems reasonable.

ping @matiasgarciaisaia

matiasgarciaisaia:
We're deleting the limit here so we don't override the runtime-defined limit that we can now set in the base `instedd/nginx-rails` image since instedd/docker-nginx-rails@2771b2a36732877f82c4ff1f6d06a2b18a5dd325.

There's always nginx's default if users don't set a custom one.

ysbaddaden:
Oh sweet!

matiasgarciaisaia (approved):
👍 🎉

PR #1432, opened 2022-01-27
Add notification when sending user and institution invitations

ftarulla:
Show a notification after sending email for **Institution** invites.
Also added the same for **User** invites.


devduarte (approved):
🚀 🍟

Issue #1418, opened 2022-01-21
[BUG] Console returns 404 when sending an invite for invited users, UI does not warn user


gbatiston:
### Details
After sending out an invite for an institution, if I try to send an invite to the same email the site seems to be unresponsive when hitting the send button "Send" after providing Admin contact details. When inspecting the console to see what is actually happening, I noticed there is an HTTP 404 ("not found") response. but there are no error messages displayed to the user. 

### Steps to reproduce
1.- Complete all steps to send an invite to an institution
2.- Start the process to invite an institution, use the exact same email as before when completing the Admin Contact Details

Expected: error message "An invitation was already sent to this user"
Actual: 404 NOT FOUND reported via console, no error message display on the UI

diegoliberman:
@gbatiston let's negotiate your expectations
> Expected: error message "An invitation was already sent to this user"

For sure the 404 is a bug, but we will send the invitation as if it was a new invitation.  This is how it's already implemented.  If we need to restrict users from sending multiple invitations to the same recipient, that would be a new scope and we need to implement as well a feature for resending invitations in case the recipient didn't get the email.

gbatiston:
@diegoliberman sure, let's go ahead with that.

PR #1433, opened 2022-01-28
Fix 404 when inviting institutions with an email that belongs to a user that already exists

ftarulla:
Related #1418

When the user already exists in the db, the `invited_by_id` attribute is `nil` (see `UsersController#send_institution_invite` and `UsersController#mark_as_invited`) and this lead to an error when trying to send the email (the sender could not be found)


devduarte (approved):
🚀

PR #1434, opened 2022-01-31
GHA: optimize workflow runs

ysbaddaden:
Avoid running the workflow twice, once for a push and another one for a pull request.

Also avoids building the docker image until we are on a branch or tag for which we want to build an image (main, preview/, release/ or we pushed a release tag).


ysbaddaden:
This patch is identical to the change I made for Surveda :)

ysbaddaden:
I'm trusting myself here :)

Issue #1422, opened 2022-01-21
[BUG] Multiple invites for Institutions can be sent at once


gbatiston:
### Details
A user can send an invite to multiple email addresses at once, provided the addresses are separated by commas. All users get the invitation, all addresses it was sent to are displayed on the email details. Only one user will be able to actually create the account, the rest of the users are redirected to the Login page, not the create account flow. 

### Steps to reproduce
1.- Complete the steps to send an Invite to an institution
2.- In the Admin Contact Details, enter more than 1 email address, separated by commas (e.g: germanbatiston@manas.tech, gbatiston@manas.tech)

Expected: user is notified that only one email address can be provided per invite
Actual: both invites are sent. Both users get the email. The first one to accept the invite can create the institution.
![Screen Shot 2022-01-21 at 1 30 51 PM](https://user-images.githubusercontent.com/2902379/150528174-274b023d-1639-476c-8f57-327b86c1cf04.png)

diegoliberman:
@gbatiston how come this is a bug? 

1. This was never discussed AFAIK.  Nor mentioned in the epic #1349.  
2. Adding multiple addresses doesn't make sense in the [current UI](https://projects.invisionapp.com/share/Z7OD63XAWKF#/screens/460430729) as it doesn't make sense to enter a name of a person and multiple email addresses.
3. Implementing this **feature**, wouldn't be trivial, nor easy, it's kind of medium.  
4. If we were to implement this feature, what role would the users have? The first one is the admin, but what about the rest? In my opinion, the admin needs to invite the rest of the users after creating the institution.

gbatiston:
@diegoliberman I think there might be some misunderstanding as to how you understood what it is I am reporting here. I absolutely agree with you that it should be the admin for the other institution the one who should invite the rest of the users, that is why I am saying that it should **not** be possible to enter multiple email addresses, and just asking that we make sure that it is not possible to do so, letting the user know that only one email address can be provided. 

I have read again the issue and I can see that perhaps it's the description of the details that caused some confusion? Those are the details/description of what the issue is and what is happening, if you check the **actual** in the steps to reproduce, you'll see that it should not be possible to send multiple invites when creating an institution. 

I hope this makes it clearer.

diegoliberman:
ha! what a big misunderstanding.

@ftarulla, we got it wrong.  Here we just need to prevent users from entering multiple email addresses.

PR #1435, opened 2022-01-31
Add frontend email input validation

devduarte:
Fixes: #1422 

## Added validation in this PR
Screen Shot 2022-01-31 at 14 38 28

### DIFF
--- a/app/assets/javascripts/components/admin_invite_form.js.jsx
+++ b/app/assets/javascripts/components/admin_invite_form.js.jsx
@@ -6,7 +6,7 @@ var AdminInviteForm = React.createClass({
       email: '',
       includeMessage: false,
       message: '',
-      hasEmailError: false,
+      hasEmailError: null,
       sendButtonDisabled: true
     };
   },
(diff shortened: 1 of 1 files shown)


ysbaddaden (commented):
Looks good to me. Just a suggestion to set the error message as a prop, instead of a boolean.

What about doing the validation on the backend now? So we're secure and fine.

Adding a simple `validates_format_of` to `User` + a check in UsersController to not send the invitation if the user is invalid should be enough:

```ruby
if user.persisted?
  # send create institution
elsif user.valid?
  # send invitation to register & create institution
else
  # oops, error
end
```

Ideally, a tiny controller test to verify that no invitation is sent with an invalid email would be :100:

devduarte:
For the backend validation I have created #1436 so we can tackle it later. Meanwhile, we have this little validation for the user. I'm adding a reference of [this comment](https://github.com/instedd/cdx/pull/1435#pullrequestreview-868220779) to #1436 for when the time comes we'll have an idea of how it can be solved.

ysbaddaden (approved):
Allright!

Issue #1439, opened 2022-02-01
[BUG] Samples and batches cannot be created


mmuller:
**Steps to reproduce:**

1. Sign in as Institution admin.
2. Create a sample/batch.

**Obtained:**
- The submit button is missing in the sample/batch form.
- The fields are disabled in the sample/batch form.

PR #1440, opened 2022-02-01
Allow authorized users to create Batches and Samples

devduarte:
Fixes: #1439 

-Add can_update intance variable to allow authorized users to be able to create Batches and Samples.


diegoliberman (approved):
It makes sense that the update instance needs to be set to true also for the creation.  I hesitated because I'm used to see the create permission separated from update.  But in CDx, the permission to update grants permission to create.  

So, LGTM.

ysbaddaden (commented):
Good catch!

Issue #1438, opened 2022-02-01
[BUG] Internal server error in Institution creation


mmuller:
**Steps to reproduce:**

1. Create an account.
2. Create your first Institution. 
3. Open the browser console.

**Obtained:**
- The institution is not crated.
- An internal server error is logged in the browser console:

PR #1441, opened 2022-02-01
Presence of pending_invite_id conditional

devduarte:
Fixes: #1438 

-Fix conditional to check if pending_invite_id param is not blank, since it could be present but have no value in it.

### DIFF
--- a/app/controllers/institutions_controller.rb
+++ b/app/controllers/institutions_controller.rb
@@ -34,7 +34,7 @@ class InstitutionsController < ApplicationController
     inst_params = institution_params
     pending_invite_id = get_pending_invite_id_from_params(inst_params)
 
-    if pending_invite_id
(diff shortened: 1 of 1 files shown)


diegoliberman (approved):
Let's ship it!

ysbaddaden:
Tiny alternative: move the `.blank?` check into the method and make it simpler as a bonus:

```ruby
  def get_pending_invite_id_from_params(inst_params)
    inst_params["pending_institution_invite_id"].blank?
  end
```

diegoliberman:
> Tiny alternative: move the `.blank?` check into the method and make it simpler as a bonus:
> 
> ```ruby
>   def get_pending_invite_id_from_params(inst_params)
>     inst_params["pending_institution_invite_id"].blank?
>   end
> ```

To be tackled in #1444

Issue #1442, opened 2022-02-01
[BUG] Samples/batches cannot be created after a validation error occurs


mmuller:
**Steps to reproduce:**

1. Sign in as Institution admin.
2. Go to create a sample/batch.
3. Don't fill in any fields.
4. Press save.

**Obtained:**
- The submit button is missing in the sample/batch form.
- The fields are disabled in the sample/batch form.

PR #1443, opened 2022-02-01
Allow Samples and Batches editing when save fails

devduarte:
- Fixes: #1442 

-Added `can_update` instance variable when create and update actions fails, since they need the variable needs to be set again before render to allow the user to edit the form.

Here's a list of the actions in the controller and the value that `can_update` needs to have in each one.

List of actions / can_update value

Index -> does not need can_update
New -> true
Show -> false
Create/sucess -> does not need can_update since it redirects to index
Create/failure -> true
Edit -> true
Update/sucess -> does not need can_update since it redirects to index
Update/failure -> true
Print -> does not need can_update
Bulk_print -> does not need can_update
destroy -> does not need can_update
Bulk_destroy -> does not need can_update

### DIFF
--- a/app/controllers/batches_controller.rb
+++ b/app/controllers/batches_controller.rb
@@ -61,6 +61,7 @@ class BatchesController < ApplicationController
     if @batch_form.create
       redirect_to batches_path, notice: 'Batch was successfully created.'
     else
+      @can_update = true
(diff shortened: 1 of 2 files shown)


diegoliberman (approved):
:rocket: :fries:

PR #1458, opened 2022-02-11
Add specs for institution invites

### LINKED ISSUE #1437: The invites feature doesn't have tests
The whole feature doesn't have any test. There is no level of confidence that the feature will continue to work properly as we continue to refactor & implement features. Regressions _will_ happen at some point with no safe points.

straight-shoota:
Resolves #1437

This adds specs for the invite feature for institutions (added in #1411). They are mostly unit tests concerning `InstitutionsController` and `UsersController`.

I did not add any specs for the frontend because I don't think there are any yet in this project. And I have no idea how to start with that.

I also included some changes to the actual app code which where pretty necessary or at least looked like a very good idea. I'm adding code comments to elaborate. Let me know if that's fine in this PR or I should extract that.


app/controllers/institutions_controller.rb
```diff
@@ -128,15 +128,15 @@ def new_from_invite_data
     end
   end
 
+  private
+
   def accept_pending_invite
-    unless @pending_invite.nil? and current_user.email != @pending_invite.invited_user_email
+    if @pending_invite && current_user.email == @pending_invite.invited_user_email
```
straight-shoota, line 134: This was an actual logic error, probably caused by confusion about double negation (it should've been `or`, not `and`). I applied a De Morgan transform to make the expression easier to understand.

app/controllers/institutions_controller.rb
```diff
@@ -128,15 +128,15 @@ def new_from_invite_data
     end
   end
 
+  private
```
straight-shoota, line 131: `accept_pending_invite` is not a public controller method, so moving it into the private section.

app/controllers/users_controller.rb
```diff
@@ -145,16 +145,13 @@ def mark_as_invited(user)
...
-      pending_invite.save!
-      @pending_invite = pending_invite
-    end
+    pending_invite = PendingInstitutionInvite.new
+    pending_invite.invited_user_email = params[:user_invite_data]["email"]
+    pending_invite.invited_by_user = current_user
+    pending_invite.institution_name = params["institution_data"]["name"]
+    pending_invite.institution_kind = params["institution_data"]["type"]
+    pending_invite.save!
+    @pending_invite = pending_invite
```
straight-shoota, line 154: The guard for `invited_user_email.nil?` is unnecessary. If that value is nil, it blows up before reaching this method. So I figure this simplification would be nice.

app/views/invitation_mailer/invite_message.html.erb
```diff
@@ -1,4 +1,4 @@
-<h1><%= "#{User.find(@user.invited_by_id).full_name} has invited you to join the team as #{@role.name}" %></h2>
+<h1><%= "#{@invitation_sender.full_name} has invited you to join the team as #{@role.name}" %></h2>
```
straight-shoota, line 1: Templates should not pull in data from the data base. This makes it easier to test. The inviting user does not need to be persisted.

spec/controllers/institutions_controller_spec.rb
```diff
@@ -40,6 +40,51 @@
...
+
+        invite.reload
+        expect(invite).to be_is_pending
+      end
+
+      it "fails if invite already accepted" do
+        invite = PendingInstitutionInvite.make(invited_user_email: user.email, status: "accepted")
+        expect do
+          post :create, {institution: {name: invite.institution_name, kind: invite.institution_kind, pending_institution_invite_id: invite.id}}
+        end.to raise_error(ActionView::MissingTemplate)
```
straight-shoota, line 85: This error is actually a bit weird and I believe it's a bug that this happens.

When there is no matching invite, the controller calls `#no_data_allowed` which is supposed to redirect. But the redirect is guarded by a restriction that seems to prevent it from executing. As a result, the `new_from_invite_data` falls through and rails tries to render the respective template (which doesn't exist).

spec/controllers/institutions_controller_spec.rb
```diff
@@ -63,4 +108,32 @@
...
+
+    it "with accepted invite" do
+      invite = PendingInstitutionInvite.make(status: "accepted", invited_user_email: user.email)
+      expect do
+        get :new_from_invite_data, {pending_institution_invite_id: invite.id}
+      end.to raise_error(ActionView::MissingTemplate)
+    end
+
+    it "gets redirect from other page when pending invitation" do
+      dummy_institution = Institution.make # there must be some institution in the database for the redirect to trigger
```
straight-shoota, line 132: This is a bit uncool. Reason is here: https://github.com/instedd/cdx/blob/41c36c823840d4c126afa67f4d5bff3c81aec065/app/controllers/application_controller.rb#L63-L68
If there are no institutions, a different redirect rule catches first.

spec/features/invitation_spec.rb
```diff
@@ -58,4 +59,40 @@ def last_email
...
+          expect(page.name.value).to eq invite.institution_name
+          # TODO: Checking the radio button doesn't work for some reason
+          # expect(page).to have_checked_field("institution_kind_#{invite.institution_kind}")
+          # page.choose("institution_kind_manufacturer")
+
+          page.submit
+        end
+
+        # FIXME: This flash message probably shouldn't be there
+        expect(page).to have_content("You are already signed in")
```
straight-shoota, line 94: This flash message is probably a bug?

spec/controllers/institutions_controller_spec.rb
```diff
@@ -63,4 +108,32 @@
...
+        get :new_from_invite_data
+      end.to raise_error(ActionView::MissingTemplate)
+    end
+
+    it "with accepted invite" do
+      invite = PendingInstitutionInvite.make(status: "accepted", invited_user_email: user.email)
+      expect do
+        get :new_from_invite_data, {pending_institution_invite_id: invite.id}
+      end.to raise_error(ActionView::MissingTemplate)
+    end
```
ysbaddaden, line 129: I think this is also a bug: shouldn't we report a message that the invitation has already been accepted, instead of crashing? @gbatiston @diegoliberman

spec/features/invitation_spec.rb
```diff
@@ -58,4 +59,42 @@ def last_email
   def email_count
     ActionMailer::Base.deliveries.count
   end
+
+  describe "insitution invite" do
```
ysbaddaden, line 63:
```suggestion
  describe "institution invite" do
```

spec/features/invitation_spec.rb
ysbaddaden: Yeah, that sounds fishy. Maybe the test logs a user in already? And when we accept the invitation we try to log the user in _again_ and that prints that error?

ysbaddaden (approved):
Great job, here! You even went into the Capybara specs :)

I just ~~had~~ left a few comments to maybe improve things a bit.

spec/controllers/institutions_controller_spec.rb
diegoliberman: Yes, that would be better.

spec/controllers/institutions_controller_spec.rb
straight-shoota: I understand that. Still think it would be cleaner to act on a pending invitation first, if just to keep the test simple.
It's not practically relevant at all, so we can just leave it as is.

spec/mailers/invitation_mailer_spec.rb
```diff
@@ -0,0 +1,45 @@
+require "spec_helper"
+
+RSpec.describe InvitationMailer, type: :mailer do
+  let!(:current_user) { User.new(email: "existing@example.com", first_name: "Current", last_name: "User") }
+  let!(:invite) { PendingInstitutionInvite.make(invited_by_user: current_user) }
+  let!(:new_user) {
+    User.make(:invited_pending).tap do |user|
+      user.send("generate_invitation_token")
+    end
```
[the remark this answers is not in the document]
straight-shoota: I agree. But I have no idea how that would work 🤷
Couldn't find a solution quickly, so I deferred it and use the ugly approach for now. I'll take another look.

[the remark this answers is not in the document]
straight-shoota: Resolved in [08d6977](https://github.com/instedd/cdx/pull/1458/commits/08d697761d4bf9174869ce3caeeef0d023911795)

spec/controllers/institutions_controller_spec.rb
straight-shoota: This is tracked in #1467

spec/controllers/institutions_controller_spec.rb
straight-shoota: This is tracked in #1467

ysbaddaden:
I'll look at the conflicts. There shouldn't be much of them.

ysbaddaden:
The patch is minimal. Let's merge!

Issue #1388, opened 2021-12-17
2- Allow to transfer a sample that belongs to a batch


diegoliberman:
### What happens in the physical world?
The Sample is sent to the recipient.  But there is not such a thing as a physical batch that is being sent as well.  They just add a stamp in the sample with the Original Batch ID.  This is just for tracking purposes.  They make a copy of the batch's QC data and send that copy along with the sample.

### CDx approach
We want to reflect reality as much as possible within the system. Hence, we want the Sample to be owned by the Recipient institution.  But we don't want the Batch to change of owner.  And we do want the recipient to be able to see a copy of the QC data.

### Potential solution
Note: This issue will remove the IF added in the item 4.2 of #1385 .

For the case that the sample being transfer belongs to a batch, we need to implement the following logic:

- [x] Same logic than implemented in #1385 
> 4.1 The send action changes the access of the Sample, to make it belong to the recipient institution.
- [x] The Batch ID needs to appear in the UI when seeing the Sample, below the Sample UID.  This field can be displayed for all samples that belong to a batch, and for all sample that used to belong to a batch.
[Invision](https://projects.invisionapp.com/d/main#/console/4103589/113463974/preview)
- [x] The sender of the sample, when accessing to the associated batch, shouldn't see the sample that was transferred in the list of samples that belong to the batch.
- [x] ~~A sample that belongs to a batch, can't be transferred again.  This item is just for simplifying, as we should think what happens with the Sample Transfer Data.~~
- [x] In the Samples view, if view filter by Batch ID, we should ideally show also the samples that were received from another institution and that used to belong to that batch id.

![image](https://user-images.githubusercontent.com/25181052/152162091-bed89af1-2fc1-4243-83ce-fad84e3c6bdb.png)

The above diagram can be found [here](https://drive.google.com/drive/folders/1u4MFveAZi6YcyQkRsmGxHEeRRl7aEIxJ).

devduarte:
Samples that belong to a batch and get transferred will be able to be transferred again as it was easier to implement this way.

diegoliberman:
> * item is just for simplifying, a

@devduarte please change the Issue description.  We aim to have in the description what was indeed implemented.

diegoliberman:
Verified in `Version: dev-525a860 (build 1058)`.  However, the batch ID isn't shown for samples that were transferred.  I won't create a ticket for this, let's see if the user asks for it.

straight-shoota:
[...]

PR #1466, opened 2022-02-17
Transfer samples with batches

devduarte:
Fixes: #1388 

-Allow samples that belong to a batch to be transferred
-Add old_batch_number to samples to able to filter by batch_number (from Batch model) and old_batch_number

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -44,12 +44,19 @@ class ApplicationController < ActionController::Base
   end
 
   def authorize_resource(resource, action)
+    unless authorize_resource?(resource, action)
+      log_authorization_warn resource, action
+      head :forbidden
+      return nil
+    end
+    true
+  end
+
+  def authorize_resource?(resource, action)
     if has_access?(resource, action)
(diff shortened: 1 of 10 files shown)


ysbaddaden (commented):
@devduarte Do we really need to keep searching a sample by its batch number _after_ a transfer? Does the new institution even know the batch number?

If we have to, then I wonder whether we shouldn't always copy the batch number to the samples? It would duplicate the value but I assume that defined once and it will never change? That would dramatically simplify the SQL for the search: no need for an INNER or LEFT JOIN and no OR!

diegoliberman:
> @devduarte Do we really need to keep searching a sample by its batch number _after_ a transfer? Does the new institution even know the batch number?

@ysbaddaden the institution does know the batch number, as it's included in the sample form.  If I remember correctly, the potential use case is that the recipient might get a request from the sender: "hey, disregard all samples from batch xxxx because they got contaminated".

@gbatiston could you please help me here?  I hesitate about my answer.  Thanks.

straight-shoota (approved):
LGTM

PR #1470, opened 2022-02-17
Re-enable non cucumber integration specs

ysbaddaden:
The specs have been disabled for some reason but they still pass, with a minor correction to account for the recently modified invitation flow (choose between new user or new institution).

We reenable them in addition to the Cucumber integration specs, that use the same Page Objects, but don't test the same things.

Note: this is based on the `chore/improve-specs-ux` branch. You should focus on the last 3 commits.


ysbaddaden:
@ftarulla Do you remember why you excluded the integration tests from `spec/features` last year (April 2021)?

Issue #909, opened 2016-02-26
Edit user page inconsistency between actions


nekron:
Adding or removing a role is being done in place, without need to save the user.

For enabling and disabling the user you need to submit the entire form by clicking the save button.

This causes a weird behavior like having a cancel button that doesn't actually cancels the modified roles. (See #908)

Unify the submit method of the page. Use the save button to be consistent with the rest of the app.

nditada:
Moving to backlog, this is minor.

PR #1471, opened 2022-02-17
GHA: optimize CI runs

ysbaddaden:
There was room to optimize the CI run-time, by caching the docker images and ruby gem dependencies across workflows, and... running the integration tests in parallel to the unit tests.

Based on the other spec rework branches, so please review the last 2 commits that change the YAML files under `/.github` :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,27 +11,63 @@ on:
     branches:
       - "**:**"
 
+env:
+  DOCKER_COMPOSE: docker-compose -f docker-compose.yml -f docker-compose.ci.yml
+
 jobs:
   test:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set environment up
+      - uses: actions/cache@v2
+        id: docker-cache
+        with:
+          path: tmp/docker
+          key: Docker-v2-${{ hashFiles('Dockerfile-dev', 'docker-compose.yml', '/docker-compose.ci.yml') }}
+          restore-keys: |
+            Docker-v2-
+
+      - name: Load Docker image
+        if: steps.docker-cache.outputs.cache-hit == 'true'
         run: |
-          docker-compose build
-          docker-compose run --rm -e RAILS_ENV=test web bundle install
-          docker-compose run --rm -e RAILS_ENV=test web bundle exec rake db:setup db:test:prepare elasticsearch:setup
-          docker-compose up -d db elasticsearch redis sshd csv_watch ftp_monitor sidekiq
+          docker load < tmp/docker/docker_images.tar
+
+      - name: Build Docker image
+        if: steps.docker-cache.outputs.cache-hit != 'true'
+        run: |
+          $DOCKER_COMPOSE build
+          docker save cdx_web cdx_csv_watch cdx_ftp_monitor cdx_sidekiq > tmp/docker/docker_images.tar
+
+      - uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('.ruby-version') }}-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gems-${{ hashFiles('.ruby-version') }}-
+            ${{ runner.os }}-gems-
+
+      - name: Install dependencies
+        run: |
+          $DOCKER_COMPOSE run --rm --no-deps -e RAILS_ENV=test web bundle install --jobs 4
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -11,25 +11,68 @@ on:
...
+      - name: Setup databases
+        run: |
+          $DOCKER_COMPOSE up -d db elasticsearch redis
+          $DOCKER_COMPOSE run --rm -e RAILS_ENV=test web bundle exec rake db:setup db:test:prepare elasticsearch:setup
+
+      - name: Start services
+        run: $DOCKER_COMPOSE up -d sshd csv_watch ftp_monitor sidekiq
+
+      - name: Run integration specs (rspec + capybara)
+        run: $DOCKER_COMPOSE run --rm -e RAILS_ENV=test web bundle exec rspec spec/features/*_spec.rb
```
straight-shoota, line 66: Why the file-level wildecard? If we'd start organizing feature specs in folders, they wouldn't be matched.
```suggestion
        run: $DOCKER_COMPOSE run --rm -e RAILS_ENV=test web bundle exec rspec spec/features/
```

ysbaddaden: This is to work around the exclude-pattern configuration!

Which we should remove, yes.

straight-shoota: Yes, please let's do that instead 🙏

straight-shoota:
Just wanted to mention that between caching and parallelizing spec runs, this shaves off 3 minutes from the already improved spec performance of #909. And it halves the total CI runtime compared to master 🚀

.github/workflows/ci.yml
```diff
@@ -11,25 +11,68 @@ on:
     branches:
       - "**:**"
 
+env:
+  DOCKER_COMPOSE: docker-compose -f docker-compose.yml -f docker-compose.ci.yml
```
matiasgarciaisaia, line 15: Not sure if it'd be cleaner to start by `mv docker-compose.ci.yml docker-compose.override.yml` instead of having to remember using the env variable everywhere?

ysbaddaden: We'll just to not forget to repeat the copy in all jobs (except `build`).

.github/workflows/ci.yml
```diff
@@ -11,25 +11,68 @@ on:
...
+
 jobs:
-  test:
+  setup:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: ./.github/actions/docker-cache
+      - uses: ./.github/actions/gems-cache
```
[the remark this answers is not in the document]
ysbaddaden: BTW: maybe `build.sh` could be wrapped as an action, also.

.github/workflows/ci.yml
straight-shoota: Could we just use an *actual* environment variable to configure the compose file? [`COMPOSE_FILE`](https://docs.docker.com/compose/reference/envvars/#compose_file) should do.

.github/workflows/ci.yml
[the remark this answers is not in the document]
ysbaddaden: It's actually quite hard to achieve. There are lots of details to take care of.

It first started with the `-f` option for docker-compose, that must be passed to the each actions... I solved it with a `cp docker-compose.ci.yml docker-compose.override.yml` in the jobs!

But you must also tell the action which docker images to save (one configuration), and we'd like to have the gems-cache action run bundler (so the action is self container), but then it needs to know in which docker-compose service it should run `bundle install`, and we have to pass arbitrary options such as `-e RAILS_ENV=test` :sob: 

In the end, I couldn't get the actions to run. Rake failed to connect to `db` and I reverted everything :sob:

.github/workflows/ci.yml
ysbaddaden: I just copied the file. That's simple.

ysbaddaden: Oh, the `COMPOSE_FILE` is a nice idea. I'll try that!

.github/workflows/ci.yml
```diff
@@ -11,25 +11,66 @@ on:
...
+
+      - name: Setup databases
+        run: |
+          docker-compose up -d db elasticsearch redis
+          docker-compose run --rm -e RAILS_ENV=test web bundle exec rake db:setup db:test:prepare elasticsearch:setup
+
+      #- name: Start services
+      #  run: docker-compose up -d sshd csv_watch ftp_monitor sidekiq
```
straight-shoota, line 61: Can this be removed?
```suggestion
```

PR #1472, opened 2022-02-18
Update to latest patch release Rails 4.2.11

straight-shoota:
This is a minor Rails update to the latest patch release of the 4.2 series. It should be safe to do this and it fixes a number of bugs. That's always a good idea. A bigger Rails upgrade would also be nice, but this is a low hanging fruit.
I noticed one of those bugs while exploring improvements to the assay UI. It's not relevant for the current code base, though (there might be other bugs affecting the current state, though).


ysbaddaden (commented):
I think we don't need most (if any) of the new binstubs, except for `bin/bundle`, and `bin/spring` should be generated by `spring binstub` itself.

Apparently it doesn't boot anymore :sob:
```
Creating cdx_web_run ... done
rake aborted!
NoMethodError: undefined method `last_comment' for #<Rake::Application:0x00000001c389c0>
```

ysbaddaden:
BTW: there is a branch to upgrade to Rails 5 but it's a lot of work (and unfinished). We also have plans to start a dualboot solution, where we can boot with Rails 4 or Rails 5, and have CDx compatible with both versions, so we can regularly work on the upgrade and test it, while still working on features and without having to rebase a branch all-the-time.

straight-shoota:
This is proving to be quite difficult because many gems have dropped support for older Ruby versions. We should upgrade to a newer Ruby version first.

ysbaddaden:
Damn, it looked like a low hanging fruit, but this is going circles :face_with_head_bandage:

diegoliberman:
Is this required for any other PR?  If not, I'd cancel.  Anyhow, we'll have a lot of fun in March continuing the upgrades that Fran had started.  And we'll end up in Rails 5 (minimum).

ysbaddaden:
Let's close for now :sob:

straight-shoota:
Yeah, I tried upgrading to a newer Ruby version, but those that are still supported right now (2.6, 2.7) don't seem to work very well with Rails 4.2. So we'd need to take smaller steps somewhere in between.
It's probably better to just take bigger steps then and upgrade to Rails 5+.

@diegoliberman It's not directly a requirement for any pending PRs. I just noticed a bug that's fixed in the latest patch release while experimenting with https://github.com/arielj/vanilla-nested for polishing the assay attachments UI. If we want to go that route, we can simply patch the bug in our code.

ysbaddaden:
Yes, Rails 4.2 recommends Ruby 2.2, and trying to upgrade Ruby without upgrading Rails to a major version will be impossible (e.g. you must upgrade to Rails 6 to upgrade to Ruby 2.7). We can probably upgrade to the latest Rails 4.2 patch release, but it seems to require to upgrade a few more libraries, and we must be _very_ conservative with them.

Since we also monkey-patch Rails a lot... that probably explains the _very_ high hanging fruit. Let's (sadly) resume the Rails 5 upgrade which already fixed a number of bugs (but not all) and already upgraded to a newer Ruby (I think 2.6).

PR #1494, opened 2022-03-01
Fix: can't invite institution with multiple emails

### LINKED ISSUE #1436: [BUG] Invites - Backend email validation
Add a backend email validation when inviting new users/institutions. Frontend validation fixed in: #1422

ysbaddaden:
fixes #1436

### DIFF
--- a/app/models/pending_institution_invite.rb
+++ b/app/models/pending_institution_invite.rb
@@ -6,6 +6,7 @@ class PendingInstitutionInvite < ActiveRecord::Base
   statuses = %w(pending accepted)
 
   validates_presence_of :invited_user_email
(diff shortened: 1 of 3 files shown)


straight-shoota:
I realize this is *technically* sufficient to fix the issue. But just checking for a single character as email validation doesn't seem right.
Why not do a proper email format validation? We're already doing this here: https://github.com/instedd/cdx/blob/0cb48ff423477cc2968db3144ac90100cbe122b0/app/models/alert_recipient.rb#L10

ysbaddaden:
You're right, this is much better!
I'm always afraid of validating emails, but the regexp is simple enough :+1:

straight-shoota (approved):

PR #1495, opened 2022-03-01
Implement sample transfer workflow with confirming receipt

### LINKED ISSUE #1389: 5- Allow to accept a transferred Sample
- [x] Transferred samples should have an initial **Transfer Status** of `In Transit`.  

### Log of transfers
https://projects.invisionapp.com/d/main#/console/4103589/462385786/preview
- [x] `In Transit` samples don't display the last 4 numbers of the sample ID.
- [x] `In Transit` samples have an action that presents a modal called "Confirm receipt". 
- [x] Ignore the blue dot that appears in the Samples list icon.  This dot is to represent that there are new received samples, but we are not implementing that right now.  Just mark this check box to acknowledge that you won't implement it now.

### Confirm receipt modal
https://projects.invisionapp.com/d/main#/console/4103589/461729947/preview

- [x] At this point, we won't include the Sample ID field.  Just mark this check box to acknowledge that you won't implement it now.
- [x] The modal has the designed look & feel.
- [x] The modal has a cancel action that simply closes the modal and nothing is done.  The same for the top-right X.
- [x] The modal has a confirm action that changes the transfer status of the Sample to `Received`.

### Note
Only 1 Sample at a time can be confirmed.

### Test cases
- [ ] Confirm the receipt of a Sample, and make sure that the sample is now shown normally in the list of Samples.
- [ ] Also, confirm that you can access to that Sample, you can edit the data, you can add assays, notes.

straight-shoota:
Resolves #1389

The confirmation action is currently unrestricted. I'm not sure about the policy requirements for that.

### DIFF
--- a/app/controllers/sample_transfers_controller.rb
+++ b/app/controllers/sample_transfers_controller.rb
@@ -31,12 +31,7 @@ class SampleTransfersController < ApplicationController
   end
 
   def confirm
-    transfer = SampleTransfer.find(params[:sample_transfer_id])
-
-    unless transfer.receiver_institution == @navigation_context.institution
-      render json: { status: "error" }, status: :not_found
-      return
-    end
+    transfer = SampleTransfer.with_receiver(@navigation_context.institution).find(params[:sample_transfer_id])
 
     resource = {
       resource_type: "sample",
--- a/app/controllers/sample_transfers_controller.rb
+++ b/app/controllers/sample_transfers_controller.rb
@@ -13,4 +13,61 @@ class SampleTransfersController < ApplicationController
     @sample_transfers = perform_pagination(@sample_transfers)
     @sample_transfers = @sample_transfers.map { |t| SampleTransferPresenter.new(t, @navigation_context) }
   end
+
+  def create
+    new_owner = Institution.find_by(uuid: params["institution_id"])
+    if new_owner.nil?
+      flash[:notice] = "Destination Institution does not exists"
+    else
+      begin
+        change_ownership(new_owner)
+        flash[:notice] = "All samples has been transferred successfully."
+      rescue
+        flash[:notice] = "Samples transfer failed."
+      end
+    end
+
+    render json: { status: :ok }
+  end
+
+  private
+
+  def change_ownership(new_owner)
+    samples = Sample.joins(:sample_identifiers).where("sample_identifiers.uuid": params["samples"])
+    Sample.transaction do
+      samples.each do |to_transfer|
+        raise "User not authorized for transferring Samples " unless authorize_resource?(to_transfer, UPDATE_SAMPLE)
+
+        transfer = SampleTransfer.new(
+          sample: to_transfer,
+          receiver_institution: new_owner,
+        )
+        transfer.confirm
+        transfer.save!
+
+        unless to_transfer.batch.nil?
+          if params["includes_qc_info"] == "true"
+            qc_info = create_qc_info(to_transfer)
+            to_transfer.qc_info_id = qc_info.id if qc_info
+            to_transfer.old_batch_number = to_transfer.batch.batch_number
+          end
+        end
+
+        to_transfer.batch_id = nil
+        to_transfer.site_id = nil
+        to_transfer.institution = new_owner
+        to_transfer.save!
+      end
+    end
+  end
+
+  def create_qc_info(sample)
+    sample_qc = sample.batch.qc_sample
+    return if sample_qc.nil?
+
+    qc_info = QcInfo.find_or_duplicate_from(sample_qc)
+    qc_info.samples << sample
+    qc_info.save!
+    qc_info
+  end
 end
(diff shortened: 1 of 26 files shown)


app/controllers/samples_controller.rb
```diff
@@ -190,46 +190,8 @@ def bulk_destroy
...
-          receiver_institution: new_owner
-        )
-        transfer.confirm
-        transfer.save!
-
-        if to_transfer.batch_id.nil?
-          to_transfer.site_id = nil
-          to_transfer.institution = new_owner
-          to_transfer.save!
-        end
```
ysbaddaden, line 228: You no longer check for `batch_id.nil?` when transferring samples when confirming reception. Is it on purpose?

app/models/sample.rb
```diff
@@ -80,4 +83,15 @@ def empty_entity?
   def has_entity_id?
     entity_ids.compact.any?
   end
+
+  def transfer_to(new_owner)
```
ysbaddaden, line 87: Perhaps the name could reflect the in-transition state, for example `transferring_to()`? The `transfer_to` feels like it's being done right now, which is no longer the case.

config/routes.rb
```diff
@@ -103,15 +103,16 @@
       post 'reprocess'
     end
   end
-  resources :sample_transfers, only: [:index]
+  resources :sample_transfers, only: [:create, :index] do
+    put "confirm"
```
ysbaddaden, line 107: I usually use `POST` for actions on a resource, and `PUT` to replace a resource (and `PATCH` for updating).

app/controllers/sample_transfers_controller.rb
```diff
@@ -12,4 +12,49 @@ def index
...
+        flash[:error] = "Samples transfer failed."
+        render json: { status: :error }, status: 500
+      end
+    end
+  end
+
+  def confirm
+    transfer = SampleTransfer.find(params[:sample_transfer_id])
+
+    transfer.confirm_and_apply!
```
ysbaddaden, line 35: There is no validation that the current user has any access to the transferred sample, without even talking about the permission to do this action.

Maybe duplicate the `authorize_resource?(to_transfer, UPDATE_SAMPLE)` from the `#create_transfer` method below?

ysbaddaden (commented):
Nothing much to say. It's only missing a basic permission check that the user can access the sample in the `confirm` action.

app/controllers/samples_controller.rb
straight-shoota: Yes, good catch. I wanted to add a comment about that but forgot.
I intentionally omitted that because it's going to be removed in https://github.com/instedd/cdx/issues/1388. I suppose we should merge these PRs in order, so this PR will need an update anyways.

app/models/sample.rb
straight-shoota: Maybe `#start_transfer`?

app/controllers/sample_transfers_controller.rb
straight-shoota: I added this in [f559b3a](https://github.com/instedd/cdx/pull/1495/commits/f559b3a9bb377b3419146ee1c2502a1108e655e8)

ysbaddaden: Oh, that's great! If we could just extract an `#authorize_confirm?(transfer)` method that would be perfect for now.

straight-shoota: Yes, I realized that should better be extracted because we need it again for showing the confirm button.

app/controllers/sample_transfers_controller.rb
```diff
@@ -32,6 +32,24 @@ def create
   def confirm
     transfer = SampleTransfer.find(params[:sample_transfer_id])
 
+    unless transfer.receiver_institution == @navigation_context.institution
```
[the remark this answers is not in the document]
straight-shoota: Sounds good. I was wondering whether there could be a more specific error than just "not found" to tell the user that they're trying to confirm from a context that is not the intended receiver. But that's probably not necessary. This route would only be hit when a specific link is generated and that should only work from the receiver context. We don't need to pay much attention to error handling. For that reason I don't think we need to do anything about the `RecordNotFound` exception. It just returns a 404 status and that's it.

There is a check missing to verify the transfer has not already been confirmed. I'm doing that now. This should result in a proper error message, because it's an error that is likely to occur in normal usage (changing the `confirmed` status may lead to race conditions).

app/presenters/sample_transfer_presenter.rb
```diff
@@ -28,7 +28,16 @@ def other_institution
   end
 
   def can_confirm?
-    true
+    ApplicationController.helpers.authorize_resource?(confirmation_resource, Policy::Actions::UPDATE_SAMPLE, @context.user)
```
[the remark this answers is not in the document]
ysbaddaden: I'd probably have a PolicyAuthorize concern or something like that, that exposes the methods to views with `helper_method`, to try to group things up into a properly named module.

ysbaddaden (approved):
LGTM

PR #1497, opened 2022-03-02
Improve filter layout

### LINKED ISSUE #1380: UX / UI Improvements in the Batches/Samples filters
Lets add some `margin-top` to the filters, in case they drop to the line below.
image

Same here.  
image

Bonus: it would be awesome if we can easily make the 2 date fields drop to the line below together.  So, if there's not enough place for both to fit, both of them drop to the line below.

straight-shoota:
grafik

grafik

Resolves #1380


ysbaddaden (approved):

ysbaddaden:
I believe we can merge?

Issue #1217, opened 2021-06-22
Fixes in Batches


diegoliberman:
- [x] Relabel "batch number" to "Batch ID" in the batch creation form.
- [x] Fix issue where the uniqueness validation is case sensitive.
- [x] Add the column "Date produced" to the batch list.

PR #1498, opened 2022-03-02
Replace custom batch validation with `validate_uniqueness_of`

straight-shoota:
I noticed this complex validation mechanism and I think it can be replaced by `validate_uniqueness_of` with a `scope`. Seems like a nice improvement.

There are no unit tests for the model, but there is a uniqueness spec on the controller which ensures the behaviour is still correct. And I can't see any other relevant difference that would make this validation helper inadequate for this.

Depends on https://github.com/instedd/cdx/issues/1517

### DIFF
--- a/app/models/batch.rb
+++ b/app/models/batch.rb
@@ -30,7 +30,7 @@ class Batch < ApplicationRecord
   validates_numericality_of :volume, greater_than: 0, message: "value must be greater than 0"
   validates_presence_of :batch_number
   validates_presence_of :isolate_name
-  validates_uniqueness_of :batch_number, scope: :isolate_name
+  validates_uniqueness_of :batch_number, scope: :isolate_name, case_sensitive: false
 
   validates_associated :samples, message: "are invalid"
 
--- a/app/models/batch.rb
+++ b/app/models/batch.rb
@@ -28,9 +28,9 @@ class Batch < ApplicationRecord
   validates_presence_of :volume
   validates_presence_of :lab_technician
   validates_numericality_of :volume, greater_than: 0, message: "value must be greater than 0"
-
-  validate :isolate_name_batch_number_combination_create, on: :create
-  validate :isolate_name_batch_number_combination_update, on: :update
+  validates_presence_of :batch_number
+  validates_presence_of :isolate_name
+  validates_uniqueness_of :batch_number, scope: :isolate_name
 
   validates_associated :samples, message: "are invalid"
 
@@ -39,7 +39,7 @@ class Batch < ApplicationRecord
   }
 
   def qc_sample
-    self.samples.select {|sample| sample.is_quality_control?}.first
+    self.samples.select { |sample| sample.is_quality_control? }.first
   end
 
   def has_qc_sample?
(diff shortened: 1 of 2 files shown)


ysbaddaden:
There is just one difference in that the current implementation uses `LOWER` for case-insensitive comparison, whereas `validates_uniqueness_of` won't. I don't know if it has much impact.

diegoliberman:
Is this validation regarding the Batch ID?

straight-shoota:
@diegoliberman Yes, exactly. It makes sure that the combination of Isolate Name and Batch Number is unique. But you can apparently use the same Batch Number for a different Isolate Name.

@ysbaddaden We can configure `validate_uniqueness_of` with `case_sensitive: false`.
This seems a bit odd to me, though. I don't see any handling of case insensitivity anywhere else for those fields. The filters on the index view for example are case sensitive.

gbatiston:
> I think that being a non-IT target audience, they might get confused about case-sensitivity. I'd make it case-insensitive everywhere. @gbatiston what do you think?

I agree, case-insensitive makes sense here

ysbaddaden (approved):

straight-shoota:
It turns out `case_sensitive: false` isn't even necessary.

All the text columns in the database use the collation `utf8_unicode_ci` which is already case-insensitive.

So we don't even need to take any action to make everything case-insensitive everywhere.

straight-shoota:
Ooh, it turns out this is essentially a revert of #1221 (which was later refactored in #1317). 
The original PR doesn't state a reason for this change other than it's a fix. I can only assume it might've been caused by the case-insensitivity concern.
That's not really an issue though, as explained in my previous comment - and demonstrated by the specs.

@devduarte Can you recall the particular reason for the original change?

straight-shoota:
The PR actually references #1217 which names the reason: `Fix issue where the uniqueness validation is case sensitive.` I don't know why this was a problem back then because the code was equivalent to the one in this PR, but it works case-insensitive. Only explanation for that would be that the db collation might have changed in the mean time.

ysbaddaden:
It's possible that there was an issue somewhere in production, and the collation isn't case insensitive there.

ysbaddaden:
We should probably keep [case_sensitive: false](https://github.com/instedd/cdx/pull/1498/commits/be134dac92ed80d3448f254a73d64dee8bdd7f99) then, and double check that it's working in staging.

diegoliberman:
@mmuller had [reported](https://github.com/instedd/cdx/issues/1201#issuecomment-865435420) in #1201 that the validation of isolate name was case-sensitive.  Perhaps this piece of history helps on this PR discussion.

ysbaddaden:
@straight-shoota I don't know about production databases, but the staging db is using `latin1_swedish_ci` by default :confused:

straight-shoota:
Sigh. At least the staging db is case insensitive. But apparently not production.

`latin1` though is not a good idea. I think we should migrate to a sane charset and collation. That would probably be `utf8mb4_unicode_ci`.

diegoliberman:
@straight-shoota is this PR ready to be merged? We're not sure because it was approved, but there were modifications afterwards.

straight-shoota:
Depends on #1517

diegoliberman:
@straight-shoota @ysbaddaden this PR is unblocked now.

ysbaddaden (approved):

Issue #1235, opened 2021-06-30
Upgrade from 4.2 to 5.0


diegoliberman:
### Resources:

- [Rails 5.0 release notes](https://guides.rubyonrails.org/5_0_release_notes.html)
- [Upgrading from Rails 4.2 to Rails 5.0](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-2-to-rails-5-0)

### [Prepare for Rails Upgrade](https://www.fastruby.io/blog/rails/upgrade/prepare-for-rails-upgrade.html)

- [x] Cleanup unused (and underused) gems (see #1511)
- [x] Code Coverage >= 80%, but consider:
  - [ ] adding specs for controllers at 0%
  - [ ] adding specs for controllers < 50%
  - [ ] adding specs for controllers < 80%
- [x] Staging and Production
- [x] Upgrade to latest patch release
- [x] Incompatabilities —[RailsBump](https://railsbump.org/lockfiles/c56eb7d3496ad0baf0a18d10fda215232a1e03a2) doesn't report much
  - [x] `devise_security_extensions` -> [`devise-security`](https://github.com/devise-security/devise-security)
- [x] Dual Boot (#1322)

### [Upgrade from Rails 4.2 to 5.0](https://www.fastruby.io/blog/rails/upgrades/upgrade-rails-from-4-2-to-5-0.html)

- [x] ~~Upgrade to Ruby 2.2~~ we're already on 2.2
- [x] Upgrade gems
- [x] `rails app:update`
- [x] Application Code
  - [x] Models
  - [x] Controllers
- [x] Testing
- [ ] Enable Rails 5.0 on Staging

straight-shoota:
I was wondering if we might toss out some dependencies. This would be great in general, but especially makes upgrading a little easier. For example I noticed `shoulda` in the Gemfile. That doesn't seem to be used at all (haven't verified it yet, though).

Not sure what would be the best way to approach this. I'm still new to the project and have no idea what most of the dependencies are used for. But just trying around a little bit with removing promising deps and running the spec suite could be pretty easy to do. It's a slow process to do that, but doesn't take much human time.

ysbaddaden:
I'm wondering about which Ruby version to use.

[...]

PR #1511, opened 2022-03-07
Upgrade: cleanup before the storm + Rails 4.2.11.3 (latest patch release)

ysbaddaden:
Removes a dozen unused gems, and removes the `decent_exposure` gem that wasn't used much. This is less gems to worry about during the upgrade. There are probably more gems that aren't used much that we could remove. For example RestClient is used for a single HTTP call!

I also tried to group gems into categories (databases, models, views, assets, ...).

Also adds support for code coverage (on demand) using simplecov to understand what features are missing tests, as well as where to focus QA for the upgrade process. We have a good average (80%) but we also have lots of fully untested or under tested controllers, which decreases the overall confidence in the upgrade.

refs #1235


spec/models/device_spec.rb
```diff
@@ -31,9 +31,12 @@ def should_be_valid_with_instition_and_site(device)
...
   context "validations" do
 
-    it { is_expected.to validate_presence_of :device_model }
-    it { is_expected.to validate_presence_of :name }
-    it { is_expected.to validate_presence_of :institution }
+    it do
+      expect(subject.valid?).to be false
+      expect(subject.errors[:name]).to include("can't be blank")
+      expect(subject.errors[:institution]).to include("can't be blank")
+      expect(subject.errors[:device_model]).to include("can't be blank")
```
straight-shoota, line 38: How about adding `validate_presence_of` as a custom spec helper? It's more expressive to use that method than manually checking for error messages. #1491 adds more usages of that helper.

ysbaddaden: I did that at first, but quickly removed it as there was so few occurrences. To be honest, I'd just remove those tests: they'll likely never break, and thus have zero value as tests.

straight-shoota: That's a good point. I've been thinking about that as well. When you have `validate_presence_of :name` in the model and `expect_it.to validate_presence_of :name` in the test, there's not much gained from having pretty much the same code twice. It still gives a safety net for unintentionally messing with the model, though.
It might not just prevent someone removing the validation from the model by mistake (this should be easy to spot in a review), but also side effects from other code changes. For example, if a pre-validation hook automatically assigns a value to `name` a failing validation test would alert that something might be odd.

ysbaddaden: Sorry, I don't get your example :confused:

straight-shoota: I mean when you add something like `before_validation  { this.name = "Default Name" }`. This doesn't technically contradict presence validation but it would be odd. For example, the user would be allowed to leave the name field blank.
This is just an example of a side effect that may have an influence on presence validation. Of course this one is still quite obvious, but there could be more subtle side effects.
All I'm saying is even if the expectation doesn't add much value, it still can provide some safety and peace of mind.

ysbaddaden: I had forgotten, but starting from ActiveRecord 5 the `belongs_to` associations are already required by default; we must specify `required: false` when it's optional. The only usage of `validate_presence_of` goes down to a single occurrence (`Device#name`).

I removed the specs.

ysbaddaden:
I added some backward compatible changes (ApplicationRecord, ApplicationJob, rename `before|after_filter` to `before|after_action`).

That should be good to go.

straight-shoota (approved):

ysbaddaden:
The error that popped in CI was the same: the RSpec version was incompatible with Rake that got updated by Bundler for an unknown reason. Maybe you then tried more upgrades? I chose to merely downgraded Rake back to its previous version, because there was no reason for Bundler to upgrade Rake in the first place.

Issue #1483, opened 2022-02-25
7- Q samples cannot be transferred


gbatiston:
### Details 
 As the QC Sample is the one that contains all the relevant assay information that sets the bases for all sample recipients to run their tests, these should not be allowed to be transferred. 

### Behaviour
- When a user selects Specimen Role as Q, when creating a new sample or batch, there is a message on screen that lets the user know these samples are not transferrable
- When a user transfers samples that have no QC, there is a message that lets the users know there is no QC info available for those samples (#1484)
- When users select samples from a list, the Transfer icon should not be enabled if a QC sample is selected

### Questions

jkicillof:
[Sample detail](https://projects.invisionapp.com/d/main#/console/4103589/113463974/preview)

gbatiston:
@jkicillof how should we prevent users from trying to transfer Q samples?
1. Disabling the Transfer button if the users select Q samples from a list
2. Allowing users to select all samples they want to transfer but displaying an error message if they click on the transfer button when having QC selected
  - In this scenario, I would expect the system to somehow identify which are the problematic samples that are blocking the transfer (either by an error message that lists the sample IDs, or by highlighting the samples on a list, or some other way)

jkicillof:
I go with #1 showing a message on the table title like [this](https://projects.invisionapp.com/d/main#/console/4103589/464738448/preview)

gbatiston:
Thank you, @jkicillof!

@diegoliberman I think we have everything we needed to move forward here, otherwise let me know

PR #1515, opened 2022-03-10
Dont transfer QC samples

devduarte:
Fixes: #1483 

## Before this PR, when selecting a QC sample the send icon does not get disabled

Screen Shot 2022-03-10 at 10 41 00

## After this PR 

Screen Shot 2022-03-10 at 10 42 10

Icon gets disabled and a message is shown to the user.


app/assets/stylesheets/_sprites.scss
```diff
@@ -354,6 +354,9 @@
 .icon-send:before {
   content: "\e95c";
 }
+.icon-error:before {
+  content: "\e917";
+}
```
straight-shoota, line 359: The `icon-error` class is already defined with that content. I presume you want a different icon (with filled background) but that shouldn't shadow the existing name.

app/views/samples/index.haml
```diff
@@ -45,6 +48,7 @@
                 %td= sample.inactivation_method
                 %td= sample.specimen_role_description
                 %td= sample.date_produced_description
+                %td{id: 'sample_is_qc', class: 'nodisplay'}= sample.is_quality_control?
```
straight-shoota, line 51: Does this information need to be expressed in a hidden cell? How about using a simple `data` attribute on the row instead?

app/views/samples/_index_bulk_actions_js.haml
```diff
@@ -36,6 +54,7 @@
       if(checkboxInput.is(':checked')) {
         var uuid = $(sample).children('#sample_uuid').text();
         var isolateName = $(sample).children('#sample_isolate_name').text();
+        debugger;
```
straight-shoota, line 57:
```suggestion
```

app/views/samples/index.haml
```diff
@@ -11,13 +11,16 @@
...
-              .icon-trash.btn-icon
+            %div.table-info
+              = react_component "SampleTransfer", context: @navigation_context.to_hash, institutions: @institutions
+              =button_tag id: 'bulk_transfer', class: 'btn-link', name: 'bulk_action', value: 'transfer', type: 'button' do
+                .icon-send.btn-icon
+              =button_tag id: 'bulk_print', class: 'btn-link', name: 'bulk_action', value: 'print' do
+                .icon-print.btn-icon
+              =button_tag id: 'bulk_destroy', class: 'btn-link', name: 'bulk_action', value: 'destroy', data: { method: "delete", confirm: "You're about to permanently delete the selected samples. This action CANNOT be undone. Are you sure you want to proceed?" } do
+                .icon-trash.btn-icon
+              #notification.icon-error.icon-red.nodisplay
```
straight-shoota, line 22: `#notification` is a pretty generic id that could easily collide. Maybe name it `#table-notification` or something else?

straight-shoota, line 22: `.nodisplay` means `display: hidden`. A better style would be `visibility: hidden` (`.hidden `). Space for such an error message should be reserved (even when invisible) so that it doesn't trigger a layout shift when it shows.

app/views/samples/_index_bulk_actions_js.haml
```diff
@@ -25,6 +26,23 @@
     });
   }
 
+  function toggleBulkTransfer(checkbox) {
+    var row = checkbox.parents('.laboratory-sample-row')
+    var isQc = row.children('#sample_is_qc').text() === 'true'
+    if (checkbox.is(':checked')){
+      if(isQc){
+        $("#bulk_transfer").prop('disabled', isQc);
+        $('#notification').addClass('table-info')
```
straight-shoota, line 35: Can't this class be permanent?

ysbaddaden (approved):

Issue #1426, opened 2022-01-24
Automated tests migration


diegoliberman:
What we have today:
- The integration tests are located in the `specs/features` directory.
--> They use capybara
--> Poltergeist is a gem that runs capybara tests in a headless browser
--> They use a very old browser PanthomJS

- When upgrading rails, there might be some tests that start failing because the browser we use will require to support very old versions (IE < 11).
--> We can fix the issues in those old browsers.
--> We can migrate the integration tests to a modern browser, before continuing the upgrade.
--> We can disable the failing tests until we finish the upgrade and then we migrate the integration tests.

- Migrating integration tests
--> It would require migrating Poltergeist to a driver based on Selenium/WebDriver
--> An option is Cuprite but only run tests on Google Chrome.

Next steps?
1. Define a driver for migrating the integration tests and replacing Poltergeist.
2. Migrate the driver and make sure tests are working as expected.
3. Resume rails migration.

straight-shoota:
I'm getting a JavaScript error in poltergeist while writing integration specs for #1389. It says `SyntaxError: Parse error` but no location or other info.

In the browser, everything works. So I suspect this is probably due to outdated JavaScipt engine.

ysbaddaden:
@straight-shoota Yes, PhantomJS is very outdated Webkit. You maybe introduced an ES5 or ES6 keyword or construct like `let` or `const`.

ysbaddaden:
I tried to upgrade to Selenium::WebDriver. It has been quite a pain, but I finally got Capybara to talk to a Firefox instance (but couldn't get it to communicate with Chrome). Now, the driver is much more restricted and strict than the PhantomJS driver.

Examples:
- We can't access the HTTP status code for the loaded page (which is normal), so some assertions must be rewritten to verify the page contents.
- We can't select an invisible DOM element, this prevents us from attaching a file to upload (e.g. device manifest) because we hide the input by default.

ysbaddaden:
We can just pass a `:visible` to finders: https://stackoverflow.com/a/43919728/199791 and it works :tada:

ysbaddaden:
While investigating why some Cucumber scenarios failed in `alert.feature`, I noticed that some select options are currently _never_ selected because of a PhantomJS bug. I re-enabled the selection, and some specs broke because the user to receive alerts isn't listed in the cdx_select options (oops). I disabled those scenarios.

[...]

PR #1516, opened 2022-03-10
Use Selenium as Capybara driver

ysbaddaden:
PhantomJS is so old that even IE11 feels modern compared to it.

This patch changes the Capybara driver from Poltergeist to Selenium::WebDriver. The change broke a number of things, for example the HTTP status code isn't available in Selenium, but overall I was able to replace most of them with proper tests.

The cucumber scenarios use the `:rack_test` driver by default, which is incredibly faster, and the specs are passing (they don't need or trigger any JS), so I only enabled the javascript driver manually for 2 features (Alert an Incident) that rely on JS. But maybe we should just use Selenium all the time (whatever how slow).

Refs #1426

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,14 @@ services:
       - elasticsearch
     command: /bin/sh -c './bin/rails s -b 0.0.0.0'
 
+  selenium:
+    build:
+      context: .
+      dockerfile: Dockerfile.selenium
+    working_dir: /src
+    volumes:
+      - .:/src # we mount the source to be able to attach fixture files
+
   db:
     image: mysql:5.7
     environment:
--- a/features/step_definitions/alert_steps.rb
+++ b/features/step_definitions/alert_steps.rb
@@ -3,7 +3,7 @@ include CdxPageHelper
 
 Given(/^the user has an account$/) do
   @user = User.make!(password: '12345678', first_name: 'Bob', email: 'zaa1aa@ggg.com')
-  @institution = Institution.make! user_id: @user.id
+  @institution = Institution.make! user_id: @user.id, name: "LABORATORY"
   @user.grant_superadmin_policy
   authenticate(@user)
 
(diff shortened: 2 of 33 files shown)


ysbaddaden:
It's actually super nice that the remote is easy, because we can just start geckodriver on the host, `export SELENIUM_URL=http://<host.ip>:4444 HEADLESS=false` in the container, and then see the integration specs run live.

docker-compose.yml
```diff
@@ -27,6 +27,17 @@ services:
       - elasticsearch
     command: /bin/sh -c './bin/rails s -b 0.0.0.0'
 
+  selenium:
+    build:
+      context: .
+      dockerfile: Dockerfile.selenium
```
straight-shoota, line 33: I prefer the traditional `name` + `ext` scheme for file names: `selenium.dockerfile`

Also wondering whether we could use a pre-built image for selenium. Did you look into that?

docker-compose.yml
```diff
@@ -27,6 +27,17 @@ services:
...
+    working_dir: /src
+    volumes:
+      - .:/src # we mount the source to be able to attach fixture files
+    #ports:
+    #  - "4444:4444"
+    #  - "5900:5900"
```
straight-shoota, line 39: I suppose this can be removed?
```suggestion
```

features/step_definitions/alert_steps.rb
```diff
@@ -36,17 +36,17 @@ def create_entity_device(entity)
...
 
     form.device_errors_value.set 2
-    form.sites.set_exact_multi "Mrs. Terry Goyette"
-    form.devices.set_exact_multi "Mr. Alphonso Witting"
-    form.roles.set_exact_multi "Institution Aric Smith Reader"
-    form.internal_users.set_exact_multi @user.email
+    form.sites.set_exact_multi "LAB SITE"
+    form.devices.set_exact_multi "INTERNAL DEVICE"
+    form.roles.set_exact_multi "LABORATORY"
+    # form.internal_users.set_exact_multi @user.email # FIXME: not listed!
```
straight-shoota, line 42: Just wondering: Instead of using fixed the names, we could store the models in the instance and refer to the values as `@site.name` etc.
That way we'd have faker provide different test values.

However, it seems the previous implementation also used fixed strings, but they were not specified at the creation... how did that work? Are these specs running with a fixed seed?

straight-shoota:
The readme also needs an update. It mentions PhantomJS and Poltergeist.

docker-compose.yml
ysbaddaden: Yeah, they weight 1GB with lots of useless junk, versus 380MB for a custom Alpine image :shrug:

features/step_definitions/alert_steps.rb
ysbaddaden: Easy: they didn't work :shrug:

straight-shoota (approved):

Issue #1510, opened 2022-03-07
UUID presentation


straight-shoota:
UUIDs are quite a bunch of characters. Displaying them on the page is sometimes a bit of a challenge, especially when space is rare, like in tables. The exact UUID value bears no real information, it's only use case is for identifying objects. That's somewhat restricted when part of the UUID is chopped off because it doesn't fit in the table column for example.

![grafik](https://user-images.githubusercontent.com/466378/157053710-7cfb6cd4-5f72-4b0f-9b59-51d52f185870.png)

Some of the design previous present UUIDs as five groups of four uppercase characters (https://projects.invisionapp.com/d/main#/console/4103589/462385786/preview). That's less than the 36 characters of a full UUID and thus it takes less space in the preview than the real full UUID. And in other previews as well as in the actual app, UUIDs use lowercase characters.
I'm not sure if these other designs are just a mistake/unspecific or if they represent some idea for usability improvement.

Technically, it should be possible to limit UUIDs to a part of themselves. For example, we could always chop off the last 8 characters. That should still have plenty of space for avoiding collisions and would make it easier for users because they have to deal with less characters.

As an example for some practical effects:  #1389 asks that the last four characters of an unconfirmed UUID should be hidden. But this is pretty much without effect because the end of the UUID is always chopped off on reasonably wide viewports.

![grafik](https://user-images.githubusercontent.com/466378/157053591-7d5f80db-f604-4ed3-9a3b-185dc8876e38.png)
![grafik](https://user-images.githubusercontent.com/466378/157053801-f269f4ce-ccbc-4ba6-81b4-4b545223bc89.png)

I'm wondering what can be done to improve UUID presentation. I think it doesn't make much sense to almost always have only a chopped off part of UUID presented in tables. If there's no other solution, we should at least add a tooltip or some other mechanism to reveal the full UUID when parts are hidden overflow. The same applies to other values in table cells, really. So maybe it's not even a specific problem about UUIDs and about table cells in general.

nditada:
IMO, the best way to display guids is to show the first few and the last few chars with an ellipsis in between, that's all you need to verify you have the right sample. Clicking or hovering should display the rest.

PR #1521, opened 2022-03-11
Fix/1510 UUID presentation

leandroradusky:
Until now UUID for samples are shown with right ellipses when text overlaps:

image

While as discussed in #1510 should be better to apply ellipses on the middle of the text because the important characters are on the beginning and the end of the UUID. 

The middle-cropping of text with ellipsis is not implemented in CSS (there is proposals: https://www.w3.org/wiki/Text-overflow_middle_cropping) 

I've been checking proposed solutions around, and the less esoteric that I've found is what I implemented:

- Two spans with the td repeating it and one with right ellipsis and the other with left ellipsis.
- Title with the whole uuid to display it on hover.

The result is the following:

image

Questions that arise before saving the PR:

1. I solved this in the samples table but I should go along the whole project to search where to apply it? 
2. This middle ellipsis behavior can be encapsulated to be applied when needed but first I want to check if you consider its ok. Tips?


leandroradusky:
> What if we just take the first and last 4 chars, and join them with the ellipsis in the middle?

The drawback that I found to this solution is that if the field overflow then you will have abcd...wx... or abcd...w... then you will have middle and end ellipsis or a double ellipsis and this could be confusing? The solution you propose will be clearer in the code since won't be two spans nor styles needed. I do as you consider :)

diegoliberman:
If we go with the middle ellipsis approach, then we shouldn't allow a right/left ellipsis.  If the text doesn't fit, let it wrap.  Ideally, let it wrap at the beginning and/or end of the ellipsis.

PS: Alert --> I don't know if Rails has a feature similar to "virtual fields", you should research about it.

straight-shoota:
What @diegoliberman said 💯
Explicitly restricting which characters are displayed so that it results in a fixed width sounds good. We should aim to keep the format consistent and expectable. That makes it easier to read. And really, a couple of characters from each end should be enough. I'd go for 4 or 5.

We don't need a virtual attribute. This looks like a perfect use case for a simple view helper. @leandroradusky I'm not sure how familiar you are with Rails, let me know if you could use some pointers. It's not hard at all, but it might help getting a hint where to look.

ysbaddaden:
You may add `white-space: nowrap` to the table column, so that it will never wrap.

Since you're not familiar with Ruby nor Rails, here is a detail on @straight-shoota's proposition: you can create a method in a helper module under `app/helpers`. All the methods there will be available in your view. For example a `SampleHelper#short_sample_uuid(sample)` method would be nice.

straight-shoota:
> For example a SampleHelper#short_sample_uuid(sample) method would be nice.

I believe this should actually be a generic view helper (i.e. `ApplicationHelper#format_uuid`). UUIDs are not just used by samples.

leandroradusky:
Thanks a lot for the help! I started some minutes ago, I implemented it following @straight-shoota suggestion, If you confirm that like this is OK I move forward to apply it in all places where its needed.

app/helpers/application_helper.rb
```diff
@@ -201,4 +201,8 @@ def filters_params
     end
     filters_params
   end
+
+  def format_uuid(uuid)
+    uuid[0,4]+"…"+uuid[uuid.length-4,uuid.length]
```
straight-shoota, line 206: This is a bit more concise and idiomatic:

```suggestion
    "#{uuid[0,4]}…#{uuid[-4,4]}"
```

app/views/samples/index.haml
```diff
@@ -36,7 +36,7 @@
                 %td
                   =check_box_tag 'sample_ids[]', sample.id, false, { id: "sample_ids.#{sample.id}" }
                   %label.row{for: "sample_ids.#{sample.id}"}
-                %td= sample.uuid
+                %td{title:sample.uuid}= format_uuid(sample.uuid)
```
straight-shoota, line 39: I think it makes sense to extract the combination of short formatted UUID with full value in the title into another helper method. This is a common pattern. There are probably more locations where a short formatted UUID should be used, and all of them ought to include the full version.
Extracting that into a helper avoids forgetting to add the title on every use. And it allows to easily enhance the component. For example, a feature to copy the full UUID into the clipboard could be useful.

leandroradusky:
I created a new helper following @straight-shoota comments. 

**Regarding uuid shown in tables**

I found no other places where its shown and this helper should be called.

Since the search in the code for uuid was too big to go through I went to the database. These are the tables with uuid as column, and I proceeded to check in these tables when we show them in tables:

`SELECT TABLE_NAME  FROM INFORMATION_SCHEMA.COLUMNS  WHERE TABLE_SCHEMA = 'cdp_test' AND COLUMN_NAME='uuid';`

And I got: batches, devices, encounters, institutions, patients, sample_identifiers, sites & test_results

Then what I did is to search for uuid in the views to see where are displayed for each of these ones listed. I also navigated to all the tabs in the app seeing if there are uuids in the inner tables shown and there are (I found) none.

ysbaddaden (approved):

ysbaddaden:
There are errors in the test suite:

```
  1) SamplesController index should list samples if can read
     Failure/Error: get :index
     ActionView::Template::Error:
       undefined method `[]' for nil:NilClass
     # ./app/helpers/application_helper.rb:210:in `format_uuid'
     # ./app/helpers/application_helper.rb:206:in `short_uuid_with_title'
```

This is caused by `Sample#uuid` being `nil` for some reason. You can run the failing tests locally:

```console
$ docker-compose run --rm web bash
# bin/rspec spec/controllers/samples_controller_spec.rb

**Comment by leandroradusky** [comment]
Now the issue is solved, the tests passed and the changes are approved. 

Before this PR the samples table looked like this:

image

Now it will look as follows:

image

File: app/views/samples/index.haml [hunk 3]
```diff
@@ -36,7 +36,7 @@
                 %td
                   =check_box_tag 'sample_ids[]', sample.id, false, { id: "sample_ids.#{sample.id}" }
                   %label.row{for: "sample_ids.#{sample.id}"}
-                %td= sample.uuid
+                = short_uuid_with_title :td, sample.uuid
```
straight-shoota, line 39: I don't think it's a good idea to merge the UUID element with the enclosing table cell. This doesn't offer much flexibility and could easily lead to conflicts between styles that apply to a UUID display and the enclosing element. They're separate concerns. As mentioned before, I think `<abbr>` would be the best element type for short UUIDs. And that could be hard-coded into `short_uuid_with_title`.

leandroradusky: I see what you mean. I'm doing the following now:

`%td= short_uuid_with_title :span, sample.uuid`

And it looks exactly as before_

image

While for `:abbr` an underline appears:

image

I should proceed with `<span>` then?

straight-shoota: I would still use `abbr` because it's the correct semantics. We can easily adjust the visuals for `abbr`.

leandroradusky: I commited it using abbr. I let there the html tag on the helper as a parameter for flexibility... Just to be sure it makes sense: should be always abbr and remove the parameter name in the helper:

In other words, now we have this:
```
def short_uuid_with_title(name, uuid)
    content_tag(name, format_uuid(uuid), title: uuid)
  end
```

We should change it to this:
def short_uuid_with_title(uuid)
    content_tag(:abbr, format_uuid(uuid), title: uuid)
  end
```
?

Comment (straight-shoota) [reply app/views/samples/index.haml]: Yes, I'd always use `abbr`. The other alternative is not really flexibly, because only the element name can be customized, but not its attributes. That would even require more effort, but it's simply not necessary nor helpfule.
The combination `abbr` + short UUID + title is a functional unit. There's no need for any other element type. It can easily be wrapped in anything else (like `td`).

File: app/helpers/application_helper.rb [hunk 4]
```diff
@@ -201,4 +201,12 @@ def filters_params
     end
     filters_params
   end
+
+  def short_uuid_with_title(uuid)
+    content_tag(:abbr, format_uuid(uuid), title: uuid)
+  end
+  
+  def format_uuid(uuid)
+    uuid.nil? ? "" : "#{uuid[0,4]}…#{uuid[-4,4]}"
```
straight-shoota, line 210: I believe the nil check should be moved to `short_uuid_with_title`. If the UUID is missing, there shouldn't be any output at all. The way it is now, a missing UUID would produce `<abbr title=""></abbr>` which is no good.

straight-shoota (approved):

straight-shoota:
This needs a rebase in order to solve the merge conflict.

PR #1532, opened 2022-03-14
Add note about missing confirmation link for sample transfer

straight-shoota:
Since there was some confusion about what "Unconfirmed" means in the sample transfer list, I added a help icon next to it with a tooltip message.

grafik

@jkicillof WDYT?


ysbaddaden:
Your screenshot is weird. This is the receiver view, so how can I see `Sent on` messages, which are meant for the sender?

straight-shoota:
Ignore their contents :P The top two lines are only there for context.

diegoliberman:
> Your screenshot is weird. This is the receiver view, so how can I see `Sent on` messages, which are meant for the sender?

The receiver could also be a sender and have the "sent" logs in their view. Am I missing something?

ysbaddaden (approved):

Issue #1484, opened 2022-02-25
88 - "Include QC Info" checkbox is disabled for batches with no QC samples in them


gbatiston:
### Details
 When a user transfers samples from a batch that has no QC sample determined, the checkbox to "Include QC Info" is hidden, and we show a proper message in its place, as [designed](https://projects.invisionapp.com/d/main#/console/4103589/464604107/preview). 

![image](https://user-images.githubusercontent.com/25181052/156381995-bba1b6a8-6f7d-4d46-bf24-e6ec50d8c0d5.png)
![image](https://user-images.githubusercontent.com/25181052/156382253-a75cdcae-4a35-49e2-9aa7-2764896e77a8.png)

diegoliberman:
@gbatiston this has already been done as part of #1287.  The link shows only when there's a link to a QC Info.  @devduarte can you confirm?

gbatiston:
@diegoliberman am I mistaken in thinking there was a comment from you here?

jkicillof:
[Transfer sample](https://projects.invisionapp.com/d/main#/console/4103589/464604107/preview)

diegoliberman:
> @diegoliberman am I mistaken in thinking there was a comment from you here?

You are correct, but I deleted it because I was saying nonsenses :P

devduarte:
@diegoliberman @gbatiston What should happen with the checkbox when I want to transfer two samples from two different batches, one having a QC Sample and the other one doesn't have a QC Sample?
I thought about showing both things and tweaking the info description to be "There is no Quality Control (QC) info available for some of these samples". If the user checks the checkbox, the QcInfo will be available for the samples that actually have a QC sample in their batches. On the other hand, this might be misleading for the user.

ysbaddaden:
Some ideas out my brain:
- specify "[ ] transfer QC info (when available) in the checkbox?
- visually mark the samples without QC?

diegoliberman:
@devduarte 

1. If there's no QC for none of the samples, we show the message as design.
2. If all of them have a QC sample, we don't show the message and display the check box as it is.
3. If some have and some doesn't, show the check box and show a new info message below: "There is no Quality Control (QC) info available for N samples".   N should be dynamic. 
4. I like the idea of visually marking the samples without QC, but for now let's not do it.  I'll create a separated ticket for that.

PR #1539, opened 2022-03-15
Show no qc info warning and/or checkbox based selected samples

devduarte:
Fixes: #1484 

## All selected samples have a QC Sample from their batches, only show "include qc info checkbox"

Screen Shot 2022-03-14 at 23 19 52

Screen Shot 2022-03-14 at 23 20 16

## All selected samples don't have a QC Sample from their batches, only show warning

Screen Shot 2022-03-14 at 23 21 55

Screen Shot 2022-03-14 at 23 22 24

## A combination of both cases shows checkbox and warning (with different messages)

Screen Shot 2022-03-14 at 23 23 56

### DIFF
--- a/app/assets/javascripts/components/sample_transfer.js.jsx
+++ b/app/assets/javascripts/components/sample_transfer.js.jsx
@@ -38,19 +38,19 @@ var SampleTransferModal = React.createClass({
   },
 
   showQcWarningCheckbox: function(selectedSamples) {
-    const hasQc = selectedSamples.filter((sample) => sample.existsQcReference === true).length
-    if(selectedSamples.length > 0 && selectedSamples.length === hasQc) {
+    const haveQc = selectedSamples.filter((sample) => sample.existsQcReference === true).length
+    if(selectedSamples.length > 0 && selectedSamples.length === haveQc) {
       return this.includeQcInfoCheckbox()
     }
     else {
-      if (hasQc === 0) {
+      if (haveQc === 0) {
         return this.qcInfoMessage()
       }
       else {
         return (
           <span>
             {this.includeQcInfoCheckbox()}
-            {this.qcInfoMessage(selectedSamples.length - hasQc)}
+            {this.qcInfoMessage(selectedSamples.length - haveQc)}
           </span>
         )
       }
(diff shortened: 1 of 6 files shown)


straight-shoota:
So it's like a UI hack? It definitely doesn't look like a deliberate choice. So I'd suggest to add a comment about it.

Or maybe better, implement this in a different way that makes it more explicit (with margin assignments for example).

devduarte:
> So it's like a UI hack? It definitely doesn't look like a deliberate choice. So I'd suggest to add a comment about it.
> 
> Or maybe better, implement this in a different way that makes it more explicit (with margin assignments for example).

Oh, ok! So the idea would be to explicitly have those classes (and divs) and just fix the separation overwriting some margins/paddings, right?

app/models/batch.rb
```diff
@@ -37,6 +37,10 @@ def qc_sample
     self.samples.select {|sample| sample.is_quality_control?}.first
   end
 
+  def has_qc_sample?
+    self.qc_sample.present?
+  end
```
ysbaddaden, line 42: We really need JSON queries, because this is a super expensive call :fearful:

app/assets/javascripts/components/sample_transfer.js.jsx
```diff
@@ -96,14 +108,35 @@ var SampleTransferModal = React.createClass({
     )
   },
 
+  qcInfoMessage: function(missingQuantity, selectedSamples) {
+    let infoMessage = "There is no Quality Control (QC) info available for these samples"
+    if(missingQuantity > 0 && missingQuantity != selectedSamples.length) {
+      infoMessage = `There is no Quality Control (QC) info available for ${missingQuantity} ${missingQuantity === 1? 'sample' : 'samples'}`
+    }
```
ysbaddaden, line 115: Just a very minor readability improvement. I usually don't like multiline ternary operator, but here it would read well:

```js

    const infoMessage = (missingQuantity > 0 && missingQuantity != selectedSamples.length)
      ? `There is no Quality Control (QC) info available for ${missingQuantity} ${missingQuantity === 1 ? 'sample' : 'samples'}`
      : "There is no Quality Control (QC) info available for these samples"
```

ysbaddaden (approved):

app/assets/javascripts/components/sample_transfer.js.jsx
straight-shoota: An `if` with `else` branch would also do... But either is better than a conditional override of the previously assigned value.

ysbaddaden: Yes, in a language where everything is an expression... but this is JS, so I resolve myself to see a multiline ternary!

straight-shoota (approved):

Issue #1525, opened 2022-03-14
Transfer sample action is enabled even when selecting QC samples


diegoliberman:
In order to reproduce this bug, these exact steps must be followed:

1. Go to he samples view
2. Select a QC sample
3. Result: the transfer action is disabled and the error message displayed.  So far, so good.
4. Select a non QC sample
5. Result: the transfer action gets enabled.  The error message remains.  Wrong and inconsistent.
6. Select another non QC sample.  Same result than step 3.  Wrong.
7. Remove the selection of a non-QC sample.  
8. The message disappears and the action gets enabled.  Even when there's still a QC sample select.   Wrong.

I didn't continue testing, but let's try to test everything that comes to our mind when fixing this.

straight-shoota:
This should've been implemented in #1515. The screenshot there indicates that it works. So apparently, something's gone wrong.

diegoliberman:
> This should've been implemented in #1515. The screenshot there indicates that it works. So apparently, something's gone wrong.

@straight-shoota yes, it works but there are bugs in several test cases.

PR #1540, opened 2022-03-15
Disable send icon according to selected qc samples

devduarte:
Fixes: #1525

Before this fix we were disabling (or not) the icon based on the last row change. This fix checks for all the selected sample list every-time a checkbox change event happens.

Note: branched from #1484 because I'm reusing a modified function from that branch.

### DIFF
--- a/app/views/samples/_index_bulk_actions_js.haml
+++ b/app/views/samples/_index_bulk_actions_js.haml
@@ -2,7 +2,7 @@
   (function() {
     allCheckboxes().on('change', function(e) {
       toggleBulkActions();
-      toggleBulkTransfer($(e.target))
+      toggleBulkTransfer()
     });
 
     // initial bulk actions state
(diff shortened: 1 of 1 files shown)


straight-shoota (approved):

app/views/samples/_index_bulk_actions_js.haml
```diff
@@ -26,14 +26,12 @@
     });
   }
 
-  function toggleBulkTransfer(checkbox) {
-    var row = checkbox.parents('.laboratory-sample-row')
-    if (checkbox.is(':checked')){
-      if(row.data().isQc){
+  function toggleBulkTransfer() {
+    var areQc = selectedSamplesIds().filter((sample) => sample.isQc === true).length
```
ysbaddaden, line 30: You can just ask whether the collection contains a QC sample with [Array.prototype.some()](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Array/some). This has two benefits: 1. you don't allocate another array and 2. the iteration will stop as soon as a QC sample is found.

```javascript
if (selectedSamplesId().some(sample => sample.isQc)) {
  // at least one selected sample is a QC
}
```

ysbaddaden (changes requested):
Just a tiny optimization, then please merge!

ysbaddaden (approved):

Issue #1509, opened 2022-03-07
Form fields are slightly offset


straight-shoota:
Baseline (vertical alignment):

![grafik](https://user-images.githubusercontent.com/466378/157044643-72acaca2-c00c-4a22-b4d8-c14bd07bfba1.png)

I think the vertical alignment is off on all fields.

Horizontal alignment:
![grafik](https://user-images.githubusercontent.com/466378/157044487-2b176ebb-73a7-4a90-93c9-113b2324d64a.png)

This applies only to some fields. As you can see, the CdxSelect box is horizontally aligned with the UUID (which is not in a form field).

leandroradusky:
Here I think is better if I ask since don't dare to apply my criteria and moving forward:

![image](https://user-images.githubusercontent.com/13782680/158225314-08b3da1e-457e-4a2b-b902-4387996db092.png)

Here the fields on the form are highlighted in blue (as happens when you input a value) and we can see the "boxes" containing them.

Respect to the horizontal alignment is quite clear that the text should start in the same pixel for all the types of fields. But respect to the vertical, I should make the blue boxes smaller in order to accomodate the text at the same level? or I should keep the box size and displace it down to have the text at the same level?

diegoliberman:
I guess that the Input element should have the same height that the label, and both should align its content vertically in the middle.  

I see that the label has a height of 40px, whereas the input is 30px.  I guess this could be fixed by increasing the inputs' height to 40px.  

![image](https://user-images.githubusercontent.com/25181052/158240156-43e06374-87c3-45cd-b369-e31019aed398.png)
![image](https://user-images.githubusercontent.com/25181052/158240235-dbcb4f11-46d4-4078-83c0-66a348832a5b.png)

I think the result is better by reducing the label height to 30px:
![image](https://user-images.githubusercontent.com/25181052/158240689-8a8f7f84-af45-483a-b4d2-17493f6a5fc0.png)
![image](https://user-images.githubusercontent.com/25181052/158240725-b9a569e6-5536-4add-82c6-bf90f0819adc.png)

Because when assigning a height of 40px to the input, the bottom border line gets too far away from the actual input's text.

The [design](https://projects.invisionapp.com/d/main#/console/4103589/113463974/preview) might help.

straight-shoota:
The total height of the individual elements actually doesn't matter that much for this. It's the line height that counts, as well as the paddings (and margins).

[...]

PR #1543, opened 2022-03-15
form fields properly shown in samples

leandroradusky:
Solves #1509: form fields (just for samples, see below) are now accommodated according to the mockup.

[Mockup](https://projects.invisionapp.com/d/main/#/console/4103589/113463974/preview):

image

Sample form now looks like this:

image

This is functional but there are things to iterate (@straight-shoota? @devduarte? @ysbaddaden?):

- To make all the fields to have the same line height I've created the **display** class. Cannot apply this property in the **col** class, but then this will have to be implemented in all the forms? Or there is a better solution?
- The **display** name is not a good name, I think of form-col?, col-form?... Just want to check if there is any naming convention for this. 
- While for all the other elements to assign the display class works, if I do it in the cdx_select breaks it... please note that all the "cols" have display class except for _specimen role_ and _inactivation method_. This feels incorrect but I need a hint on how to solve this.

### DIFF
--- a/app/assets/stylesheets/_forms.scss
+++ b/app/assets/stylesheets/_forms.scss
@@ -90,7 +90,8 @@ label {
 
 .form-field {
   label {
-    line-height: 40px;
+    line-height: 1.6;
+    margin-top: 9px;
   }
 
   .Select-control {
--- a/app/assets/stylesheets/_forms.scss
+++ b/app/assets/stylesheets/_forms.scss
@@ -88,6 +88,21 @@ label {
   }
 }
 
+.form-field {
+  label {
+    line-height: 40px;
+  }
+
+  .Select-control {
+    padding-bottom: 6px;
+    height: 34px;
+  }
+
+  input:not(.input-small) {
+    margin-top: 5px;
+  }
+}
+
 // Checkboxes ---------------------------------------------------------//
 
 input[type="checkbox"] {
--- a/app/assets/stylesheets/_forms.scss
+++ b/app/assets/stylesheets/_forms.scss
@@ -98,6 +98,10 @@ label {
     padding-bottom: 6px;
     height: 34px;
   }
+  
+  .value{
+    margin-left: 5px;
+  }
 
   input:not(.input-small) {
     margin-top: 5px;
--- a/app/assets/stylesheets/_grid.scss
+++ b/app/assets/stylesheets/_grid.scss
@@ -11,21 +11,6 @@
   &.center {
     @include justify-content(center);
   }
-
-  .form-field {
-    label {
-      line-height: 40px;
-    }
-
-  .Select-control {
-      padding-bottom: 6px;
-      height: 34px;
-    }
-
-    input:not(.input-small) {
-      margin-top: 5px;
-    }
-  }
 }
 
 /* default for even distribution */
--- a/app/assets/stylesheets/_grid.scss
+++ b/app/assets/stylesheets/_grid.scss
@@ -108,8 +108,16 @@
     @include flex(0 0 1000px);
   }
 
-  &.display{
+  &.form-field-label label{
     line-height: 40px;
+  }
+
+  &.form-field-value div.value{
+    margin-top: -1px;
+  }
+
+  &.form-field-value input:not(.input-small){
+    margin-top: 5px;
   }
 }
 
--- a/app/assets/stylesheets/_grid.scss
+++ b/app/assets/stylesheets/_grid.scss
@@ -108,7 +108,9 @@
     @include flex(0 0 1000px);
   }
 
-
+  &.display{
+    line-height: 40px;
+  }
 }
 
 
--- a/app/assets/stylesheets/_grid.scss
+++ b/app/assets/stylesheets/_grid.scss
@@ -11,6 +11,21 @@
   &.center {
     @include justify-content(center);
   }
+
+  .form-field {
+    label {
+      line-height: 40px;
+    }
+
+  .Select-control {
+      padding-bottom: 6px;
+      height: 34px;
+    }
+
+    input:not(.input-small) {
+      margin-top: 5px;
+    }
+  }
 }
 
 /* default for even distribution */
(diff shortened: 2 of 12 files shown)


straight-shoota:
> To make all the fields to have the same line height I've created the `display` class. Cannot apply this property in the col class, but then this will have to be implemented in all the forms? Or there is a better solution?

Yeah, such a class seems like a good idea and I don't think there is any other reasonable solution. We need to indicate that these columns (or rather cells in the form layout) are form field labels and values. That's a common pattern actually.
The class name should be more specific, though. I'd suggest `.form-field-label` and `.form-field-value`. Separating both makes sense because they have different functions. Even if it doesn't make a difference right now.

straight-shoota:
> While for all the other elements to assign the display class works, if I do it in the cdx_select breaks it... please note that all the "cols" have display class except for specimen role and inactivation method. This feels incorrect but I need a hint on how to solve this.

I'm sorry, apparently I have lead you wrong with my comment in https://github.com/instedd/cdx/issues/1509#issuecomment-1067178433

I didn't mean to fix the actual `line-height` CSS property. It's about the perceived line height, which includes margins and paddings of inner elements and maybe even other spaces.

The solution is not to assign a fixed `line-height` everywhere (40px would be way to large anyways), but to adjust the margins and paddings of the elements inside the `form-field-*` to match up.

straight-shoota:
An alternative approach could be to try `align-items: baseline;` for the form rows. This should technically do the right thing, but it is a bit rough for some fields (assays and notes for example). This should be fixable, though.
Going this way could save a lot on fiddling with details of every single form field type.

ysbaddaden:
The problem is that forms use the grid directly, instead of using it through a component abstraction, so we need CSS fixes to adjust things. Oh well, most CSS frameworks do just that :shrug:

Fixing the `line-height` is fine. This is what Foundation 5 does for example (with a `label.inline` selector), but yeah, `.display` isn't very explicit.

I see you just changed to `.form-field-x` which is nicer, but maybe wa can simplify even more with a mere `.form-field`?

```haml
.row.form-field
  .col
    label
  .col
    .value
      input
```
```scss
.form-field {
  label {
  }

  .value {
  }

  input:not(.input-small){
    margin-top: 5px;
  }
}
```

leandroradusky:
I'm recommitting with new doubts, maybe I should have a sync with Jony to clarify stuff.

First, `align-items: baseline` didn't worked, then I moved forward with the other proposed solution.

I arribe to this, which is super close, almost there:

image

As you can see, all is good except for the margin-bottom of the select (measured with "blue ruler") of the specimen role. But... that works in the case of the inactivation method, which is exactly the same component in term of classes.

I've been playing with 

Line height:
Line-height



Padding-top:
padding-top



Padding bottom <- **This seems the right property, but no negative values are allowed and no combinations with other of the ones I'm touching arrive to a solution**
padding-bottom



& Margin bottom:
margin-bottom

And combinations of them and there is not a way to "make the grey line go closer to the text". And the fact that it works in inactivation method is quite confusing. If no hints here I will go to Jony.

straight-shoota:
An obvious difference between *Inactivation Method*  and *Specimen Role* is that the latter has the `input-x-large` class.
This "only" assigns the width of the field and I don't observe a direct effect on the height alignment, though. 

So then the only other difference left is the number of items and size of the selection options. But I can't make out any direct effect of them either. The element for the dropdown view isn't even mounted in the DOM when it's not visible.

diegoliberman:
@leandroradusky let's find a time we can meet and pair this issue.

leandroradusky:
Suggestions by julien applied, problem with react select solved. Styles applied to all forms I could navigate to.

As a sample, the samples form:

image

All the **_form.haml** files are modified in order to incorporate these changes. 

**Some remarks regarding other files modified:**

**alert_category_select** & **address** selects had to be modified in the render function.

Regarding sites, this is how it looks on staging:

Sites_stg

As you can see, the first three field labels are placed displaced to the right.

To modify it according to the [mockup](https://projects.invisionapp.com/d/main/#/console/4103589/104839794/preview) I modified as well the structure of the page. Now it looks as follows:

Sites

app/assets/stylesheets/_grid.scss
```diff
@@ -11,6 +11,21 @@
...
+
+  .Select-control {
+      padding-bottom: 6px;
+      height: 34px;
+    }
+
+    input:not(.input-small) {
+      margin-top: 5px;
+    }
+  }
```
ysbaddaden, line 28: These style don't _declare the grid_ but how to style form inputs when they are _within a grid_, so I think they'd fit better in a stylesheet related to forms.

Maybe `_forms.scss`? But `.Select-control` is defined in `_react_select.scss`, so I'm not sure. I'd love your input @devduarte @straight-shoota!

leandroradusky: I moved them to `_forms.scss` and let the `.Select-control` specific style for forms in there since they have to be form-fields **and** Select-control. If you think that is not ok I change it to` _react_select.scss`.

app/assets/stylesheets/_forms.scss
```diff
@@ -88,6 +88,21 @@ label {
   }
 }
 
+.form-field {
+  label {
+    line-height: 40px;
```
straight-shoota, line 93: As I said in https://github.com/instedd/cdx/pull/1543#issuecomment-1069044354, fixing `line-height` to an absurdly large value is not a proper solution. It breaks as soon as the line breaks:
grafik

`line-height` controls the distance between the baselines of consecutive lines of texts. It's usually just a multiplier. In this case, it should be something around 1.6.
We actually want to control the distance from the top of the cell to the top of the text. That's `padding` or `margin`.

In combination with the adjusted `line-height`, a top margin of 9px should work well:
```suggestion
    line-height: 1.6;
    margin-top: 9px;
```
grafik

straight-shoota:
I think the non-editable fields UUID and Batch ID should have the same alignment as editable fields. After all, they are field values as well, just that you can't change them.
```css
.form-field .value {
  margin-left: 5px;
}
```

leandroradusky:
> I think the non-editable fields UUID and Batch ID should have the same alignment as editable fields. After all, they are field values as well, just that you can't change them.
> 
> ```css
> .form-field .value {
>   margin-left: 5px;
> }
> ```

I was following the mockups, and when "everything is blank" I agree with you that it looks better, but when "inputs turn blue" because of a modification the UUID fields could look not aligned... 

image

Maybe to confirm with Die/jony in the standup.

diegoliberman:
I asked Johny about the non-editable fields alignment in the design page: https://projects.invisionapp.com/d/main/#/console/4103589/113463974/comments/147827603

ysbaddaden (approved):

PR #1545, opened 2022-03-16
Show `old_batch_number` for transferred samples

### LINKED ISSUE #1537: Show batch ID for samples that were transferred
If I transfer a sample that belongs to a batch, the recipient should be able to see the batch id field in the sample form.

straight-shoota:
Fixes #1537

### DIFF
--- a/app/models/sample_form.rb
+++ b/app/models/sample_form.rb
@@ -58,6 +58,10 @@ class SampleForm
       end
   end
 
+  def batch_number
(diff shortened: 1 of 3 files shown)


ysbaddaden (approved):
TBH I'm not sure this is the kind of bugfix that needs a dedicated test.

straight-shoota:
I like tests 🤷 

But yeah, such isolated it's not super useful.

Issue #1547, opened 2022-03-18
Transferring sample fails since UUID shows with ellipsis


diegoliberman:
When transferring samples, the UUID is obtained from the UI element.  As we are now showing the UUID with middle ellipsis (#1510), the code fails to find the right sample that was selected for transfer.

Issue #1529, opened 2022-03-14
Show error message in sample transfer modal when mandatory fields are missing 


diegoliberman:
Right now, the only mandatory field is the Institutions.  If the user presses the transfer button without selecting an institution, today, nothing happens.  We should show an error message.

<img width="456" alt="image" src="https://user-images.githubusercontent.com/25181052/158157317-7cd831aa-8005-46be-96f1-ac8c71fa4ba9.png">

PR #1548, opened 2022-03-18
Allow sample transfers with ellipsis shown in Samples table

leandroradusky:
Solves #1547. Since we start to show the middle ellipsis in the table of samples:

image

This text was used to fulfill the sample transfers UUIDs.

Now its taking the value attribute of the samples and its working back:

image

**Note:** There was not a failure from the server side, the operation appeared to be successful (not like in #1529 that there is a server error but no message is shown). A check of an existing uuid should be added in the sample transfers? a missing "foreign key"?

### DIFF
--- a/app/views/samples/index.haml
+++ b/app/views/samples/index.haml
@@ -43,7 +43,7 @@
                 %td
                   =check_box_tag 'sample_ids[]', sample.id, false, { id: "sample_ids.#{sample.id}" }
                   %label.row{for: "sample_ids.#{sample.id}"}
-                %td{id: 'sample_uuid', value: sample.uuid}= short_uuid_with_title sample.uuid
+                %td{id: 'sample_uuid', data: {uuid: sample.uuid}}= short_uuid_with_title sample.uuid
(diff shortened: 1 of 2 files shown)


app/views/samples/index.haml
```diff
@@ -43,7 +43,7 @@
                 %td
                   =check_box_tag 'sample_ids[]', sample.id, false, { id: "sample_ids.#{sample.id}" }
                   %label.row{for: "sample_ids.#{sample.id}"}
-                %td{id: 'sample_uuid'}= short_uuid_with_title sample.uuid
+                %td{id: 'sample_uuid', value: sample.uuid}= short_uuid_with_title sample.uuid
```
straight-shoota, line 46: Rather than inventing a custom `value` attribute, this should use a `data-*` attribute. `data-` is essentially a namespace for custom attributes.
```suggestion
                %td{id: 'sample_uuid', data: {uuid: sample.uuid}}= short_uuid_with_title sample.uuid
```

straight-shoota (approved):

Issue #513, opened 2015-11-12
Add a toggle to choose to grant permission to a site and the entire hierarchy or just to that site


nekron:
And implement the underlying functionality in the policies model.

nditada:
@nekron Do we have a design in invision for this? can you link?

nekron:
https://projects.invisionapp.com/d/main#/console/4103589/113464930/preview
and
https://projects.invisionapp.com/d/main#/console/4103589/113464672/preview

mmuller:
Ok in version 0.11-pre7

PR #1554, opened 2022-03-21
Check for institutions when a sample is attempted to be transferred

### LINKED ISSUE #1529: Show error message in sample transfer modal when mandatory fields are missing

leandroradusky:
Fix #1529. Now, when an institution is missing in a sample transfer the following message is shown.

image

The error message shape & style was copied from sample_transfer_confirm when an ID is invalid.

When an institution is selected, the error message is hidden.

### DIFF
--- a/app/assets/javascripts/sample_transfer.js.jsx
+++ b/app/assets/javascripts/sample_transfer.js.jsx
@@ -50,16 +50,20 @@ var SampleTransferModal = React.createClass({
       includes_qc_info: this.state.includeQcInfo,
       samples: this.state.selectedSamples.map((sample) => sample.uuid)
     }
-    $.ajax({
-      url: '/sample_transfers',
-      method: 'POST',
-      data: data,
-      success: function () {
-        this.closeModal();
-        window.location.reload(true); // reload page to update users table
-      }.bind(this)
-    });
-
+    if (data.institution_id == null){
+      $(".error").prop("hidden",false);
+    }
+    else{
+      $.ajax({
+        url: '/sample_transfers',
+        method: 'POST',
+        data: data,
+        success: function () {
+          this.closeModal();
+          window.location.reload(true); // reload page to update users table
+        }.bind(this)
+      });
+    }
 
   },
 
@@ -70,6 +74,9 @@ var SampleTransferModal = React.createClass({
   },
 
   changeInstitution: function(newValue) {
+    if ($(".error").prop("hidden") === false){
+      $(".error").prop("hidden", true)
+    } 
     this.setState({
       institutionId: newValue,
     })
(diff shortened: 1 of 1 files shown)


straight-shoota:
I'd suggest to place the error message next to the form field it relates to. That makes it easier to understand for the user.
It's also helpful to highlight the errorenous field itself (add `input-required` class to the input, like in #1513).

diegoliberman:
Let's change the error message to "Please choose an institution".

straight-shoota:
I'd try to make UI wording consistent with everything else. If it was an `ActiveModel` error message, it would read "Institution can't be blank". That's probably best.

diegoliberman:
Let's go with Johannes message suggestion

app/assets/javascripts/sample_transfer.js.jsx
```diff
@@ -50,16 +50,20 @@ var SampleTransferModal = React.createClass({
...
-      method: 'POST',
-      data: data,
-      success: function () {
-        this.closeModal();
-        window.location.reload(true); // reload page to update users table
-      }.bind(this)
-    });
-
+    if (data.institution_id == null){
+      $(".error").prop("hidden",false);
```
ysbaddaden, line 54: This is a little too generic, maybe use `.institution-error`?

straight-shoota: In #513 I introduced a generic CSS selector based on adjacency to an `input-required` field: https://github.com/instedd/cdx/pull/1513/files#diff-a46c1b13bcada89b4a3a77b540b1ff786ca75ec41c4e190dd1614b14e78dfca2
This probably won't work directly because the CdxSelect is not an `input`, but we can enhance the CSS rule to support that as well.

straight-shoota: Then you don't have to worry about toggling `error` messages, just `input-required` on the field.

leandroradusky: The use of `error` as a class was taking as example what Johannes previously did in sample transfers confirm:

```
<span className="error"><div className="icon-error icon-red" /> Invalid sample ID</span>
```

but I wasn't applying his logic. Now I've modified the code to make it work with this logic.

**Note:** I have to assign the "institution-select" _class_ and I couldn't do it with element's _id_ as johannes did for sample transfers confirm because the react component is not keeping the id. Hope it's ok like this.

straight-shoota (approved):

ysbaddaden:
@leandroradusky I can't squash & merge because there is a conflict with the `main` branch!

PR #1557, opened 2022-03-21
Fix filter sample transfers by `old_batch_number`

### LINKED ISSUE #1538: The search by batch ID in the log of transfer doesn't work
When searching by Batch ID, it's not finding any matches even when I did transferred samples that used to belong to that batch id.

1. Create a batch with N samples
2. Transfer 1 of the samples to institution B
3. Go to the log of transfers and filter by batch ID
4. You'll get no results.
5. Login to institution B.  You'll get to the same bug.

image
image

image

straight-shoota:
Fixup for  #1544
I missed to add a left join on `batches` because the relation doesn't exist when `old_batch_number` is set.

The spec was also broken because the default sample has a batch associated and that made the spec pass.

Resolves #1538

### DIFF
--- a/app/controllers/sample_transfers_controller.rb
+++ b/app/controllers/sample_transfers_controller.rb
@@ -8,7 +8,7 @@ class SampleTransfersController < ApplicationController
       .ordered_by_creation
 
     @sample_transfers = @sample_transfers.joins(sample: :sample_identifiers).where("sample_identifiers.uuid LIKE concat('%', ?, '%')", params[:sample_id]) unless params[:sample_id].blank?
-    @sample_transfers = @sample_transfers.joins(sample: :batch).where("batches.batch_number LIKE concat('%', ?, '%') OR samples.old_batch_number LIKE concat('%', ?, '%')", params[:batch_number], params[:batch_number]) unless params[:batch_number].blank?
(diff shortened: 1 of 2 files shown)


app/controllers/sample_transfers_controller.rb
```diff
@@ -8,7 +8,7 @@ def index
       .ordered_by_creation
 
     @sample_transfers = @sample_transfers.joins(sample: :sample_identifiers).where("sample_identifiers.uuid LIKE concat('%', ?, '%')", params[:sample_id]) unless params[:sample_id].blank?
-    @sample_transfers = @sample_transfers.joins(sample: :batch).where("batches.batch_number LIKE concat('%', ?, '%') OR samples.old_batch_number LIKE concat('%', ?, '%')", params[:batch_number], params[:batch_number]) unless params[:batch_number].blank?
+    @sample_transfers = @sample_transfers.joins(:sample).joins("LEFT JOIN batches ON batches.id = samples.batch_id").where("batches.batch_number LIKE concat('%', ?, '%') OR samples.old_batch_number LIKE concat('%', ?, '%')", params[:batch_number], params[:batch_number]) unless params[:batch_number].blank?
```
ysbaddaden, line 11: Damn, we need ActiveRecord 5 to leverage `#left_joins`!

ysbaddaden (approved):

PR #1559, opened 2022-03-24
Success message on transfer samples

### LINKED ISSUE #1528: The transfer action should be showing a success message
As described in the [design](https://projects.invisionapp.com/d/main#/console/4103589/461729949/preview), after transferring samples, we should be showing a success toast message.

leandroradusky:
Close #1528.

**For this issue I took advantage to learn more about rails and react then:** 
- @diegoliberman maybe I should change some of the hours to the learning keyword.
- I'm suggesting @ysbaddaden and @straight-shoota both as reviewers since I'm not confident this is a good solution and all your input will be welcome. 

On successful sample transfers now a success message is displayed

image

Based on the [mockup](https://projects.invisionapp.com/d/main#/console/4103589/461729949/preview) of successful transfer samples, but also taking into account that there is a whole set of [snackbars](https://projects.invisionapp.com/d/main#/console/4103589/123213438/preview) that will be displayed I've created the _FlashSnackbar_ component, and its "child" component _SampleTransferSuccessSnackbar_. 

The snackbar name was taken from the Jony's mockups. It works like the flash notice but has a different design, then I've duplicated what is for flash notice for the snackbar (this wont work on redirect_to+:notice actions, all the messages now will look like these snackbars in the future and then it will replace them?)

_FlashSnackbar_ 

It takes the icon and the message's lines as parameters and display them with the proper style.

_SampleTransferSuccessSnackbar_

Creates a FlashSnackbar displaying the proper message with the number of transferred samples and the institution as parameters. 

@straight-shoota I didn't replaced the ajax call, I think I will need some pairing on this if you think what is in sample_transfer.js.jsx is a deficient solution.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -70,6 +70,13 @@ module ApplicationHelper
     I18n.localize(value, locale: current_user.locale, format: I18n.t('date.input_format.pattern'))
   end
 
+  def snackbar
+    res = content_tag :div, id: "snackbar" do 
+      '' 
+    end
+    res
+  end
+
   def flash_message
     res = nil
 
--- a/app/assets/javascripts/components/flash_snackbar.js.jsx
+++ b/app/assets/javascripts/components/flash_snackbar.js.jsx
@@ -1,0 +1,33 @@
+var FlashSnackbar = React.createClass({
+  componentDidMount: function() {
+    var flashn = $(".flash-snackbar");
+    flashn.delay(3000).fadeOut(300);
+      $(".close-notification").click(function(){
+      flashn.hide();
+    });
+  },
+  render: function() {
+    return (
+        <div className="flash flash-snackbar">
+            <div className={"icon "+this.props.icon} />
+            <div className="flash-snackbar-message" > 
+                <span className="flash-snackbar-line1">{this.props.line1}</span> <br />
+                <span className="flash-snackbar-line2">{this.props.line2}</span>
+            </div>
+            <a className='close-notification'>x</a>
+        </div>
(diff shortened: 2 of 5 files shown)


straight-shoota:
Oh yeah. The snackbar looks beefy =)

This is great. But I think it is actually a second step. Visuals of notifications are a general improvement, not specific to the sample transfer action.

The first step to solve https://github.com/instedd/cdx/issues/1528 should be to transform the ajax request + reload into a direct  request by the browser, using just standard rails behaviour. This doesn't need any UI changes at all for fixing the bug.

app/helpers/application_helper.rb
```diff
@@ -70,6 +70,13 @@ def format_date(value, time_zone = nil)
     I18n.localize(value, locale: current_user.locale, format: I18n.t('date.input_format.pattern'))
   end
 
+  def snackbar
+    res = content_tag :div, id: "snackbar" do 
+      '' 
+    end
+    res
+  end
```
ysbaddaden, line 78: This can be simplified:
```ruby
content_tag :div, "", id: "snackbar"
```

Or just use a plain string since we don't interpolate anything —we declare it HTML safe, so the renderer won't escape entities:

```ruby
%(<div id="snackbar"></div>).html_safe
```

But since there isn't anything special, you can just do it right in HAML:

```haml
  #snackbar
```

ysbaddaden:
@leandroradusky I agree with @straight-shoota : let's avoid _adding more JS_ when we can _remove some JS_ and let the natural form submission -> redirection -> page reload flow happen.

PR #1560, opened 2022-03-24
Merge institutions#new_from_invite_data into institutions#new

### LINKED ISSUE #1467: Alert that invitation has already been accepted instead of crashing
We don't properly handle the case where an invitation has already been accepted, which results in a MissingTemplate exception, reported to the user as a 500 Internal Server Error. We should instead report the actual error with an alert message such as "Error: this invitation has already been accepted. You can login to access your institution".

Follow up on these comments:
- https://github.com/instedd/cdx/pull/1458#pullrequestreview-880427158
- https://github.com/instedd/cdx/pull/1458#discussion_r805736422

ysbaddaden:
Refactor of pending institution invite in institutions controller. Properly takes care of pending and accepted invites (or no invite), with an error flash message when the invitate has already been accepted, and a redirection when the invite can't be found.

closes #1467

### DIFF
--- a/app/controllers/institutions_controller.rb
+++ b/app/controllers/institutions_controller.rb
@@ -2,6 +2,7 @@ class InstitutionsController < ApplicationController
   before_filter :load_institutions
   skip_before_filter :check_no_institution!, only: [:new, :create, :pending_approval, :new_from_invite_data]
   skip_before_action :ensure_context, except: [:no_data_allowed]
+  before_action :load_pending_invite!, only: [:new, :new_from_invite_data, :create]
 
   def index
     @can_create = has_access?(Institution, CREATE_INSTITUTION)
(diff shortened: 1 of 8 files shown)


app/controllers/institutions_controller.rb
```diff
@@ -112,30 +100,21 @@ def no_data_allowed
...
   end
 
   private
 
-  def accept_pending_invite
-    if @pending_invite && current_user.email == @pending_invite.invited_user_email
-      @pending_invite.status = 'accepted'
-      @pending_invite.save!
+  def load_pending_invite!
+    return unless id = params[:institution] ? params[:institution][:pending_institution_invite_id] : params[:pending_institution_invite_id]
```
straight-shoota, line 112: There's so much in this line... what about this instead:
```suggestion
    id = params.dig(:institution, :pending_institution_invite_id) || params[:pending_institution_invite_id] || return
```

straight-shoota (approved):
❤️

app/controllers/institutions_controller.rb
ysbaddaden: I thought of `#dig` but we need Ruby 2.3, or the patch from the rails5 upgrade, or I can just use the custom `#get_in` method we currently have (then replace for `#dig` when rebasing the rails5 branch).

PR #1561, opened 2022-03-25
Sample transfer modal converted into form and success message shown

### LINKED ISSUE #1528: The transfer action should be showing a success message

leandroradusky:
Closes #1528.

Samples transfer modal is now an html form and is not making ajax call to create sample transfers.

On success it shows the following notice

image

(same on errors, keeping the logic and the messages that were implemented in sample transfers controller create method)

**Note:** I had to add the following in the sample transfers controller (I saw that there is used in other parts of the code but wanted to be sure this is noted since I'm not sure of the consequences): 
```
skip_before_filter :verify_authenticity_token, :only => :create
skip_before_filter :ensure_context, :only => :create
```

### DIFF
--- a/app/assets/javascripts/components/sample_transfer.js.jsx
+++ b/app/assets/javascripts/components/sample_transfer.js.jsx
@@ -133,7 +133,7 @@ var SampleTransferModal = React.createClass({
         <div className="row">
           <div className="col pe-3"><label>Institution</label></div>
           <div className="col">
-            <CdxSelect className="institution-select" name="institution" items={this.props.institutions} value={this.state.institutionId} onChange={this.changeInstitution} />
+            <CdxSelect className="institution-select" name="institution_id" items={this.props.institutions} value={this.state.institutionId} onChange={this.changeInstitution} />
             <span className="error"><div className="icon-error icon-red" /> Institution can't be blank</span>
           </div>
           
--- a/app/assets/javascripts/components/sample_transfer.js.jsx
+++ b/app/assets/javascripts/components/sample_transfer.js.jsx
@@ -118,8 +118,9 @@ var SampleTransferModal = React.createClass({
         <label htmlFor="include-qc-check">Include a copy of the QC data</label>
       </div>
     </div>)
-  }
-  , render: function() {
+  }, 
+  
+  render: function() {
     return(
       <div className="samples-transfer-modal">
         <form action="/sample_transfers" method="post">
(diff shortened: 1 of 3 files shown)


straight-shoota:
Yeah, disabling those security measures raises some flags. We should try to avoid that.

For `ensure_context` I suppose that's simple: Just forward the context property to the form.
For `authenticity_token`, we could just generate a token in the index controller action and send it out so the form can use it.

leandroradusky:
The tests are failing since on the sample transfer creation tests there are

```expect(response).to be_success```

But now there are redirections with notice, which are distinct http codes... Should I change the spec to expect a redirection?

File: app/assets/javascripts/components/sample_transfer.js.jsx [hunk 1]
```diff
@@ -132,46 +103,51 @@ var SampleTransferModal = React.createClass({
     this.setState({
       includeQcInfo: !oldValue
     });
+
+    if ($("#include-qc-check").is(":checked")) {
+      $("#include-qc-check").attr("value", "true");
+    } else {
+      $("#include-qc-check").attr("value", "false");
+    }
```
ysbaddaden, line 111: This could optimized & simplified: 
```js
var qcChecked = $("#include-qc-check");
qcChecked.attr("value", !qcChecked.is(":checked");
```

ysbaddaden (approved):
Let's merge as-is.

I didn't notice that the form wasn't rendered by Rails, but by React. It seems this is only because of the <Select> component and maybe the close modal button. This still isn't the Rails' workflow (render forms in Rails then inject the HTML into the DOM).

I'm... forecasting forms breaking and a lot of developer headaches when we'll try to enable the per-form CSRF tokens of Rails 5 :sob:

app/assets/javascripts/components/sample_transfer.js.jsx
straight-shoota: Why is this even necessary? Shouldn't the value already be set accordingly?

app/controllers/sample_transfers_controller.rb
```diff
@@ -19,15 +21,15 @@ def index
...
     else
       if create_transfer(new_owner, params["samples"])
-        flash[:success] = "All samples have been transferred successfully."
-        render json: { status: :ok }
+        flash[:notice] =  "All samples have been transferred successfully. "
+        redirect_to samples_path
       else
-        flash[:error] = "Samples transfer failed."
-        render json: { status: :error }, status: 500
+        flash[:error] = "Samples transfer failed. "
```
straight-shoota, line 31: There are some weird whitespace changes here.

app/assets/javascripts/components/sample_transfer.js.jsx
leandroradusky: If the checkbox is checked the true value is sent by the form, but if it is not checked, no information is sent for this input by the form, that's why I added it.

PR #1562, opened 2022-03-25
Fix: wrong icon for batches in tab selector

### LINKED ISSUE #1425: Icons for Samples & Batches don't match design
Johny spotted that these icons are not exactly the same than provided in the designs.

Implemented
image
image

Designed
image
image

ysbaddaden:
Capture d'écran du 2022-03-25 19 34 17@2x

I had to replace the icon in the icomoon font, because it didn't render properly, and I added some notes about how to regenerate a font with added (or removed) icons.

@diegoliberman @jkicillof I didn't reorder the tabs to match the design (batches, samples, transfers), since we already deployed the app a while ago with the wrong order (samples, batches, transfers). I'm not sure if we want to change it _now_?

fixes #1425


diegoliberman:
Let's not worry about this.  Anyhow, there are plans for doing greater changes in the navigation, we'll discuss it then.

app/assets/stylesheets/_sprites.scss
```diff
@@ -5,21 +5,21 @@
...
-        font-url('icomoon.ttf') format('truetype'),
-        font-url('icomoon.woff') format('woff'),
-        font-url('icomoon.svg#icomoon') format('svg');
-    font-weight: normal;
-    font-style: normal;
+  font-family: 'icomoon';
+  src:
+    font-url('icomoon.ttf') format('truetype'),
+    font-url('icomoon.woff') format('woff'),
+    font-url('icomoon.svg#icomoon') format('svg');
```
ysbaddaden, line 12: I simplified the font-face declaration and removed support for IE8. We could probably remove the TTF and SVG formats, too from the font-face declaration, too. All browsers (including IE9) support the WOFF format.

We could create a WOFF2 version, which would be half the size of the WOFF version (7KB instead of 16KB).

devduarte (approved):
looks good!

PR #1569, opened 2022-03-29
Fix: tiny optimizations quick wins

ysbaddaden:
While trying out a pull request, I noticed lots of repeated SQL requests in the Rails' log, so I went to add a few simple optimizations to heavily reduce the number of SQL requests in `samples#index` and `batches#show`.

With a list of 10 samples, this replaces 20 SQL requests for 2 requests just by preloading the `:batch` and `:sample_identifiers` associations (a one line optimization), and removes 20 other SQL requests by delaying the access check into an intermediary action (`edit_or_show`) that will only check the permission for one sample on demand.

This is a total of 38 less SQL requests each time we render a list of samples :tada: 

The same `edit_or_show` action was added to batches, with the same result. There was no association to preload so we only removed 20 requests in total :shrug: 

**We can certainly do the same in other controllers!**

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -25,6 +25,16 @@ class SamplesController < ApplicationController
       .preload(:batch, :sample_identifiers)
   end
 
+  def edit_or_show
+    sample = Sample.find(params[:id])
+
+    if has_access?(sample, UPDATE_SAMPLE)
+      redirect_to edit_sample_path(sample)
+    else
+      redirect_to sample_path(sample)
+    end
+  end
+
   def show
     sample = Sample.find(params[:id])
     @sample_form = SampleForm.for(sample)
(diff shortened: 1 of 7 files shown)


app/controllers/samples_controller.rb
```diff
@@ -22,6 +22,17 @@ def index
       .map { |uuid, name| { value: uuid, label: name } }
 
     @samples = perform_pagination(@samples)
+      .preload(:batch, :sample_identifiers)
```
ysbaddaden, line 25: Note that I use `#preload` that generates an explicit SQL request for each association to preload, instead of `#includes` that would often SELECT and JOIN right into the initial SQL request. I always prefer explicit requests, which are much simpler (and saner).

devduarte (approved):
looks good!

straight-shoota:
Just a quick thought that came into mind about this: Do `edit` and `show` actually need to be different actions? They're pretty much the same in the way that they present the data. The only difference is that authorized users can modify the data and persist the changes.
So I'm wondering whether we could get rid of the two separate actions and just merge them into `edit_or_show` directly, without a redirect.

PR #1574, opened 2022-03-31
Introduce TransferPackage + Responsible Person

### LINKED ISSUE #1572: Add responsible recipient when transferring samples
When transferring samples, a very important piece of information is the name of the person that is legally responsible for receiving the samples.  For traceability, they need to keep track of who was this person for each transfer.

- [x] Let's add a free text field, below the institution field.  [Invision](https://projects.invisionapp.com/d/main#/console/4103589/465275290/preview).

The field should be optional, as for use cases beyond UCSD, it might not be relevant data.

- [x] Let's add a column in the log of transfers.  [Invision](https://projects.invisionapp.com/d/main#/console/4103589/465275288/preview).

### Note
We already have 3 fields for the transfer action: Institution, Responsible, and Include QC Info.   It's becoming logic to think that the transfer should be an entity of its own in the system's backend and frontend.  For now, let's not do nothing non-trivial to go in this direction.  Just bear in mind that at one point, we will have to consider it so that we better represent reality.

ysbaddaden:
This is an internal refactor and thus totally transparent. It somewhat helps to clean up the `Sample` and `SampleTransfersController` objects by moving the data manipulation into the `TransferPackage` object.

~~All that's left is to add the recipient field and the table column.~~

closes #1572

### DIFF
--- a/app/models/sample_transfer.rb
+++ b/app/models/sample_transfer.rb
@@ -2,7 +2,9 @@ class SampleTransfer < ActiveRecord::Base
   belongs_to :sample
   belongs_to :sender_institution, class_name: "Institution"
   belongs_to :receiver_institution, class_name: "Institution"
+  belongs_to :transfer_package, required: false
 
+  # TODO: remove these after upgrading to Rails 5.0 (belongs_to associations are required by default):
   validates_presence_of :sample
   validates_presence_of :sender_institution
   validates_presence_of :receiver_institution
--- a/app/models/transfer_package.rb
+++ b/app/models/transfer_package.rb
@@ -1,0 +1,45 @@
+class TransferPackage < ActiveRecord::Base
+  belongs_to :receiver_institution, class_name: "Institution"
+  has_many :sample_transfers
+
+  after_initialize do
+    self.uuid ||= SecureRandom.uuid
+  end
+
+  def self.sending_to(institution, attributes = nil)
+    create!(attributes) do |package|
+      package.receiver_institution = institution
+    end
+  end
+
+  def add!(sample)
+    transfer = sample_transfers.create!(
+      sample: sample,
+      receiver_institution: receiver_institution,
+    )
+
+    if sample.batch
+      if includes_qc_info && (qc_info = create_qc_info(sample))
+        sample.qc_info = qc_info
+      end
+
+      sample.old_batch_number = sample.batch.batch_number
+    end
+
+    sample.update!(batch: nil, site: nil, institution: nil)
+
+    transfer
+  end
+
+  private
+
+  def create_qc_info(sample)
+    sample_qc = sample.batch.qc_sample
+    return unless sample_qc
+
+    qc_info = QcInfo.find_or_duplicate_from(sample_qc)
+    qc_info.samples << sample
+    qc_info.save!
+    qc_info
+  end
+end
--- a/db/migrate/20220331164557_create_transfer_packages.rb
+++ b/db/migrate/20220331164557_create_transfer_packages.rb
@@ -1,0 +1,19 @@
+class CreateTransferPackages < ActiveRecord::Migration
+  def up
+    create_table :transfer_packages do |t|
+      t.belongs_to :receiver_institution,            null: false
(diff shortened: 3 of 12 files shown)


app/models/sample_transfer.rb
```diff
@@ -2,7 +2,9 @@ class SampleTransfer < ActiveRecord::Base
   belongs_to :sample
   belongs_to :sender_institution, class_name: "Institution"
   belongs_to :receiver_institution, class_name: "Institution"
+  belongs_to :transfer_package, required: false
```
ysbaddaden, line 5: This is in preparation for ActiveRecord 5!

app/models/transfer_package.rb
```diff
@@ -0,0 +1,50 @@
...
+  self.primary_key = :id
+
+  belongs_to :receiver_institution,
+    class_name: "Institution",
+    primary_key: "uuid"
+
+  has_many :sample_transfers
+
+  after_initialize do
+    self.id ||= SecureRandom.uuid
```
ysbaddaden, line 11: CDx uses `Guid.new` everywhere but that gem is no longer maintained, and it looks like it generates invalid UUIDs (e.g. all 16 bytes are fully random). Let's start moving to the proper UUIDs.

db/migrate/20220331164557_create_transfer_packages.rb
```diff
@@ -0,0 +1,20 @@
+class CreateTransferPackages < ActiveRecord::Migration
+  def up
+    create_table :transfer_packages, id: false do |t|
+      t.string :id,                      limit: 36, null: false
+      t.string :receiver_institution_id, limit: 36, null: false
+      t.string :recipient
+
+      t.index :id, unique: true
```
ysbaddaden, line 8: We can't seem to generate a `:string` primary key with ActiveRecord 4. It seems we be pass `:id => :string` with AR 5.0, but for now, I couldn't find a proper solution with AR4 :cry: 

Maybe I should follow the rest of the models: using integer ids for internal relationships, and have a distinct UUID column.

ysbaddaden: I did just that: we now have an `id` (int) and an `uuid` (string) columns.

db/migrate/20220331164557_create_transfer_packages.rb
```diff
@@ -0,0 +1,19 @@
+class CreateTransferPackages < ActiveRecord::Migration
+  def up
+    create_table :transfer_packages do |t|
+      t.belongs_to :receiver_institution,            null: false
+      t.string     :uuid,                 limit: 36, null: false
+      t.string     :recipient
+      t.boolean    :includes_qc_info, default: false
```
ysbaddaden, line 7: We wouldn't need to persist the value with AR5, where we could just have a virtual boolean attribute on the model:

```ruby
class TransferPackage
  attribute :includes_qc_info, :boolean, default: false
end
```

db/schema.rb
```diff
@@ -662,6 +663,13 @@
   add_index "test_results", ["site_id"], name: "index_test_results_on_site_id", using: :btree
   add_index "test_results", ["uuid"], name: "index_test_results_on_uuid", using: :btree
 
+  create_table "transfer_packages", force: :cascade do |t|
+    t.integer "receiver_institution_id", limit: 4,                   null: false
+    t.string  "uuid",                    limit: 36,                  null: false
+    t.string  "recipient",               limit: 255
+    t.boolean "includes_qc_info",                    default: false
+  end
```
straight-shoota, line 671: Is there a particular reason for omitting timestamps on this model? We have them pretty much everywhere else and it's good practice for some limited keeping track of changes.

ysbaddaden: Yes: I forgot :face_exhaling: 

I used `rails g migration` instead of `rails g model` so I ended with an empty migration file, instead of a basic one with `create_table`.

ysbaddaden: And would be super useful. Let's add a migration to add them and push a patch release quickly?

PR #1580, opened 2022-04-01
Add parallel_tests to run specs in parallel

ysbaddaden:
Installs [parallel_tests](https://github.com/grosser/parallel_tests/tree/v2.32.0) to allow running specs in parallel. You may run the following commands (where `4` can be replaced with how many processes you really want):

```console
$ rake parallel:setup[4]
$ parallel_rspec -n 4 spec --exclude-pattern spec/features
$ parallel_rspec -n 4 spec/features
$ parallel_cucumber features/ -n 4
```

NOTE: test runs seems to go slower and slower as we repeat them. Re-creating the databases with `rake parallel:drop[4]` followed by `rake parallel:setup[4]` fixes the issue. I guess this is MySQL struggling?

NOTE: you'll have to rebuild the `selenium` image because we now start 8 geckodriver to be able to drive many firefox instances in parallel :grin: 

NOTE: parallel_tests will save logs of each run to optimize the distribution of specs in the different runners!

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,4 @@
---color --require spec_helper
+--color
+--require spec_helper
+--format progress
+--format ParallelTests::RSpec::RuntimeLogger --out tmp/parallel_runtime_rspec.log
--- a/Dockerfile.selenium
+++ b/Dockerfile.selenium
@@ -12,4 +12,15 @@ RUN apk del curl
 EXPOSE 4444
 EXPOSE 5900
 
-ENTRYPOINT ["geckodriver", "--host", "0.0.0.0"]
+RUN echo "#! /usr/bin/env sh" > /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4444 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4445 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4446 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4447 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4448 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4449 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4450 &" >> /entrypoint.sh && \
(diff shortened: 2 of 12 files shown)


ysbaddaden:
ping @straight-shoota

Dockerfile.selenium
```diff
@@ -12,4 +12,15 @@ RUN apk del curl
...
+RUN echo "#! /usr/bin/env sh" > /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4444 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4445 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4446 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4447 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4448 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4449 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4450 &" >> /entrypoint.sh && \
+    echo "geckodriver --host 0.0.0.0 -p 4451" >> /entrypoint.sh && \
+    chmod +x /entrypoint.sh
```
straight-shoota, line 24: I'm wondering if instead of running multiple drivers in the same container, we could make use of docker-compose's scaling feature to automatically spawn multiple containers from the same image.

.rspec
```diff
@@ -1 +1,4 @@
---color --require spec_helper
+--color
+--require spec_helper
+--format progress
+--format ParallelTests::RSpec::RuntimeLogger --out tmp/parallel_runtime_rspec.log
```
straight-shoota, line 4: These flags apply to non-parallel test runs as well, right? Would there be any effect in that case?

Dockerfile.selenium
ysbaddaden: Oooh, can you please try that?

.rspec
ysbaddaden: Yeah, it will calculate the run times and update the log. Sounds harmless?

straight-shoota: I suppose it's fine.

ysbaddaden:
@straight-shoota shall I merge? It greatly improves the testsuite speedup locally —not on CI where we probably only have 1 shared CPU.

straight-shoota:
Sure go ahead!

straight-shoota (approved):

Issue #1339, opened 2021-10-06
Sample saved even when entering invalid specimen role


diegoliberman:
### Steps to reproduce the error
1. Create a new sample
2. Enter a production date
3. Select a specimen role, for instance "F"
![image](https://user-images.githubusercontent.com/25181052/136192286-b4c5fe57-30e6-42f5-83e8-48af2f6abd66.png)
4. Place the focus in the specimen role and type an invalid option
![image](https://user-images.githubusercontent.com/25181052/136192323-25070238-fb17-44eb-9860-7f5db99aac19.png)
5. Save the sample

The form will be saved without showing any errors.  The specimen role will be set to the one selected in step 3. 

### Expected behaviour
When entering an invalid option in the specimen role field, and attempting to save the form, it should show the error message that shows today if you skip step 3.
![image](https://user-images.githubusercontent.com/25181052/136192610-625a875e-7b41-4354-81c5-df447fa14753.png)

### Notes
This should be considered for all autocompletes.  For instance, for the LOINC field.

mmuller:
@diegoliberman I couldn't reproduce it by following these steps. 

Please see the screen recording:
https://recordit.co/11FLzQQiT8

diegoliberman:
@mmuller perhaps it's a missed frame in the screen recording, but as far as I can see, the steps weren't reproduced exactly as I tried to explain.  In step 3, pay attention to the screenshot, where I show that the selected option is the one that starts with F, not the letter "F".

straight-shoota:
The same issue applies to the LOINC field.  And probably other autocomplete form fields as well.

mmuller:
Working properly in version fd72912

PR #1584, opened 2022-04-06
Invalid options in cdx_select component

devduarte:
Fixes: #1339

-If the cdx_component rendered is searchable, then we set typed options to be the value of the select and they get validated in the backend. For now, only batches and samples forms are searchable.

Make the select searchable (thus, allow the user to type in the input) is a functional requirement since they may not know all the available options and they can be searched by their description. The fix feels a little hacky to me, since we are forcing the input value but the component rejects invalid and uses the last one valid, which is the behaviour that is causing the issue. 
Another possibility was clear the select whenever a valid option is not selected, but then the user would save an empty value  without even noticing it.


app/assets/javascripts/components/cdx_select.js.jsx
```diff
@@ -16,21 +16,27 @@ var CdxSelect = React.createClass({
     }
   },
 
+  onInputChange: function (newValue) {
+    this.selectRef.setValue(newValue)
```
ysbaddaden, line 20: I'd suggest to check for `this.props.searchable` in this callback, and to add a comment explaining why we're updating the selectRef value or not.

straight-shoota:
I think I don't quite understand the purpose of [app/views/samples/_form_specimen_role_js.haml](https://github.com/instedd/cdx/pull/1584/files#diff-01706b92862baf74eaa742770b719cf14bb36409e5ee977521160e658445edc9) and why it's removed (or rather why removing it is fine).

devduarte:
> I think I don't quite understand the purpose of [app/views/samples/_form_specimen_role_js.haml](https://github.com/instedd/cdx/pull/1584/files#diff-01706b92862baf74eaa742770b719cf14bb36409e5ee977521160e658445edc9) and why it's removed (or rather why removing it is fine).

That file had an old autocomplete component we were using for that select and since we're not longer using it, that file is not being used anymore.

app/assets/javascripts/components/cdx_select.js.jsx
devduarte: do you mean something like this? 

```  
 if(this.props.searchable){
   this.selectRef.setValue(newValue) 
 }
```

devduarte: we already check for that property here: `onInputChange={this.props.searchable && this.onInputChange}>`, adding that check would redundant, wouldn't it? Or there is something that I'm missing?

ysbaddaden (approved):
I reproduced the bug, and added a note about the behavior to explain why we use this hook.

app/views/samples/_form.haml
```diff
@@ -42,7 +42,7 @@
           = f.label :specimen_role
         .col
           - if @can_update
-            = cdx_select form: f, name: :specimen_role, searchable: true, :class => 'input-x-large' do |select|
+            = cdx_select form: f, name: :specimen_role, searchable: true,  :class => 'input-x-large' do |select|
```
straight-shoota, line 45:
```suggestion
            = cdx_select form: f, name: :specimen_role, searchable: true, :class => 'input-x-large' do |select|
```

straight-shoota (approved):

PR #1591, opened 2022-04-11
Added validation for institution name

### LINKED ISSUE #1585: Validate uniqueness of institution name
Now that we can transfer samples, having duplicated names of institutions, causes some problems: you can't differentiate which institutions is the one that you actually want.
image

- [ ] When creating an institution, its name must be unique across CDx.
- [ ] When editing the institution, the same validation applies.
- [ ] When inviting institution, the same validation applies.
- [ ] Manually edit institutions that remained with duplicate names in Prod or Staging.

leandroradusky:
Close #1585. 

When an institution is added or edited and the name already exists (case-insensitive) now validation fails and this error message is shown:

On create: 

image

On edit:

image

### DIFF
--- a/app/assets/stylesheets/_errors.scss
+++ b/app/assets/stylesheets/_errors.scss
@@ -9,5 +9,22 @@ form {
     font-style: italic;
     color: $red;
     margin-top: 5px;
+    visibility: hidden;
+    display: flex;
+    flex-direction: column;
+
+    span {
+      display: inline-flex;
+      align-items: center;
+    }
+  }
+
+  .field_with_errors {
+    color: #EB2122;
+    border-bottom-color: #EB2122 !important;
+  }
+    
+  div.field_with_errors + .error {
+    visibility: visible;
   }
 }
--- a/app/assets/stylesheets/_errors.scss
+++ b/app/assets/stylesheets/_errors.scss
@@ -4,6 +4,7 @@
 }
 
 form {
+  // Deprecated: to be generalized
   .error {
     font-size: 12px;
     font-style: italic;
@@ -11,7 +12,6 @@ form {
     margin-top: 5px;
     visibility: hidden;
     display: flex;
-    flex-direction: column;
     
     li {
       display: inline-flex;
--- a/app/assets/stylesheets/_errors.scss
+++ b/app/assets/stylesheets/_errors.scss
@@ -4,7 +4,7 @@
 }
 
 form {
-  .error-field {
+  .error {
     font-size: 12px;
     font-style: italic;
     color: $red;
@@ -12,8 +12,8 @@ form {
     visibility: hidden;
     display: flex;
     flex-direction: column;
-
-    span {
+    
+    li {
       display: inline-flex;
       align-items: center;
     }
@@ -24,7 +24,7 @@ form {
     border-bottom-color: #EB2122 !important;
   }
     
-  div.field_with_errors + .error-field {
+  .field_with_errors + .error {
     visibility: visible;
   }
 }
--- a/app/assets/stylesheets/_errors.scss
+++ b/app/assets/stylesheets/_errors.scss
@@ -4,7 +4,7 @@
 }
 
 form {
-  .error {
+  .error-field {
     font-size: 12px;
     font-style: italic;
     color: $red;
@@ -24,7 +24,7 @@ form {
     border-bottom-color: #EB2122 !important;
   }
     
-  div.field_with_errors + .error {
+  div.field_with_errors + .error-field {
     visibility: visible;
   }
 }
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -211,10 +211,21 @@ module ApplicationHelper
     "#{uuid[0,4]}…#{uuid[-4,4]}"
   end
 
-  def field_errors(obj, field)
-    content_tag :div, :class => "error-field" do
-      obj.errors.full_messages_for(field).each do |msg|
-        concat( content_tag(:span, content_tag(:div,"",class:["icon-error","icon-red"]) + msg) )
+  def field_errors(record, attr_name)
+    #content_tag :div, :class => "error-field" do
+    #  obj.errors.full_messages_for(field).each do |msg|
+    #    concat( content_tag(:span, content_tag(:div,"",class:["icon-error","icon-red"]) + msg) )
+    #  end
+    #end
+    messages = record.errors.full_messages_for(attr_name)
+    return if messages.empty?
+  
+    content_tag :ul, class: "error" do
+      messages.each do |message|
+        concat (content_tag :li do
+          concat content_tag :i, "", class: "icon-error icon-red"
+          concat message
+        end)
       end
     end
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -210,4 +210,13 @@ module ApplicationHelper
   def format_uuid(uuid)
     "#{uuid[0,4]}…#{uuid[-4,4]}"
   end
+
+  def field_errors(msgs)
+    content_tag :div, :class => "error" do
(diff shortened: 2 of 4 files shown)


straight-shoota (approved):

leandroradusky:
Improved way to show the errors, now alongside the proper field. 

Now it looks like this:

image

If several errors are present for the field, its shown like this:

image

An application helper was added in order to display the field related messages. While I think this is going to be useful for future implementation in other forms, what I’m not sure is if I’m calling it in its proper way within the form (cc @straight-shoota)

```
= f.text_field :name, readonly: @readonly, :class => ['input-block', @institution.errors.has_key?(:name) ? 'input-required':'']
= field_errors (@institution.errors.full_messages_for(:name))
```

The styles modification in `_errors.scss` does not affect the display of multiple messages in other forms:

image

app/assets/stylesheets/_errors.scss
```diff
@@ -9,5 +9,22 @@ form {
     font-style: italic;
     color: $red;
     margin-top: 5px;
+    visibility: hidden;
```
straight-shoota, line 12: I'm not sure it's a good idea to add `visibility: hidden` by default to any `.error`. This might unintentionally hide errors in some places. We would need to make sure `.error` isn't used somewhere without `.field_with_errors` when it's supposed to show.
In any case, this doesn't seem to be related to this change, right? So I'd rather move it to a separate PR.

app/views/institutions/_form.haml
```diff
@@ -37,7 +36,8 @@
       .col.px-1
         = f.label :name, :class => 'block'
       .col
-        = f.text_field :name, readonly: @readonly, :class => 'input-block'
+        = f.text_field :name, readonly: @readonly, :class => ['input-block', @institution.errors.has_key?(:name) ? 'input-required':'']
+        = field_errors (@institution.errors.full_messages_for(:name))
```
straight-shoota, line 40: I think retrieving the actual error messages could be deferred to the helper method.
```suggestion
        = field_errors f, :name
```

So `field_errors` receives the form builder and the name of the field, and retrieves the messages on its own.

app/assets/stylesheets/_errors.scss
leandroradusky: To make the name less general i rename it to `error-field`

straight-shoota: But now you're still changing an existing style defintion which will impact existing references to that.
And some of these style changes do not seem to be relevant for fixing the issue for validation of institution name.

So I'd suggest to remove all the unrelated changes, such as the visibility, and introduce them in a separate PR that should deal with standardizing form markup and styles all over the app.

app/views/institutions/_form.haml
straight-shoota: No, I meant `f`, the form context. It contains a reference to the object (`f.object`). This avoids some duplication or far-away contextual matching to the form object, which is just easily accessible through the builder.

straight-shoota:
> I've "copied" `input-required` and `error` into _errors.scss form handling, since appeared as the proper place to handle this. Johannes remarked that these two are not well defined, but I don't catch what is needed to do it properly.

It seems these classes had been introduced and adapted ad hoc and are not used consistently throughout the code base. Their function overlaps for the most part, so we could use some consolidation here. For that we need to look at the different (current) use cases and establish a standardized error style system.

leandroradusky:
> It seems these classes had been introduced and adapted ad hoc and are not used consistently throughout the code base. Their function overlaps for the most part, so we could use some consolidation here. For that we need to look at the different (current) use cases and establish a standardized error style system.

Do you think is a good idea to meet next week to define this? (monday I'm out, except for it whenever you want :) )

app/helpers/application_helper.rb
```diff
@@ -210,4 +210,13 @@ def short_uuid_with_title(uuid)
   def format_uuid(uuid)
     "#{uuid[0,4]}…#{uuid[-4,4]}"
   end
+
+  def field_errors(obj, field)
+    content_tag :div, :class => "error-field" do
+      obj.errors.full_messages_for(field).each do |msg|
+        concat( content_tag(:span, content_tag(:div,"",class:["icon-error","icon-red"]) + msg) )
```
ysbaddaden, line 217: - SPAN is an inline element while DIV is a block element; you may nest DIV>SPAN but shouldn't nest SPAN>DIV.
- We're dealing with a _list of errors_, so maybe a UL>LI would be nice to have?
- We can avoid the `visibility:hidden` by only generating the list where there are errors (no need to hide what's not here in the first place).

For example (I'm using `record` and `attr_name` to be closer to Rails' namings):

```ruby
def field_errors(record, attr_name)
  messages = record.errors.full_messages_for(attr_name)
  return if messages.empty?

  content_tag :ul, class: "field-errors" do
    messages.each do |message|
      concat content_tag :li, message, class: "icon-error icon-red"
    end
  end
end
```

app/assets/stylesheets/_errors.scss
ysbaddaden: Yes, please keep `.error` for the time being, with a note that it's deprecated!

ysbaddaden (changes requested):
Hey Lean, great idea to introduce a helper to render individual errors!

Using [FormBuilder](https://api.rubyonrails.org/v4.2/classes/ActionView/Helpers/FormBuilder.html) as suggested by @straight-shoota is a good idea, but it would be better as an actual method to leverage the abstraction (i.e. `f.field_errors :name` for example), otherwise it's not much different (and less generic, since you'd need a FormBuilder)

But that would mean to either monkeypatch the FormBuilder class (nope, bad), or to create a subclass which... is maybe a little overkill?

app/helpers/application_helper.rb
ysbaddaden: @straight-shoota noted in #1597 that we're abusing icomoon's `.icon-` definition, and that maybe it should be:

```ruby
concat content_tag :li do
  concat tag :i, class: "icon-error icon-red"
  concat message
end

Comment (leandroradusky) [reply app/helpers/application_helper.rb:217]: Have to make this to make it work:

```
concat (content_tag :li do
  concat content_tag :i, "", class: "icon-error icon-red"
  concat message
end)
```

Without the parentheses it doesn't work.

File: app/assets/stylesheets/_errors.scss (hunk 1 above)
Comment (leandroradusky) [reply app/assets/stylesheets/_errors.scss:12]: If I rollback to `.error` then I have to keep the visibility rule, since another rule in `_forms.scss`

```
div + .error,
input + .error {
  visibility: hidden;
}

div.input-required + .error,
input.input-required + .error {
  visibility: visible;
}
```
keep it hidden. So I leave it with a different class name and no visibility rule or I keep the `error` class with the visibility rule to make it work.

Comment (straight-shoota) [reply app/assets/stylesheets/_errors.scss:13]: Let's leave it like that and refactor these error styles in a follow-up.

File: app/helpers/application_helper.rb [hunk 4]
```diff
@@ -210,4 +210,24 @@ def short_uuid_with_title(uuid)
   def format_uuid(uuid)
     "#{uuid[0,4]}…#{uuid[-4,4]}"
   end
+
+  def field_errors(record, attr_name)
+    #content_tag :div, :class => "error-field" do
+    #  obj.errors.full_messages_for(field).each do |msg|
+    #    concat( content_tag(:span, content_tag(:div,"",class:["icon-error","icon-red"]) + msg) )
+    #  end
+    #end
```
straight-shoota, line 219:
```suggestion
```

straight-shoota (approved):

app/helpers/application_helper.rb
ysbaddaden: Oh, I think it has to do with the precedence of ruby blocks? It translates `concat content_tag :li do;end` to `concat(content_tag :li) do;end`. Oops.

Issue #1517, opened 2022-03-11
Mixed MySQL string collations in different environments/deployments


ysbaddaden:
In light of #1498 we noticed that there are different string collations in the different MySQL databases of CDx. For example for local development @straight-shoota and I have `utf8_unicode_ci` but in staging we noticed it's `latin1_swedish_ci`? We don't about the production deployments.

We should investigate and have a single collation for all environments that we can rely on. Ideally it should be `utf8mb4_unicode_ci` which is the current recommendation.

For reference, this is what we used to notice the difference:
```sql
SELECT COLUMN_NAME, COLLATION_NAME
FROM information_schema.columns
WHERE TABLE_NAME = 'batches' AND COLUMN_NAME IN ('isolate_name', 'batch_number');
```

ysbaddaden:
@matiasgarciaisaia when you have some minutes :)

matiasgarciaisaia:
Staging:

```
mysql> SELECT COLUMN_NAME, COLLATION_NAME
    -> FROM information_schema.columns
    -> WHERE TABLE_NAME = 'batches' AND COLUMN_NAME IN ('isolate_name', 'batch_number');
+--------------+-------------------+
| COLUMN_NAME  | COLLATION_NAME    |
+--------------+-------------------+
| isolate_name | latin1_swedish_ci |
| batch_number | latin1_swedish_ci |
+--------------+-------------------+
2 rows in set (0.03 sec)
```

Production:

```
mysql> SELECT COLUMN_NAME, COLLATION_NAME
    -> FROM information_schema.columns
    -> WHERE TABLE_NAME = 'batches' AND COLUMN_NAME IN ('isolate_name', 'batch_number');
+--------------+-----------------+
| COLUMN_NAME  | COLLATION_NAME  |
+--------------+-----------------+
| isolate_name | utf8_unicode_ci |
| batch_number | utf8_unicode_ci |
+--------------+-----------------+
2 rows in set (0.00 sec)
```

RadX:

```
mysql> SELECT COLUMN_NAME, COLLATION_NAME
    -> FROM information_schema.columns
    -> WHERE TABLE_NAME = 'batches' AND COLUMN_NAME IN ('isolate_name', 'batch_number');
+--------------+--------------------+
| COLUMN_NAME  | COLLATION_NAME     |
+--------------+--------------------+
| isolate_name | utf8mb4_unicode_ci |
| batch_number | utf8mb4_unicode_ci |
+--------------+--------------------+
[...]

PR #1593, opened 2022-04-12
Showing all messages on sample form validation

### LINKED ISSUE #1476: [Bug] Incomplete sample form validation
I noticed another UX issue related to the samples form. When there's an error in some of the sample data that's validated in `SampleForm`, other errors won't be shown because validation stops before validating the actual mode.

For example, when I create a new sample but fill no input field, only create a new, empty assay attachment. There should be at least two error messages: Production date can't be empty and the assay needs a LOINC code as well as either result or a file attachment.
But only the former messages actually shows up:

grafik

It looks like everything's okay except for the production date. Only after filling in a value on the date field, the form shows the additional errors in the assay attachment.

The relevant code is here:

https://github.com/instedd/cdx/blob/4c31f3011f7a2ac143b7ec292f1dcd769e4ed56b/app/models/sample_form.rb#L83-L89

leandroradusky:
Close #1476.

Before, if date produced was throwing a validation error and other error were present in the sample form they weren't displayed.

Now, when several error are present alongside with date produced, all errors are shown:

image

As a free corollary errors are highlighted for assays because the check is done, thing that wasn't happening before.

image

### DIFF
--- a/app/models/sample_form.rb
+++ b/app/models/sample_form.rb
@@ -35,7 +35,9 @@ class SampleForm
   delegate :uuid, :assay_attachments, :notes, :qc_info, to: :sample
 
   validates_presence_of :date_produced
-
+  validates_each :assay_attachments, :notes do |record, attr, value|
+    record.errors.add(attr, "are invalid") unless value.all?(&:valid?) && !record.errors.include?(attr)
+  end
   def self.for(sample)
     new.tap do |form|
       form.sample = sample
@@ -87,18 +89,12 @@ class SampleForm
 
     # validate forms. stop if invalid
     form_valid = self.valid?
-    self.assay_attachments.each do |aa|
-      form_valid &= aa.valid?
-    end 
-    self.notes.each do |n|
-      form_valid &= n.valid?
-    end 
+    return false unless form_valid
     
     # validate/save. All done if succeeded
     is_valid = sample.save
     return true if is_valid
 
-    # copy validations from model to form (form is valid, but model is not)
     sample.errors.each do |key, error|
       errors.add(key, error) if self.class.shared_attributes.include?(key) && !errors.include?(key)
     end
--- a/app/models/sample_form.rb
+++ b/app/models/sample_form.rb
@@ -87,17 +87,13 @@ class SampleForm
 
     form_valid = self.valid?
     sample_valid = sample.valid?
-    # copy validations from patient to form (form is valid, but patient is not)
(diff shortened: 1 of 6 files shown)


app/models/sample_form.rb
```diff
@@ -87,15 +87,22 @@ def save
 
     # validate forms. stop if invalid
     form_valid = self.valid?
+    self.assay_attachments.each do |aa|
+      form_valid &= aa.valid?
+    end 
+    self.notes.each do |n|
+      form_valid &= n.valid?
+    end
```
ysbaddaden, line 95: Hum, ActiveRecord has `validates_associated` and Sample will validate these two associations using it, but here we only have an ActiveModel so we don't have this helper. Yet, we can use [validates_each](https://api.rubyonrails.org/v4.2/classes/ActiveModel/Validations/ClassMethods.html#method-i-validates_each), so that `valid?` would always tell us whether the form is valid or not:

```ruby
class SampleForm
  validates_each :assay_attachments, :notes do |record, attr, value|
    errors.add(attr, "are invalid") unless value.all?(&:valid?)
  end
end
```

leandroradusky:
Question before squash and merge... Given the conversation that I have today in slack with @straight-shoota maybe I can use this PR to solve this in the other (2) places where is happening, Or I should create another issue TBD after? @diegoliberman .

straight-shoota:
Ooh, I didn't realize this fix wasn't already merged. I think it would be good to amend the fixes for the other forms. They're exactly the same bug, thus it makes sense to fix it in one go.

diegoliberman:
I agree to fix them now.

leandroradusky:
There was a mislead here, before (when changes were approved) we were adding the same validation present in `sample.rb` into the `sample_form.rb` file, then it was working ok but was not repairing the logic flaw.

The good fix is to check the validity of the model **&** the form, not to repeat checks. 

```
form_valid = self.valid?
sample_valid = sample.valid?
...
return false unless form_valid && sample_valid 
```

Now the logic its solved (in samples, batches & patients).

Also, uniqueness of `batch number` & `isolate name` pairs within batches are checked as in #1498 by suggestion of johannes.

straight-shoota:
The changes from https://github.com/instedd/cdx/pull/1498 shouldn't be merged in this PR but in #1498. Before that we need to make sure to use the correct collation in all database systems (https://github.com/instedd/cdx/issues/1517).

Issue #1566, opened 2022-03-29
Change the "transfer samples" modal into a full page form


diegoliberman:
### Context
The transfer samples modal doesn't make sense when thinking on the scanning QR codes flow.  It used to make sense when the modal was working on the samples selected in the background list.  Now, the sample transfer will be more independent, and it makes sense to make it a full page form.  It will also provide value as there will be more space for showing more data if needed.

[This](https://projects.invisionapp.com/d/main#/console/4103589/465342627/preview) is the full page design.

### Game plan
- [x] step 1: no change in behavior (keep form, autocomplete, ...);
  - [x] step 2:
    - [x] add `samples#show` (JSON) to request a Sample by UUID (or use autocomplete?);
    - [x] request sample when a valid UUID is scanned / typed / pasted;
    - [x] validate the sample and either add it to the form, or display an error/notice message + clear input.

### Acceptance criteria
1. When I open the transfer form, the institution field is the first one I see.
2. After selecting an institution, the focus is placed automatically into the empty field for entering a sample ID.
3. In the empty field, I see a hint `Enter, paste or scan the sample ID`.
4. **Optional**: if trivial/easy, when I start typing, I see an autocomplete with suggested IDs.
5. If I enter paste or scan a sample ID that matches with a sample within the current institution, the sample is added to the list of samples to be transferred, the field gets emptied and receives the focus.
6. If we implemented the autocomplete, and after typing some characters, there are no matchings, the autocomplete shows the message `No matches found` as the unique/unselectable option.
7. If I paste or scan a sample ID that doesn't match, a feedback message appears below the field `The sample id xxxxx wasn't found`, the field gets emptied and receives the focus.
8. The same logic than described for `No matches` applies if the sample ID has already been added to the list of samples to be transferred.  In this case, the message is `The sample ID xxxx can't be added twice`.
9. Same logic for message `The sample ID xxxx can't be added because it's a QC`.
10. I see a button for cancelling the transfer action, which takes me back to the list of transfers.
11. I see a button called `Transfer` which executes the transferring action.  I see a success message after pressing it. 

### Considerations
- [ ] Some scanners might include a carriage return at the end of the string.  We should handle this use case by correctly reading the sample ID and searching for matches without the carriage return as part of the search string.

straight-shoota:
> After selecting an institution, the focus is placed automatically into the empty field for entering a sample ID.

[...]

PR #1594, opened 2022-04-12
Add `TransferPackagesController#find_sample`

straight-shoota:
This is a prerequisite feature for #1566. It was easy to extract, so I figured a separate PR makes review easier.

There isn't anything on its own though, except that it provides an API endpoint for querying samples by UUID. The endpoint is located in `TransferPackageController` which will be used for the sample transfer screen as well. Its behaviour is specific to the sample transfer action, that's why it's not in the more generic `SamplesController`.

### DIFF
--- a/app/controllers/transfer_packages_controller.rb
+++ b/app/controllers/transfer_packages_controller.rb
@@ -1,0 +1,37 @@
+class TransferPackagesController < ApplicationController
+  def find_sample
+    @navigation_context = NavigationContext.new(nil, params[:context])
+
+    uuid = params[:uuid]
+    full_uuid = uuid.size == 36
+    @samples = Sample
+      .within(@navigation_context.entity, @navigation_context.exclude_subsites)
+      .joins(:sample_identifiers).where(
+        (full_uuid ? "sample_identifiers.uuid = ?" : "sample_identifiers.uuid LIKE concat(?, '%')"),
+        uuid
+      )
+      .order("created_at DESC")
+      .limit(5)
+
+    @samples = check_access(@samples, READ_SAMPLE)
+
+    if full_uuid && @samples.any?(&:is_quality_control?)
+      render json: { error: "Sample #{uuid} is a QC sample and can't be transferred." }
+      return
+    end
+
+    render json: { samples: samples_data(@samples) }
+  end
+
+  private
+
+  def samples_data(samples)
+    samples.map { |sample|
+      next if sample.is_quality_control?
+      {
+        uuid: sample.uuid,
+        hasQcReference: sample.has_qc_reference?,
+      }
+    }.compact
+  end
+end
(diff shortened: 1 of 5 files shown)


app/controllers/transfer_packages_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    uuid = params[:uuid]
+    full_uuid = uuid.size == 36
+    @samples = Sample
+      .within(@navigation_context.entity, @navigation_context.exclude_subsites)
+      .joins(:sample_identifiers).where(
+        (full_uuid ? "sample_identifiers.uuid = ?" : "sample_identifiers.uuid LIKE concat(?, '%')"),
+        uuid
+      )
+      .order("created_at DESC")
+      .limit(5)
```
ysbaddaden, line 14: Maybe we could introduce a `Sample.autocomplete(uuid)` scope, so we could do:

```ruby
@samples = Sample.within(...).autocomplete(params[:uuid])
```

app/controllers/transfer_packages_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    @samples = check_access(@samples, READ_SAMPLE)
+
+    if full_uuid && @samples.any?(&:is_quality_control?)
+      render json: { error: "Sample #{uuid} is a QC sample and can't be transferred." }
+      return
+    end
```
ysbaddaden, line 21: Let's say I'm autocompleting: I type `ae` see my sample and select it. It was a QC sample. I won't get an error?

ysbaddaden (commented):
I fail to see why this couldn't be `samples#autocomplete`? Except for the 'refuse QC sample' it's quite generic, and we could reuse it from anywhere in the UI. Sure, you only return what's interesting for this view, but returning more data wouldn't hurt.

Maybe the "refuse QC sample" check should be done on the frontend, showing an error message when trying to select the sample. I'm not fond of the backend refusing to autocomplete QC samples, and the user be left in the dark of 'why isn't it autocompleting'?

straight-shoota:
> I fail to see why this couldn't be `samples#autocomplete`? Except for the 'refuse QC sample' it's quite generic, and we could reuse it from anywhere in the UI. Sure, you only return what's interesting for this view, but returning more data wouldn't hurt.

What benefit it would bring? Generalization is nice, sure. But it doesn't always make sense.
We'd be adding more complexity in the frontend by duplicating business logic that already exists in the backend.

I don't like that. Having a separate endpoint for this specific use case is cheap.
We can add a more generic autocomplete endpoint, too (if we need that).

ysbaddaden:
I'm okayish with a specific controller action, but still not convinced. It doesn't render a HTML fragment to be injected in the page as-is, but returns an Array<Sample> as JSON that will need to be processed by the frontend anyway, so checking for QC doesn't seem like much more processing to me.

With a generic autocompletion, we could introduce a _reusable_ INPUT for searching samples, just like we do for institutions for example. In this case, we'd just add a JS callback to display an error when trying to choose a QC sample, and that's about it.

Let's implement the frontend right in this PR (and make it a draft for the time being). That would help.

straight-shoota:
I'm closing this. Let's just merge this in #1625 directly.

Issue #1542, opened 2022-03-15
Modal window inherits style from button location context


straight-shoota:
<img width="619" alt="image" src="https://user-images.githubusercontent.com/25181052/158152731-4ed3c620-44f6-4612-9a4e-bd86ed1bc06a.png">

> Why are the select placeholder and values upper-cased?

The React component is mounted at the button location, which is inside a table header cell, and there's `.table th { text-transform: uppercase; }`. 🤦‍♂️
The checkbox label has an explicit `text-transform: none`, that's why it in normal case.

These modals should really be mounted outside the table. I had already noticed a similar issue with the confirmation modal which apparently inherits the `title` attribute of the parent table cell.

_Originally posted by @straight-shoota in https://github.com/instedd/cdx/issues/1523#issuecomment-1066701744_

PR #1596, opened 2022-04-12
Fix tableheader element type

straight-shoota:
This fixes the primary effect of #1542.

Changing tableheader element from `th` to `td` removes some special style properties. The transfer modal does no longer have uppercased text which it inherited from the `th`.

It would probably still be better to mount the modal somewhere else (or do a complete reset somewhere), but this is already a quick and simple fix for the main issue. And it makes sense regardless to use `td` here. The table cell is not a column header.

### DIFF
--- a/app/assets/stylesheets/_tables.scss
+++ b/app/assets/stylesheets/_tables.scss
@@ -12,6 +12,29 @@
       color: $gray4;
     }
 
+    .table-title {
+      font-size: 120%;
+      font-weight: bold;
+      color: $black;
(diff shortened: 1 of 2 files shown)


ysbaddaden:
@straight-shoota you also changed some CSS, is that on purpose?

straight-shoota:
Yes, it mirrors the HTML change and moves the nested rules from `th` to `td`.

ysbaddaden (approved):

PR #1597, opened 2022-04-12
Fix icon class applying icomoon font to its content instead of just the icon pseudoelement

straight-shoota:
There is a style rule for classes that start with `icon-` and it sets a lot of font properties for these classes.
This is fine when icons are used like `<i class="icon-foo">` because the element itself has no content.

But when the icon class is used on an element with content, that content is displayed alongside the icon image (which comes into existence via pseudoelement).

An example of this can be seen here:

https://github.com/instedd/cdx/blob/eecef1c84a246833390dca73e039ba9f6766fb2a/app/assets/javascripts/components/sample_transfer.js.jsx#L96-L99

Note the `notification-text` class around the text content. This class is necessary to reset the font properties to something that matches the surrounding. This is a workaround, but we can fix the bug instead.

This PR moves the original font properties to the pseudoelement with the icon image, so the styles of the icon element' content won't be modified. Thus we can remove `notification-text` (requires #1596).


ysbaddaden:
Note that the SCSS is generated by icomoon.

It looks like this has been abused, here. The HTML should use a nested `<i>` or `span` element instead, to properly mark where the icon should be:

```html
<div className="col qc-info-message"> 
  <input type="hidden" name="includes_qc_info" value="false" /> 
  <i class="icon-info-outline icon-gray"></i>
  {infoMessage}
</div>
```

PR #1588, opened 2022-04-07
Fix: add timestamps to transfer packages table

ysbaddaden:
Fixes https://github.com/instedd/cdx/pull/1574#pullrequestreview-935424140

PR #1599, opened 2022-04-13
Move receiver and sender institutions from `SampleTransfer` to `TransferPackage`

### LINKED ISSUE #1595: Consolidate `SampleTransfer` and `TransferPackage`
`TransferPackage` has taken over some of the shared data between multiple sample transfers that happen together.

This is the current situation of metadata fields of both models:

```
TransferPackage:
  receiver_institution
  recipient
  includes_qc_info

SampleTransfer:
  sender_institution
  receiver_institution  
```

`receiver_institution` is duplicated on both models, `sender_institution` is not.
It is already pretty confusing that institution references are handled differently.

Duplication should usually be avoided. In this case, the values never change, so it wouldn't cause any consistency issues. But still, we can easily avoid it. Sample transfers refer to the transfer package and the shared metadata can be easily aquired from there.

I'd like to move *all* shared metadata between sample transfers of the same package to `TransferPackage`. All samples have the same sender and receiver when they're in one package. So there should be no need to store the information for each individual sample transfer instance. They can just share it in the package.

The refactor would include adding `sender_institution` to `TransferPackage` and dropping both institution properties from `SampleTransfer`.
Overall, this simplifies the data model.

I'm not sure if the sample transfer workflow is already actively used in production. If yes, we have to think about how to migrate data. The easiest migration would be to create an individual `TransferPackage` for each existing `SampleTransfer` and copy the institution properties there. We could consider merging multiple instances of `SampleTransfer` that have been created at the same instant (and identical sender/receiver properties), but that might not really be necessary.

straight-shoota:
Resolves #1595

Builds on top of #1588 to avoid merge conflicts

### DIFF
--- a/app/models/transfer_package.rb
+++ b/app/models/transfer_package.rb
@@ -3,13 +3,18 @@ class TransferPackage < ApplicationRecord
   belongs_to :receiver_institution, class_name: "Institution"
   has_many :sample_transfers
 
+  # TODO: remove these after upgrading to Rails 5.0 (belongs_to associations are required by default):
+  validates_presence_of :sender_institution
+  validates_presence_of :receiver_institution
+
   after_initialize do
     self.uuid ||= SecureRandom.uuid
   end
 
-  def self.sending_to(institution, attributes = nil)
+  def self.sending_to(sender, receiver, attributes = nil)
     create!(attributes) do |package|
-      package.receiver_institution = institution
+      package.sender_institution = sender
+      package.receiver_institution = receiver
     end
   end
 
--- a/app/models/transfer_package.rb
+++ b/app/models/transfer_package.rb
@@ -21,7 +21,6 @@ class TransferPackage < ApplicationRecord
   def add!(sample)
     transfer = sample_transfers.create!(
       sample: sample,
-      receiver_institution: receiver_institution,
     )
 
     if sample.batch
--- a/app/models/transfer_package.rb
+++ b/app/models/transfer_package.rb
@@ -1,4 +1,5 @@
 class TransferPackage < ApplicationRecord
+  belongs_to :sender_institution, class_name: "Institution"
   belongs_to :receiver_institution, class_name: "Institution"
   has_many :sample_transfers
 
--- a/app/models/transfer_package.rb
+++ b/app/models/transfer_package.rb
@@ -23,7 +23,7 @@ class TransferPackage < ApplicationRecord
           where(receiver_institution_id: institution.id)
         }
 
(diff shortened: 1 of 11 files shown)


app/models/transfer_package.rb
```diff
@@ -1,21 +1,38 @@
...
+          else
+            where(arel_table[:sender_institution_id].eq(institution.id).or(arel_table[:receiver_institution_id].eq(institution.id)))
+          end
+        }
+
+  scope :with_receiver, ->(institution) {
+          where(receiver_institution_id: institution.id)
+        }
+
+  def self.sending_to(sender, receiver, attributes = nil)
```
ysbaddaden, line 26: Hum, the method signature doesn't read well anymore. Maybe it doesn't make sense to use `#sending_to` and we should find a better name for that constructor?

db/migrate/20220408103851_add_sender_institution_to_transfer_packages.rb
```diff
@@ -0,0 +1,18 @@
+class AddSenderInstitutionToTransferPackages < ActiveRecord::Migration
+  def up
+    add_reference :transfer_packages, :sender_institution, null: true
+
+    # Fill sender_institution on existing packages
+    TransferPackage.find_each do |package|
+      package.update_columns(
+        sender_institution_id: package.sample_transfers.first.sender_institution_id,
+      )
+    end
```
ysbaddaden, line 10: Using models inside migration is often a gamble. For this reason I'm trying more and more to use only SQL for data migrations.

You're adding a column, so you must tell ActiveRecord to reload the model's definition from the database before using it, because Rails' may have already load it (especially in production).

Quick performance optimization: `#take` is better than `#first` to pick whatever row (it avoids an `ORDER BY`).

This is not worth it for this specific migration, but for learning purposes, I'm wondering if we could write this as a simple SQL request? With PostgreSQL we could (using `WITH` or something), but with MySQL 5.7? :thinking:

ysbaddaden: Answering myself: it looks like this is possible with MySQL 8.0 :smile: but not before :disappointed:

ysbaddaden (approved):

ysbaddaden (approved):
LGTM!

straight-shoota:
Note: I'm not dropping the references on `sample_transfers` just yet. They're not used in the app anymore, but it's easier to roll back in case something goes wrong with the migration. We can drop them as soon as we're sure everything works find with the migration to `transfer_package`. Dropping those columns won't have any impact in the app, just the database.

PR #1623, opened 2022-04-26
Improve the ux for missing mandatory fields in all forms

### LINKED ISSUE #1065: Improve the UX for missing mandatory fields in all forms
The look and feel of the listed fields when missing mandatory details at the time of creation of a new device can be improved (see image attached). 

I'd suggest
.- highlighting the areas to which users should pay attention, to save them time and the effort of comparing the names of the fields and the fields listed at the top
.- ordering the listed fields following the order in which they appear on the form, so it's intuitive and saves time for users if they need to double-check if they have covered it all

Screen Shot 2021-03-18 at 10 47 49 AM

Screen Shot 2021-03-18 at 10 34 09 AM

leandroradusky:
Closes #1065.

In order to full cover all the places/conditions this should be implemented for I went through the `*_form.rb` files checking the `validate_*` conditions, plus all the date formats plus I took a look at the logic. After this I went through all the forms but **I didn’t find a systematic way to check that the coverage was full, this needs an in-depth QA**. 

Implemented in:
- Samples,
- Sample transfer (modal),
- Sample transfer confirm (modal),
- Batches, 
- Sites,
- Devices,
- Roles

From the implementation POV, now the `error` class was renamed to `errors-field` (in plural, since there will be an `<ul>` element displaying `<li>`s for each error present in the field. Also, the `input-required` CSS class was deleted, and the rails `field_with_errors` class is used to highlight fields with errors. When it is not automatically applied by rails, it is assigned programmatically as was done with `input-required` (i.e. in the sample transfer modal). 

Also, it wasn’t reachable for the modals if present in` _errors.scss`, all this had to be moved to `_forms.scss` and now is reached both in the forms and in the modals. Is this correct?

As an example, in batches, errors are now shown like this:

image

Another example, in modals:

image

And a special mention to Roles, which maybe needs a new mockup? I didn’t dare to enter within the react component for the policies and I solved it showing errors here:

image

And another special mention to **Alert Groups**, that have a different way of showing errors, then I guess it's not in use? I didn’t modify this form:

image

In my ideal implementation 

```
= field_errors @form, :field
```

shouldn't be call for each field that could present errors, but I didn't found a way to solve it that way (append DOM elements to `field_with_errors` div, cab be done without js?).

### DIFF
--- a/app/assets/javascripts/components/sample_transfer.js.jsx
+++ b/app/assets/javascripts/components/sample_transfer.js.jsx
@@ -134,7 +134,8 @@ var SampleTransferModal = React.createClass({
 
   validateForm: function(e) {
     if (this.state.institutionId == null){
-      $(".institution-select").addClass("input-required");
+      $(".institution-select").addClass("field_with_errors");
+      $("#institution_id_error").removeClass("hidden-error");
       e.preventDefault();
     }
   },
@@ -147,7 +148,7 @@ var SampleTransferModal = React.createClass({
             <div className="col pe-4"><label>Institution</label></div>
             <div className="col">
               <CdxSelect className="institution-select" name="institution_id" items={this.props.institutions} value={this.state.institutionId} onChange={this.changeInstitution} />
-              <span className="error"><div className="icon-error icon-red" /> Institution can't be blank</span>
+              <ul className="errors-field hidden-error" id="institution_id_error"><li><i className="icon-error icon-red" /> Institution can't be blank</li></ul>
             </div>
           </div>
           <div className="row">
--- a/app/assets/javascripts/sample_transfer_confirm.js.jsx
+++ b/app/assets/javascripts/sample_transfer_confirm.js.jsx
@@ -59,12 +59,13 @@ var SampleTransferConfirmModal = React.createClass({
 
   checkUUID: function(event) {
     let uuidCheck = document.getElementById("uuid_check")
+    let uuidCheckError = document.getElementById("uuid_check_error")
     let submitButton = document.getElementById("samples-transfer-form-submit")
 
     let value = uuidCheck.value
     if(value.length < 4 && value.match(/^[0-9a-f]*$/i)) {
       // don't show error for incomplete input as long as the format is valid
-      uuidCheck.classList.remove("input-required")
+      uuidCheck.classList.remove("field_with_errors")
       submitButton.disabled = true
       return;
     }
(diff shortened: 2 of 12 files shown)


straight-shoota:
The error you see in *Alert Groups* is just a client-side validation error created by the browser (in this case from `<input type="text" required />`). These validations make sense to quickly show the user that they need to so something (before submitting the entire form).
It's possible to hook into the Browser API in order to inject these validation errors into the `errors-field` element. That would be a nice further enhancement, but it's not that important. Definitely not for this PR.

diegoliberman:
@straight-shoota @ysbaddaden we need to define what to do with this PR.  If we got to a point where there is already an improvement to the current UX, then I'd just merge it.  If there are things to be improved even further, I'd create a new ticket, which will probably be addressed in the future, as we need to prioritize VQA Boxes.

straight-shoota:
The current state is not mergeable.
The PR removes the combined display of all errors in a form. But it only adds display of errors on the field level for some fields.
We need to transform all forms to show errors for all fields. All form fields must consist of label, value and error (error is optional of course). This standardized structure should be extracted into a helper.
The question is whether we build a custom field structure helper for this, or use the simple_form gem.
@ysbaddaden WDYT?

ysbaddaden:
I just reviewed the pull request, and it's already an improvement:

- it no longer renders a flat list of errors before the form;
- it displays the error along with the invalid fields;
- it reuses the `#field_errors` helper to achieve so.

Having another helper to generate all fields (div > label + input + error) would be much better, but that's a lot more work, and we need to make technical decisions ([simple_form](https://rubygems.org/gems/simple_form) vs custom helper).

I propose to merge this PR and create a new tech-debt issue to continue to improve the forms. We could introduce such a helper while we work on the forms for the boxes.

ysbaddaden (approved):

straight-shoota:
@ysbaddaden Either I'm missing something or you didn't get the major problem with this patch. it removes the form-wide error display. That means, errors are only displayed when the field which has an error uses a dedicated `field_errors`. For all other fields, error messages would be invisible.
For example, in the `batch` form, if there's an error in the `inactivation_method` field, the form would be invalid but the user wouldn't see any error message. That's inacceptable.
For all forms that don't have a form-wide error display, we must make sure that every possible error has a place to be visible.

jkicillof:
There's an example of input error [here](https://projects.invisionapp.com/d/main#/console/4103589/121513966/preview), let's keep the input text in black and red for input border and error message.
One more thing, it would be great that in case of an error the page scrolls up to the first one

PR #1625, opened 2022-04-26
Add `TransferPackagesController#create` action

### LINKED ISSUE #1566: Change the "transfer samples" modal into a full page form

### LINKED ISSUE #1567: Change the "confirm receipt" modal into a full page form

straight-shoota:
grafik

Resolves #1566

Based on https://github.com/instedd/cdx/pull/1594

### DIFF
--- a/app/assets/javascripts/components/sample_transfer_selector.js.jsx
+++ b/app/assets/javascripts/components/sample_transfer_selector.js.jsx
@@ -1,0 +1,213 @@
+var SampleTransferSelector = React.createClass({
+  getInitialState: function() {
+    return {
+      includeQcInfo: this.props.includeQcInfo,
+      samples: this.props.samples,
+      search: "",
+      error: null,
+      status: "input",
+    };
+  },
+
+  showQcWarningCheckbox: function(selectedSamples) {
+    const haveQc = selectedSamples.filter((sample) => sample.hasQcReference === true).length
+    const missingQuantity = selectedSamples.length - haveQc
+    return (
+      <div>
+        {haveQc > 0 && this.includeQcInfoCheckbox()}
+        {missingQuantity > 0 && this.qcInfoMessage(missingQuantity, selectedSamples)}
+      </div>
+    )
+  },
+
+  renderSamples: function() {
+    const selector = this;
+
+    function renderSample(sample, i) {
+      function handleClick(e) {
+        e.preventDefault();
+        selector.removeSample(sample);
+      }
+
+      let name = `transfer_package[sample_transfers_attributes][${i}][sample_uuid]`;
+      return <div className="sample-transfer-preview">
+        <div dangerouslySetInnerHTML={{ __html: sample.preview }}></div>
+        <input type="hidden" name={name} value={sample.uuid} />
+        <a href="#" className="sample-transfer-preview__remove" title="Remove sample" onClick={handleClick}>
+          <i className="icon-close" />
+        </a>
+      </div>
+    }
+
+    const listItems = this.state.samples.map(renderSample)
+    return ({ listItems });
+  },
+
+  removeSample: function(sample) {
+    let samples = this.state.samples.filter((value) => value != sample)
+    this.setState({
+      samples: samples
+    });
+  },
+
+  qcInfoMessage: function(missingQuantity, selectedSamples) {
+    const infoMessage = (missingQuantity > 0 && missingQuantity != selectedSamples.length)
+      ? `There is no Quality Control (QC) info available for ${missingQuantity} ${missingQuantity === 1 ? 'sample' : 'samples'}`
+      : "There is no Quality Control (QC) info available for these samples"
+
+    return (
+      <div className="icon-info-outline icon-gray qc-info-message">
+        <input type="hidden" name="includes_qc_info" value="false" />
+        <div className="notification-text">{infoMessage}</div>
+      </div>
+    )
+  },
+
+  handleChange: function(e) {
+    let search = e.target.value;
+
+    this.setState({ search: search });
+
+    function isUUID(str) {
+      const regexExp = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/gi;
+
+      return regexExp.test(str);
+    }
+
+    if (isUUID(search)) {
+      this.loadSample(search)
+    }
+  },
+
+  onKeyDown: function(e) {
+    this.setState({
+      error: null,
+    })
+    if (e.key == "Enter") {
+      e.preventDefault();
+
+      this.loadSample(this.state.search)
+    }
+  },
+
+  loadSample: function(uuid) {
+    url = "/transfer_packages/find_sample"
+    console.log(this.state)
+    this.setState({
+      error: null,
+      status: "loading",
+    });
+    $.ajax({
+      url: url,
+      dataType: 'json',
+      type: 'GET',
+      data: { context: this.props.context, uuid: uuid },
+      success: function(data) {
+        if (data.error) {
+          this.setState({
+            error: data.error,
+            status: "input",
+          })
+        } else if (data.samples.length == 1) {
+          let sample = data.samples[0]
+          if (sample.error) {
+            this.setState({
+              error: sample.error,
+              state: "error"
+            })
+          } else if (this.addSample(sample)) {
+            this.setState({
+              search: "",
+              status: "input",
+            })
+          } else {
+            this.setState({
+              status: "input",
+            })
+          }
+        } else if (data.samples.length == 0) {
+          this.setState({
+            error: `No sample found for ${uuid.length == 36 ? "UUID" : "UUID prefix"} ${uuid}`,
+            status: "error",
+          })
+        } else {
+          this.setState({
+            error: `Multiple samples found for ${uuid.length == 36 ? "UUID" : "UUID prefix"} ${uuid}`,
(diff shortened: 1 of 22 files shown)


app/controllers/transfer_packages_controller.rb
```diff
@@ -1,4 +1,59 @@
...
+    end
+
+    # FIXME: permit! shouldn't be necessary here
+    params.permit!
+
+    @transfer_package = TransferPackage.new(params)
+
+    if @transfer_package.sample_transfers.empty?
+      @transfer_package.errors.add :sample_transfers, "Must not be empty"
+    end
```
ysbaddaden, line 44: This could be a validation in the model, then you'd just `if @transfer_package.save` but it means that we can't create a package without sample transfers, which may break the current solution (be doable when we eventually drop the current solution).

app/controllers/transfer_packages_controller.rb
```diff
@@ -17,16 +72,54 @@ def find_sample
 
   private
 
+  def transfer_package_params
+    params.require(:transfer_package).permit(
+      :receiver_institution,
+      :recipient,
+      :includes_qc_info,
+    ).tap do |whitelisted|
```
ysbaddaden, line 80: `whitelisted` => `allowed`

But with the proposed simplification of `#create` you wouldn't need any manipulation of the params.

app/controllers/transfer_packages_controller.rb
```diff
@@ -1,4 +1,59 @@
...
+  end
+
+  def create
+    params = transfer_package_params
+    params.merge!({
+      sender_institution: @navigation_context.institution,
+    })
+    if id = params.delete(:receiver_institution).presence
+      params[:receiver_institution] = Institution.find(id)
+    end
```
ysbaddaden, line 21: The whole method is _very_ complex, and mostly because of a heavy manipulation of params. You could improve readability by relying more on rails' conventions, and using records.

For example:

- use a `receiver_institution_id` param name;
- instantiate the package through `@navigation_context.institution` (declare the association if it doesn't already);
- build the package object immediately, add sample transfers to it, and eventually save it all;
- rely on validations to do data integrity checks (e.g. sample transfer can't have a QC sample);
- we don't need attributes for sample transfers but the sample UUID, so could the frontend send a mere `params[:sample_uuids]` array?

Then the method would be more straightforward, pushing data details into the models:

```ruby
@package_transfer = @navigation_context
  .institution
  .package_transfers
  .build(package_transfer_params)

samples = Sample
  .within(@navigation_context.institution)
  .find_all_by_any_uuid(params[:sample_uuids])

samples.each do |sample|
  sample_transfer = @package_transfer.sample_transfers.build(sample: sample)

  unless authorize_resource?(sample, UPDATE_SAMPLE)
    sample_transfer.errors.add :sample, :unauthorized
  end
end

# NOTE: could push this validation to PackageTransfer
if @package_transfer.sample_transfers.empty?
  @package_transfer.errors.add :sample_transfers, :empty
elsif @package_transfer.save
  redirect_to # ...
  return
end

# ...
render "new"
```

app/controllers/transfer_packages_controller.rb
```diff
@@ -17,16 +72,54 @@ def find_sample
...
+        end
+      else
+        whitelisted[:sample_transfers_attributes] = {}
+      end
+    end
+  end
+
+  def available_institutions
+    Institution.where.not(id: @transfer_package.sender_institution.id)
+  end
```
ysbaddaden, line 101: This could be a helper method, so you wouldn't need to instantiate an `@available_institutions` ivar. Just add `helper_method :available_institutions` to the controller.

app/controllers/transfer_packages_controller.rb
straight-shoota: We'd have to refactor how `TransferPackage#add!` works because you can't save the instance before adding sample transfers, but you need sample transfers to save it. This was just the easiest path to get done.

app/controllers/transfer_packages_controller.rb
straight-shoota: Good points. I'm not happy about it either. Ended up with this as a reasonably find solution after already trying some alternatives. I'll give it another go and see what can be improved.

> use a `receiver_institution_id param` name;

IIRC this didn't seem to work well with the CdxSelect field. 

> * build the package object immediately, add sample transfers to it, and eventually save it all;
> * we don't need attributes for sample transfers but the sample UUID, so could the frontend send a mere params[:sample_uuids] array?

The tradeoff for that is to do some additional things for the sample transfers (such as error handling). Passing in the sample transfers as nested attributes is actually a nice solution. It would be great if it didn't require some extra handling. Maybe I can get rid of that.

straight-shoota: I achieved improvements on most of the mentioned points.

* I don't think it makes much sense to introduce an association just for creating a transfer package from it. Setting `sender_institution` explicitly is completely fine. I changed it to not go through params, though.
* I'm keeping the nested attributes for sample transfers because it's a neater solution IMO. It easily allows editing and removing transfers if that should ever be necessary.
* Changed sample reference to use the (internal) ID. There's no need for UUID here. It's just more overhead.

app/controllers/transfer_packages_controller.rb
ysbaddaden: In that case I just memoize the ivar :)

app/assets/javascripts/components/sample_transfer_selector.js.jsx
```diff
@@ -0,0 +1,214 @@
...
+    if (e.key == "Enter") {
+      e.preventDefault();
+
+      this.loadSample(this.state.search)
+    }
+  },
+
+  loadSample: function(uuid) {
+    url = "/transfer_packages/find_sample"
+    console.log(this.state)
```
ysbaddaden, line 96:
```suggestion
    console.log(this.state)
```

spec/support/feature_spec_helpers.rb
```diff
@@ -25,6 +25,10 @@ def sign_in(user)
     end
 
     goto_page HomePage do |page|
+      unless page.has_form?
+        page.user_menu.icon.click
+        page.user_menu.logout.click
+      end
```
ysbaddaden, line 31: Did you run into an issue to sign_in while already sign_in? What about an explicit sign_out or maybe testing the currently loaded URL instead?

ysbaddaden (approved):
Congrats, the controller is much nicer to follow, now!

app/controllers/transfer_packages_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+    if @transfer_package.sample_transfers.empty?
+      @transfer_package.errors.add :sample_transfers, "Must not be empty"
+    end
+
+    @transfer_package.sample_transfers.each do |sample_transfer|
+      sample = sample_transfer.sample
+      raise "User not authorized for transferring sample #{sample.uuid}" unless authorize_resource?(sample, UPDATE_SAMPLE)
+    end
+
+    if @transfer_package.errors.empty? && @transfer_package.save
```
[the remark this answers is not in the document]
straight-shoota: Oh, I think we can just call `validate` in any case to make sure that other errors show up as well.
And consider a refactor afterwards.

spec/support/feature_spec_helpers.rb
straight-shoota: Yes, if a user is logged in, there is no login form. We need to log out first.

I think this is fine. This is a high level action to sign in a user. The `sign_in` helper takes care of that. When I'm logged in and I tell you to sign in, it naturally requires to sign out first.
Do you see any downside to this implicit approach?

PR #1638, opened 2022-05-04
Fix timing edge case in `password_reset` spec

straight-shoota:
A weird failure has started to show up in CI builds: https://github.com/instedd/cdx/runs/6276901123?check_suite_focus=true

It seems to be caused by a too tight time interval. `user_password_expire_in_months: 3` and we're testing for exactly 3 months ago, thus creating an edge case and the password is still considered fresh. When the last password change in the test is set to more than 3 months ago, it succeeds.

Also removes hard coded time interval in the feature steps defintion.


matiasgarciaisaia (approved):
LGTM! 👌

PR #1639, opened 2022-05-05
Add `mysql-client` to `Dockerfile-dev`

straight-shoota:
This allows the use of `rails db` to investigate the database.


ysbaddaden (approved):
Though, I usually do `docker-compose exec db mysql cdp_development` —which is too long to type so I made a `dk db` shortcut :smile:

PR #1640, opened 2022-05-05
Allow to create batches with no samples

### LINKED ISSUE #1604: [Out] Allow to create empty batches
As a lab technician, I want to create empty batches, so that I can select them when I'm creating a box.

Empty batches didn't make sense at first.  But now, it does make sense thinking on the physical world flow. 

1. A virus is grown in a big container, and we call this a batch.
2. We know in advance that the batch will contain N samples, and it could make sense to create the N samples at the same time as the batch.
3. However, it could be very useful to create the samples ad hoc.  Because the technician, in the physical world, won't be aliquoting (see dictionary) as soon as the batch is created.  And it's useful to create the batch in the system to acknowledge its existence, and then when needed, to add samples.  When samples are added, we'll provide the possibility of printing those sample's labels. This will fit better into reflecting real life scenarios: 1 day the batch is grown, another day some samples are aliquoted, another day more samples are aliquoted, etc.  

All of this context explanation, is just for a simple requirement:
- [ ] Remove the validation that prevents a batch to be saved without defining the quantity of samples.

Although it would be great to allow adding N samples at once from within the Batch, the current goal is to achieve a good experience for the boxes use case.  In that journey, N samples will be created automatically when a new box is saved.

### Dictionary
1. Aliquoting is the action of taking the virus from the batch, and putting it into a tube which represents the sample.  When this action is done, they could also dilute the virus which is also known as the virus to have a certain concentration.   

### Thoughts
I imagine that in the future it might be useful to allow defining the batch capacity.  And CDx could calculate how much has been already used in samples, and how much remains in the batch.

leandroradusky:
Closes #1604. 

Based on the discussion on the issue, only the check of values greater than zero was removed from the form. Now is possible to create batches with zero samples:

Screenshot from 2022-05-05 10-17-45

For sake of testing negative values were attempted, the error is shown in this way:

Screenshot from 2022-05-05 10-15-29

Which is different of the standard way (it responds to `min="0"` in the input element). No actions were taken to change this (should we change it?)

### DIFF
--- a/app/models/batch_form.rb
+++ b/app/models/batch_form.rb
@@ -32,6 +32,7 @@ class BatchForm
   delegate :id, :new_record?, :persisted?, to: :batch
 
   validates_presence_of :date_produced
+  validates_numericality_of :samples_quantity, greater_than_or_equal_to: 0, message: "value must be greater or equal to 0", if: :creating_batch?
 
   def self.for(batch)
     new.tap do |form|
--- a/app/models/batch_form.rb
+++ b/app/models/batch_form.rb
@@ -32,7 +32,6 @@ class BatchForm
   delegate :id, :new_record?, :persisted?, to: :batch
 
   validates_presence_of :date_produced
-  validates_numericality_of :samples_quantity, greater_than: 0, message: "value must be greater than 0", if: :creating_batch?
 
   def self.for(batch)
     new.tap do |form|


app/models/batch_form.rb
```diff
@@ -32,7 +32,6 @@ def self.human_attribute_name(*args)
   delegate :id, :new_record?, :persisted?, to: :batch
 
   validates_presence_of :date_produced
-  validates_numericality_of :samples_quantity, greater_than: 0, message: "value must be greater than 0", if: :creating_batch?
```
straight-shoota, line 35: We should probably make sure the value is non-negative by keeping the validation with `greater_than_or_equal_to: 0` (and appropriate change to the message).

leandroradusky: I could keep the validation but for what I show on the second screenshot this line of code will be never reached... Should we then modify `views/batches/_form.haml` removing the min in the input??
The "drawback" that I see on this is that pressing the down arrow when value is zero goes to -1 and before the limit was zero, which is a better behaviour.

straight-shoota: > I could keep the validation but for what I show on the second screenshot this line of code will be never reached... 

Probably not, but it's still important to validate the data we receive in the backend. Client request data can't really be trusted, ever.
There's probably not much practical relevance in this case, but it's a general good practice to validate incoming data.

> Should we then modify views/batches/_form.haml removing the min in the input??

No, the immediate feedback from the HTML validation is good.

We should have both. The roles are distributed:

* Server-side validation is necessary to ensure data integrity.
* Client-side validation is an enhancement on top of that to improve UX

straight-shoota:
You should rebase on `main` with #1638 to get rid of the CI failure.

straight-shoota (approved):

app/models/batch_form.rb
ysbaddaden: TLDR:

- Server side validation -> required for security & data integrity, we must NEVER trust any client provided data, requests can (and will) be forged.

- Client side validation -> improves the user experience with a reactive UI that gives quick feedback, and also prevents useless requests to the backend. When the form is submitted it should be a happy path, with only some backend specific validations that may fail.

Issue #1634, opened 2022-05-03
Error when printing/deleting multiple samples at once


diegoliberman:
### Steps
1. In the samples list page, select more than 1 sample.
2. Print.

### Expected a result
A PDF with 1 sample per page.

### Actual result
A PDF with the only 1 sample's label, from the ones that were selected, the one that appears last in the list.

![image](https://user-images.githubusercontent.com/25181052/166510072-61189449-7a5b-4f5d-ad80-de1d2cb76d1a.png)

### Deleting
Same experience

straight-shoota:
This error is pretty interesting: It seems to be caused by using GET method to submit the form. Changing to POST makes it work. Apparently, the form serialization doesn't account for multiple checkbox values in a GET request. It results in query data like `sample_ids[]=5` instead of `sample_ids[]=4&sample_ids[]=5`.

straight-shoota:
After fixing the method, the print action errors with #1645.

For the delete action, there is some copious client-side javascript code for showing a confirmation dialog. Apparently, that breaks the method configuration and results in sending a GET request instead.

mmuller:
In version fd72912, I am obtaining a PDF with 3 pages per sample.

ysbaddaden:
It's working as expected with #1650 and #1658 in development.

mmuller:
Thanks @ysbaddaden

ysbaddaden:
Note that I don't know why it didn't work, and I didn't fix anything specifically. It just happens to be working in development :shrug:

ysbaddaden:
The problem wasn't fixed, and only appeared in the production environments (staging, production): WickedPDF failed to find and inline the precompiled assets (CSS, logo), which resulted in the CSS not being applied.

I have a fix in 61878987

PR #1646, opened 2022-05-06
Fix multi select for bulk actions on `SamplesController#index`

straight-shoota:
Resolves https://github.com/instedd/cdx/issues/1634

### DIFF
--- a/app/views/batches/index.haml
+++ b/app/views/batches/index.haml
@@ -8,7 +8,7 @@
           %h1 There are no batches
           %p Every batch will hold multiple samples
     - else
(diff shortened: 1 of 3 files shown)


ysbaddaden (approved):
That looks much better.

Note: there is a batches#bulk_action too. Isn't it affected, too?

straight-shoota:
I haven't tested it, but it looks like bulk actions for batch migth be implemented differently. 

I tried setting up an integration test for this, but after battling with it for too long, I gave up for now. The implementation is quite confusing and makes it hard to test.

ysbaddaden (approved):

PR #1650, opened 2022-05-09
New resource: boxes

### LINKED ISSUE #1606: [Out] Allow to create boxes of samples

ysbaddaden:
closes #1606


app/controllers/samples_controller.rb
```diff
@@ -122,24 +124,19 @@ def bulk_print
...
-      pdf_file = MultipagePdfRenderer.combine(sample_strings, options)
-      pdf_filename = "cdx_samples_#{samples.size}_#{DateTime.now.strftime("%Y%m%d-%H%M")}.pdf"
-
-      send_data pdf_file, type: "application/pdf", filename: pdf_filename
-    rescue
+      send_data LabelsPdfRenderer.combine(pages),
+        type: "application/pdf",
+        filename: "cdx_samples_#{samples.size}_#{DateTime.now.strftime("%Y%m%d-%H%M")}.pdf"
+    rescue => ex
+      raise ex
```
ysbaddaden, line 138:
```suggestion
      raise ex
```

app/views/boxes/_barcode_card.haml
```diff
@@ -0,0 +1,19 @@
+.barcode-card.is-box
+  .barcode-content
+    .barcode(style="float:right")
+      = image_tag barcode_svg_data_url(@box.uuid)
```
ysbaddaden, line 4: It should use the PNG generator. The SVG looks much better on screen, but the PNG has been validated on the printer!

app/views/boxes/_form.haml
```diff
@@ -0,0 +1,54 @@
...
+              = label_tag "box_batch_numbers_#{index}", "Variant #{index}"
+            .col
+              = text_field_tag "box[batch_numbers][#{index}]", "", id: "box_batch_numbers_#{index}", disabled: true, placeholder: "Enter batch id"
+
+      #Challenge
+        .row.form-field
+          .col.pe-3
+            = label_tag "box_batch_numbers_0", "Virus"
+          .col
+            = text_field_tag "box[batch_numbers][id]", "", id: "box_batch_numbers_0", disabled: true, placeholder: "Enter batch id"
```
straight-shoota, line 39: Why `id` instead of `0`?

```suggestion
            = text_field_tag "box[batch_numbers][0]", "", id: "box_batch_numbers_0", disabled: true, placeholder: "Enter batch id"
```

app/views/boxes/index.haml
```diff
@@ -0,0 +1,44 @@
...
+              %th Box Id
+              %th Purpose
+              %th Samples
+          - t.tbody do
+            - @boxes.count_samples.each do |box|
+              %tr.laboratory-sample-row{data: {href: box_path(box)}}
+                %td
+                  =check_box_tag "box_ids[]", box.id, false, { id: "box_ids.#{box.id}" }
+                  %label.row{for: "box_ids.#{box.id}"}
+                %td= box.uuid
```
straight-shoota, line 33: We should probably use the short format for UUID.
```suggestion
                %td= short_uuid_with_title box.uuid
```

app/views/boxes/show.haml
```diff
@@ -0,0 +1,45 @@
...
+  .col
+    .row
+      .col.pe-3
+        %label Contents
+      .col.pe-9
+        - @samples.each do |sample|
+          .col.pe-7.batches-samples
+            .samples-row
+              .samples-item= sample.uuid
+              .sample-row-action= sample.batch.try(&:batch_number)
```
straight-shoota, line 40: This currently won't work with transferred samples because the `batch` relation gets removed on a transfer.
I'm thinking we should keep that reference to make things easier to reason about. And a sample always comes from its original batch, this can't really change. It shouldn't matter whether the sample is in the same location and institution as the batch it was created from.

ysbaddaden: Yes, this is by design: until the box is transferred, we have the batch relationship, and can display the batch number to the sender, but once it's transferred, we lose the relationship, and the receiving institution won't see the batch number anymore.

app/views/boxes/index.haml
diegoliberman: Full ID when there is room. No doubt.  Strong opinion here.  :)

app/assets/javascripts/components/cdx_select_section_toggler.js.jsx
```diff
@@ -0,0 +1,42 @@
...
+    }
+  },
+
+  toggleSection: function (id, display, disabled)  {
+    if (!id) return
+
+    var section = document.querySelector("#" + id);
+
+    Array.from(section.querySelectorAll("input, select, textarea"))
+      .forEach(function (element) { element.disabled = disabled; });
```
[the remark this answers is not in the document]
ysbaddaden: Oh, I didn't know about that one, it's super nice!

app/views/boxes/show.haml
diegoliberman: We shouldn't lose the data.  If I recall correctly, the relationship is removed, but the original batch ID is stored in another field of the Sample.
Whether the relationship needs to be removed or not, I don't know.  When we did that, it was thinking that it would be strange for the sample to belong to an institution and the batch to another.  But Johannes is making a good point, and I always prefer to try reflecting reality in the systems.  So, if it doesn't cause any colateral damage, I wouldn't disagree on keeping the relationship to the original batch.

app/controllers/boxes_controller.rb
```diff
@@ -0,0 +1,134 @@
+require "labels_pdf_renderer"
+
+class BoxesController < ApplicationController
+  before_action :load_box, except: %i[index new create bulk_destroy]
+
+  def index
+    @can_create = has_access?(@navigation_context.institution, CREATE_INSTITUTION_BOX)
+
+    @boxes = Box.where(institution: @navigation_context.institution)
```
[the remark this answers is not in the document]
ysbaddaden: I followed samples#index and batches#index which both exhibit the same behavior:

`check_access(where(institution)).within(context)`

Let's try to optimize this.

app/views/boxes/show.haml
straight-shoota: Even with keeping the relation we can choose to not show it when a sample is transferred. But it gives the easy option to show it when necessary.

app/views/boxes/index.haml
straight-shoota: IMO it shouldn't be about available space but about what's important. If the full UUID is important, we should make it happen that we can show it everywhere in full. If it's not important and showing the shortened version is good, then we should go with that everywhere. Not just because of consistency between different displays, but also because a shorter UUID is easier on the eye of the viewer than a long string of characters that have little informative value.

app/views/boxes/_form.haml
```diff
@@ -0,0 +1,54 @@
...
+          = f.label :purpose
+        .col
+          = react_component "CdxSelectSectionToggler", { name: "box[purpose]", items: Box.purposes.map { |v| { label: v, value: v } }, value: @box.purpose, submit: "#btn-save" }
+
+      #LOD
+        .row.form-field
+          .col.pe-3
+            = label_tag "box_batch_numbers_0", "Batch"
+          .col
+            = text_field_tag "box[batch_numbers][0]", "", id: "box_batch_numbers_0", disabled: true, placeholder: "Enter batch id"
```
straight-shoota, line 24: I suppose this should eventually become an autocomplete field?

straight-shoota:
What happens when the input data is incomplete? For example, not all the batch ID's have been filled out.
I presume the form submission should fail. But there doesn't seem to be any validation for that. The controller should probably check that the number of batches matches the expected number fo the given purpose. And maybe some other validations might be necessary based on the type of batch (or disallow duplicates?). But I think that latter part is still to be determined.

app/views/boxes/index.haml
diegoliberman: Having the long ID isn't important, but it could prove to be useful.  For instance, it could be easier to copy the full ID, and then search/paste it.  Plus other use cases that we might not imagine.  I see your point that it would be easier for the human eye to read the shortened version.  Both have a contra, but one of them is about making something not as nice, and the other one is about making something impossible.  That's why I prefer to avoid the impossible situation, when there are no other drawbacks (when we add not having enough room in the picture, then the balance turns in favor of the shortened version).  Of course this is subjective, but this is my view.

straight-shoota: >  For instance, it could be easier to copy the full ID, and then search/paste it. Plus other use cases that we might not imagine.

If we need the copy functionality or any other feature, we'll also need it in places where there's less available space. Thus we should implement a UI that works in these situations.

diegoliberman: Nobody asked for being able to copy the UUID.  So there is no need per se.  But nothing prevents users from doing so, right? Now, this doesn't mean that we have to be prepared for this situation in every page.  That would be crazy, as nobody asked for it, there's no real use case.  However, we know that users can manually copy anything. If we can easily improve their experience at least in 1 page, without having a negative impact in something else, why wouldn't we do it?  Is it just for being consistent? Is it really inconsistency when the diff is based on logic and not on randomness?

I feel I'm discussing too much about a not-important detail, and based on my imagination instead of actual user feedback.  

I'm willing to let it go, so go ahead with the shortened version.

app/views/boxes/_form.haml
ysbaddaden: Yes, eventually.

ysbaddaden:
@straight-shoota I think I addressed all your points, and added a validation of selected batches. I also extracted a BoxForm pseudo-model out of the controller. It's not an ActiveModel as other forms, but it's enough for now. I usually don't use them, but here it helped me clean up the controller, without pushing much form logic into the Box model.

app/models/box_form.rb
```diff
@@ -0,0 +1,101 @@
...
+        @box.build_samples(batch, exponents: [1, 4, 8], replicas: 3)
+      end
+
+    when "Challenge"
+      if batch = get_batch("0") { "please select a virus batch" }
+        @box.build_samples(batch, exponents: [1, 4, 8], replicas: 18)
+      end
+      @batches.each do |key, batch|
+        @box.build_samples(batch, exponents: [1, 4, 8], replicas: 3) unless key == "0"
+      end
```
straight-shoota, line 60: I'm a bit sceptical of just numbering the samples and giving the number zero a special meaning instead of using an explicit name (`virus`). But changing this would probably complicate the logic for loading batches.
Just a thought for consideration.

app/models/box_form.rb
```diff
@@ -0,0 +1,101 @@
...
+  private
+
+  def get_batch(key)
+    if batch = @batches[key]
+      batch
+    else
+      @batch_errors[key] ||= yield
+      nil
+    end
+  end
```
straight-shoota, line 87: This method is a bit convoluted and confusing. I'd rather separate validating the input data from fetching. This would make the code easier to reason about.
There is also some overlap of responsibility with `validate_batches`. I would suggest to move *all* validations into that method. And then we can just use `@batches[key]` instead of `get_batch`.
It means we need to validate before building samples, but I think this would indeed be the correct order.

app/models/box_form.rb
ysbaddaden: I have the same uneasiness. I think I still prefer a flat list of batch numbers, since it's more generic, and we can interpret it in any way internally (it could even be customized by deployment).

Yet, now that I have a BoxForm, I could have a `virus_batch_number` or `specific_batch_number`.

I'd love to see the list of 3 new purposes to decide further.

ysbaddaden: I named the keys instead of relying on number indexes. It doesn't change anything, because we already expected a `Hash<String, String>`, but it improves the overall readability!

app/models/box_form.rb
ysbaddaden: Yes, it adds another `if` to the controller, but the BoxForm model is nicer.

lib/labels_pdf_renderer.rb
```diff
@@ -0,0 +1,20 @@
+require_relative "./multipage_pdf_renderer"
+
+module LabelsPdfRenderer
```
ysbaddaden, line 3: Invoking wkhtmltopdf for each page is awfully slow. It takes several dozen seconds to render a PDF with many pages (up to 109 for a Challenge box).

I'm pretty sure we could render a single HTML and invoke wkhtmltopage once, and still get the same result.

ysbaddaden: See #1658

app/helpers/application_helper.rb
```diff
@@ -164,9 +164,31 @@ def show_institution?(action, resource)
...
+      options[:class] = [*options[:class], "field_with_errors"]
+
+      content_tag :div, **options do
+        yield
+        concat content_tag(:div, messages.first, class: "error-message")
+      end
+    else
+      content_tag :div, **options do
+        yield
+      end
```
straight-shoota, line 191: Is this integral to this PR? I'd rather leave this out and introduce improvements to the form builder separately.

app/models/box_form.rb
```diff
@@ -0,0 +1,95 @@
...
+    @box = box
+    @batch_numbers = batch_numbers
+  end
+
+  def batches=(relation)
+    records = relation.to_a
+
+    @batches = Hash[@batch_numbers.map do |key, batch_number|
+      [key, records.find { |b| b.batch_number == batch_number }]
+    end]
```
straight-shoota, line 33: Looks like this could be simplified using `transform_values`:
```suggestion
    @batches = @batch_numbers.transform_values do |batch_number|
      records.find { |b| b.batch_number == batch_number }
    end
```

app/helpers/application_helper.rb
ysbaddaden: Let's add a NOTE to rewrite as a proper form builder, and call it a day :bow: ?

straight-shoota: Okay fine... I just don't like that we keep adding little bits and pieces everywhere that we need to clean up again afterwards.

straight-shoota (approved):

app/helpers/application_helper.rb
ysbaddaden: I only noticed `#field_errors` that kinda replicates `#field_with_errors` but it was used only once, so I merged the two methods together, so they match Joni's design.

I added a TODO note that `#field_with_error` and `#validation_errors` should be extracted into a custom FormBuilder!

diegoliberman:
@ysbaddaden reviewing this PR code changes is a bit too much for me.  Is there any logic in particular you'd like me to take a look at?

straight-shoota:
https://github.com/instedd/cdx/blob/8a454d4b75550eb423ec2ab7a989927a799333b1/app/models/box_form.rb might be worth a look. It has most of the meat.

ysbaddaden:
@diegoliberman I didn't ask for your review, but you left comments on Johannes' review, so GitHub added you by itself and I can't remove you (it seems)?

diegoliberman:
My only observation (weak opinion) is that it's not intuitive to know that the exponent parameter refers to the concentration of the sample.  Perhaps I would name it `concentrations` or `con_exponents`. 

image

diegoliberman (approved):

PR #1655, opened 2022-05-14
Fix `TransferPackageController#create` to update sample status

straight-shoota:
`TransferPackageController#create` doesn't use `TransferPackage#add!` which takes care of modifying the sample status (like detaching it from the sender context).
So this didn't happen.

This patch fixes that by introducing a `before_create` hook on `TransferPackage` which works well with nested attributes.

### DIFF
--- a/app/models/transfer_package.rb
+++ b/app/models/transfer_package.rb
@@ -29,6 +29,19 @@ class TransferPackage < ApplicationRecord
           where(receiver_institution_id: institution.id)
         }
 
+  before_create do
+    sample_transfers.each do |sample_transfer|
+      sample = sample_transfer.sample
+      unless confirmed?
+        sample.detach_from_context
+      end
+      if includes_qc_info
+        sample.attach_qc_info
+      end
+      sample.save!
+    end
+  end
+
   def self.sending(sender, receiver, attributes = nil)
     create!(attributes) do |package|
       package.sender_institution = sender
@@ -42,8 +55,8 @@ class TransferPackage < ApplicationRecord
     )
 
     if sample.batch
-      if includes_qc_info && (qc_info = create_qc_info(sample))
-        sample.qc_info = qc_info
+      if includes_qc_info
+        sample.attach_qc_info
       end
 
       sample.old_batch_number = sample.batch.batch_number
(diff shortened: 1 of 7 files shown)


app/models/transfer_package.rb
```diff
@@ -29,6 +29,19 @@ class TransferPackage < ApplicationRecord
...
+    sample_transfers.each do |sample_transfer|
+      sample = sample_transfer.sample
+      unless confirmed?
+        sample.detach_from_context
+      end
+      if includes_qc_info
+        sample.attach_qc_info
+      end
+      sample.save!
+    end
```
ysbaddaden, line 42: Maybe use some oneliners to reduce indentation?

```ruby
sample = sample_transfer.sample
sample.detach_from_context unless confirmed?
sample.attach_qc_info if includes_qc_info
sample.save!
```

Feel free to overlook, it's pedantic style.

ysbaddaden (approved):
I'm just wondering at which moment we drop the `Sample#batch` relation, is it still accessible during the `#before_create` callback?

straight-shoota:
> I'm just wondering at which moment we drop the `Sample#batch` relation, is it still accessible during the `#before_create` callback?

Uhm, actually we don't. Detaching the batch happened in `TransferPackage#add!`. Apparently I didn't port this over to the callback workflow following the discussion in https://github.com/instedd/cdx/pull/1650#discussion_r871200522 which happened in parallel.

ysbaddaden:
We must drop the Sample -> Batch relationship during the transfer, because the Batch is kept in the original institution. During a transfer we:

1. copy the QC info from the Batch to the Sample (when asked to);
2. copy `batch.batch_number` as `sample.old_batch_number` (so we can invalidate all samples from a bad batch, even after a transfer);
3. nullify `institution_id`, `site_id` *and* `batch_id`.

I guess nullifying `site_id` and `batch_id` could be deferred to the confirmation step, which would allow to cancel the transfer, unless we keep a copy in TransferPackage?.

ysbaddaden:
Again: the implementation related to the comment in #1650 was _on purpose_: I can use `sample.old_batch_number || sample.batch.try(&:batch_number)` to always show the batch number, but we don't want that —the batch number tells what's in the sample. That's only part of blinding the sample (we can access all the info on the sample details) but that's an easy one.

ysbaddaden (approved):

PR #1659, opened 2022-05-16
Configure human attribute names in locale instead of form UI

straight-shoota:
This makes the human attribute names known to Rails so they are used uniformly everywhere.
Without this there were discrepancies between the form labels and other display locations such as error messages.

Before:
grafik
After:
grafik

### DIFF
--- a/app/views/batches/_form.haml
+++ b/app/views/batches/_form.haml
@@ -12,7 +12,7 @@
 
       .row.form-field
         .col.pe-3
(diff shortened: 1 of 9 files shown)


straight-shoota:
Note for QA: It should be sufficient to navigate to the forms of the affected models and verify that all labels are identical to before this change.

ysbaddaden (approved):

PR #1662, opened 2022-05-16
Custom form builder for attaching error messages to fields

straight-shoota:
I've implemented a custom form builder which takes care of showing error messages adjacent to the respective fields.
It keeps track which error messages are shown directly. The remaining ones are displayed as form-wide errors.
This makes it safe to apply this helper to forms without fearing about error messages getting lost.

The builder also comes with a neat helper for grouping all pieces of a form field (label, value, errors) and thus reducing a lot of boilerplate in the form templates.

Screenshot 2022-05-16 at 20-25-04 Connected Diagnostics Platform

So far I have only adapted a couple of forms to gather some feedback. The others should follow.

This does not touch errors in React components (using `input-required`). This should be done in a follow-up and can re-use the introduced DOM structure.

Ref  #1065, #1623
Builds on #1659

### DIFF
--- a/app/assets/stylesheets/_errors.scss
+++ b/app/assets/stylesheets/_errors.scss
@@ -12,19 +12,10 @@ form {
     margin-top: 5px;
     visibility: hidden;
     display: flex;
-    
+
     li {
       display: inline-flex;
       align-items: center;
     }
-  }
-
-  .field_with_errors {
-    color: #EB2122;
-    border-bottom-color: #EB2122 !important;
-  }
-    
-  .field_with_errors + .error {
-    visibility: visible;
   }
 }
--- a/app/assets/stylesheets/_forms.scss
+++ b/app/assets/stylesheets/_forms.scss
@@ -98,7 +98,7 @@ label {
     padding-bottom: 6px;
     height: 34px;
   }
-  
+
   .value{
     margin-left: 5px;
   }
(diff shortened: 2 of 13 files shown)


ysbaddaden (commented):
That looks _really_ nice!

Just a few remarks to make it perfect:

1. If we use the `#form_field` helper then all messages should have been printed, leaving only the `:base` error messages to render; maybe we don't need to render the remaining messages at the end? Unless we have messages for _other_ fields, but then it's be a bug with the form?

2. I miss the ability to choose between the basic and full message (i.e. to add the field name or not). I use it in the new box form.

ysbaddaden:
Let's merge as soon as unit tests are passing.

ysbaddaden (approved):

straight-shoota:
Failing specs is just #1665. Rebased.

> 1. If we use the `#form_field` helper then all messages should have been printed, leaving only the `:base` error messages to render; maybe we don't need to render the remaining messages at the end? Unless we have messages for _other_ fields, but then it's be a bug with the form?

It's only guaranteed for all attributes if we actually use `#form_field` everywhere. That might not be possible all the time when custom form styles are required (thinking about the institutions form, for example).
And theoretically the error keys do not necessarily correspond to attributes 🤷 Just being safe here.
However, I was thinking about adding a report for any error key other than `:base` left in `errors_to_show`. That way we can identify possible form errors.

> 2. I miss the ability to choose between the basic and full message (i.e. to add the field name or not). I use it in the new box form.

What is that for? I think we should generally use the same style (unless there are some special circumstances demaning otherwise).

PR #1663, opened 2022-05-16
Move items from main nav to settings page

### LINKED ISSUE #1550: Main navigation menu
One of @jkicillof's  latest designs (https://projects.invisionapp.com/d/main#/console/4103589/465275288/comments) shows a main navigation menu with very few elements.
This is actually independent from the discussions about the transfer workflow and scanner interaction, so it makes sense to discuss this separately.

grafik

I like it. I've been annoyed by the navigation menu being so big and hard to actually navigate.
After some time you're used to remember where the items are that you typically use.
But for new users it's overwhelming. And it's not great for experienced users either.

Joni's approach is to have a nested menu for samples and related things (batches, transfers).
It moves elements that are not relevant in the current context, out of sight. That makes them harder to reach, but you
typically don't need to reach them anyway from the sample page.

I had been thinking about another idea, though. The pages in the main navigation menu have very different usage characteristics.
For example, if you're working with samples, you'll spent most of the time on the samples page (and related).
In contrast, configuring a site or device is pretty rare. Many users may never have to go there after initial setup.

There is a clear difference in the frequency of these pages. Navigation should make frequently used things easily accessible,
and rarely used things can be further away.
I think the split goes right in the middle, where the first half are frequently used pages and the last half are rarely used.
Of course, this also depends on the kind of work you do. IIRC, UCSD has no patient interaction, so they probably don't use the patients page at all [^1].

In any case, I believe the last half of pages (sites, devices, logs, users, settings) are probably not frequently used.
They don't need to be easily reachable there when they're only used sporadically.
They are actually all kinds of settings. So I'd suggest to move them away from the main navigation menu and summarize all of them on the *Settings* page.
The link to that page could also be separated from the main entries, and moved to the right, next to the account menu.
It could also become a cog icon instead of a text label.
Then in the top right corner you have two menu items, one for account configuration and the other for context configuration. That fits together very nicely.

grafik

As a result, we get a clean, small main navigation menu with all the important links easily accessible and less useful clutter tucked away, but still reachable.

[^1]: A further improvement could be to add configuration for hiding pages when they're not used.
  If an institution like UCSD doesn't use patients data, they could easily hide that button from the nav. They don't need to go there anyways.

straight-shoota:
Resolves #1550

grafik


app/views/shared/_header_nav.haml
```diff
@@ -32,25 +32,11 @@
...
-          %li{:class => (["roles", "policies", "alerts", "incidents", "alert_messages"].include?(params[:controller]) || params[:action] == "settings") && "active"}
-            = link_to "Settings", settings_path
+
+  - if has_access_to_settings?
+    = link_to settings_path do
+      .icon-gray-light.medium{title: "Settings"}
+        = '<svg xmlns="http://www.w3.org/2000/svg" style="width: 0.9em;" aria-hidden="true" viewBox="0 0 24 24"><path fill="currentColor" d="M12 15.5A3.5 3.5 0 0 1 8.5 12A3.5 3.5 0 0 1 12 8.5a3.5 3.5 0 0 1 3.5 3.5a3.5 3.5 0 0 1-3.5 3.5m7.43-2.53c.04-.32.07-.64.07-.97c0-.33-.03-.66-.07-1l2.11-1.63c.19-.15.24-.42.12-.64l-2-3.46c-.12-.22-.39-.31-.61-.22l-2.49 1c-.52-.39-1.06-.73-1.69-.98l-.37-2.65A.506.506 0 0 0 14 2h-4c-.25 0-.46.18-.5.42l-.37 2.65c-.63.25-1.17.59-1.69.98l-2.49-1c-.22-.09-.49 0-.61.22l-2 3.46c-.13.22-.07.49.12.64L4.57 11c-.04.34-.07.67-.07 1c0 .33.03.65.07.97l-2.11 1.66c-.19.15-.25.42-.12.64l2 3.46c.12.22.39.3.61.22l2.49-1.01c.52.4 1.06.74 1.69.99l.37 2.65c.04.24.25.42.5.42h4c.25 0 .46-.18.5-.42l.37-2.65c.63-.26 1.17-.59 1.69-.99l2.49 1.01c.22.08.49 0 .61-.22l2-3.46c.12-.22.07-.49-.12-.64l-2.11-1.66Z"/></svg>'.html_safe
```
straight-shoota, line 39: I'm not sure how the process for adding icons to the font works. I just did this quick and dirty for now.

ysbaddaden: I documented the process quickly [app/assets/fonts/README.md](https://github.com/instedd/cdx/blob/main/app/assets/fonts/README.md).

I reproduced it last week to add the Boxes icon.

PR #1668, opened 2022-05-19
Add logging for unhandled errors in `FormFieldBuilder#field_errors`

straight-shoota:
The `form_errors` helper is supposed to show form-wide error messages (`:base` key). It also displays any error message that has not been shown before in an explicit error location.
This reports such unhandled error messages.

In test env, it raises an exception.

Elsewhere, they're logged to Rails logger and reported to sentry.

I already found and fixed two instances of missing error messages in the converted forms.


app/helpers/cdx_form_helper.rb
```diff
@@ -61,10 +67,26 @@ def errors_to_show
...
 
   def error_messages_to_show
-    errors_to_show.map do |attribute|
-      @object.errors.full_messages_for(attribute)
-    end.flatten.tap do
-      @errors_to_show = []
+    Hash[errors_to_show.map do |attribute|
+      messages = @object.errors.full_messages_for(attribute)
+      next if messages.empty?
+      [attribute, messages]
```
ysbaddaden, line 73: Maybe

```suggestion
      [attribute, messages] unless messages.empty?
```

app/helpers/cdx_form_helper.rb
```diff
@@ -61,10 +67,26 @@ def errors_to_show
...
+    end.compact]
+  end
+
+  def log_unhandled_errors
+    unhandled_errors = error_messages_to_show.dup
+    unhandled_errors.delete(:base)
```
ysbaddaden, line 79: The following would achieve the same in one action, but I'm not sure it improves readability?

```suggestion
    unhandled_errors = error_messages_to_show.reject { |k, _| k == :base }
```

ysbaddaden (approved):
Super nice :+1:

PR #1669, opened 2022-05-19
Export box inventory as ~XLS~ CSV

### LINKED ISSUE #1609: Allow to print box labels and inventory sheet
As a lab technician about to send samples, I want to print the box details so that I can perform the process of aliquoting without the need of having a computer next to me.

### Context
The samples/batches are stored in freezers at -80 degrees celsius.  The technicians have to take these out of the freezer, and don't have much time before they have to put them back.   When dealing with 108 samples, having to go back and forth with a computer for each sample, would take more time than what the samples can be out of the freezer.

### Solution
By printing the labels in advance, and sticking them into the empty tubes, the technician only needs to know what to put within each tube.  And they can know it by reading an inventory sheet with the exact information about each sample.  This way, for each sample they don't need to use the computer at all, they can just take a tube, find the sample ID in the inventory sheet, and aliquote.

### Game plan
Within the detail of a box, allow printing:
1. The box label.  QR Code, ID, Purpose.
2. The sample's label, printed by respecting this order: 
      1. Batch in the order that they were selected in the form.  If the batch A was the first one, then the first group of samples to be printed are those that belong to that batch and that are linked to the current box.
      2. Concentration, ascending.  The samples from the same batch, will be sorted by concentration.
      3. Replicate number, ascending.  The samples with the same batch and with the same concentration, will be sorted by replicate number.
3. Allow printing an inventory sheet with all samples within the box (to be done in #1632).  Something like this:
image

### Note
If it was easier, we could offer the user to download a CSV instead of printing the inventory sheet.  As a matter of fact, if it takes more or less the same time, perhaps it's better to allow downloading the CSV, because it might be useful for them to have that list in a spreadsheet.  They will need an extra step for printing, but it might be worth it.  Let's decide when the implementation time comes.

### Thoughts
In the inventory sheet, I think that we also need to include a column with the Batch ID, so that the technician knows what's the batch that the sample should be taken from.

### Consistency when printing
When printing labels from the Box print feature, or from the Sample's list, of from the Sample's detail, they should all follow this logic:

1. If the sample is within a box with a purpose != None, then the label data should include QR, ID, Purpose, Production Date, Media
2. Else the label data should include QR, ID, Inactivation Method, Isolate Name, Production Date.

### Mock up of the label that is printed for a sample that belongs to a Box

image

### Invision
https://projects.invisionapp.com/d/main#/console/4103589/466673802/preview

ysbaddaden:
Generates a flat XLS file, which is usually nicer to deal with than a CSV file.

Capture d'écran du 2022-05-19 18 57 00@2x

[cdx_inventory_1953cedc-5174-cdcc-1221-a6531618f1c6.xls](https://github.com/instedd/cdx/files/8732419/cdx_inventory_1953cedc-5174-cdcc-1221-a6531618f1c6.xls). My batches didn't have a virus lineage field, and the samples don't have a media field (until we go and change it manually for each), so they're empty. BTW: we probably don't need the media field for this inventory sheet, do we @diegoliberman?

closes #1609

### DIFF
--- a/app/assets/stylesheets/_barcode_card.scss
+++ b/app/assets/stylesheets/_barcode_card.scss
@@ -10,7 +10,7 @@
 
     background: $white;
     padding: 5px 7px;
-    margin: 15px;
+    margin: 15px 15px 7px;
 
     .title {
       font-size: 14px;
@@ -21,11 +21,6 @@
     .barcode {
       @include flexbox();
       justify-content: center;
-    }
-
-    .title {
-      font-weight: bold;
-      font-size: 16px;
     }
 
     .code {
@@ -60,13 +55,25 @@
   }
 
   .actions {
-    margin: 0 15px 15px;
+    margin: 0 15px 7px;
 
     .btn-icon {
+      margin: -3px 0 0;
+      font-size: 24px;
       background: none !important;
       color: inherit;
-      margin: -3px 0 0;
     }
+  }
+}
+
+.barcode-after {
+  margin: 7px 15px 0;
+
+  .btn-icon {
+    margin: -3px 0 0;
+    font-size: 24px;
+    background: none !important;
+    color: inherit;
   }
 }
 
--- a/app/controllers/boxes_controller.rb
+++ b/app/controllers/boxes_controller.rb
@@ -17,7 +17,19 @@ class BoxesController < ApplicationController
     return unless authorize_resource(@box, READ_BOX)
     @can_delete = has_access?(@box, DELETE_BOX)
 
-    @samples = @box.scrambled_samples.preload(:batch).to_a
+    @samples = @box.scrambled_samples.preload(:batch, :sample_identifiers)
+  end
+
+  def inventory
+    return unless authorize_resource(@box, READ_BOX)
+
+    @samples = @box.samples.preload(:batch, :sample_identifiers)
+
+    respond_to do |format|
+      format.csv do
+        @filename = "cdx_box_inventory_#{@box.uuid}.csv"
+      end
+    end
   end
 
   def print
@@ -28,7 +40,7 @@ class BoxesController < ApplicationController
       layout: "layouts/pdf.html",
       locals: {
         box: @box,
-        samples: @box.samples.preload(:sample_identifiers),
+        samples: @box.samples.preload(:batch, :sample_identifiers),
       },
       margin: { top: 0, bottom: 0, left: 0, right: 0 },
       page_width: "1in",
--- a/app/models/box_form.rb
+++ b/app/models/box_form.rb
@@ -88,8 +88,12 @@ class BoxForm
     when "Variants"
       @box.errors.add(:base, "You must select at least two batches") unless count >= 2
     when "Challenge"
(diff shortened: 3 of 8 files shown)


diegoliberman:
@ysbaddaden I'm not sure if this was the best way to go.   With my MAC, usually I can open Excel files without any problem, but this one not.  Numbers, the equivalent to Excel in MAC, considers it as a plain text file and attempts to import it as a CSV.

I also tried with Google Spreadsheet, and there it also fails to load.  

I think this option is improving the experience for Windows users but it's blocking the UX for MAC users or others without a Microsoft Excel license.

diegoliberman:
1. You can complete the virus lineage in at least 1 sample for making sure it appears in the inventory.
2. The media is an important piece of information, specially for the Awardee.  In fact, recently they ran a survey among all Awardees to find out what media they preferred.  I wouldn't remove this field from the inventory sheet.

ysbaddaden:
I looked a little further, and this file format is 'XML Spreadsheet (XMLSS)' from Excel 2002. I guess only Excel and LibreOffice ever implemented support for it :sob: 

I wanted to try the [caxlsx](https://rubygems.org/gems/axlsx) and [xaxlsx_rails](https://rubygems.org/gems/axlsx_rails) gems, that are very nice to use, but it would require to upgrade to Ruby 2.3.

Let's go the CSV route.

straight-shoota:
Would've been to good to be true 😆

ysbaddaden:
Reworked to generate a CSV file. For example:
[cdx_inventory_d55888ae-0e23-f580-7a3b-2a720dcc3252.csv](https://github.com/instedd/cdx/files/8738289/cdx_inventory_d55888ae-0e23-f580-7a3b-2a720dcc3252.csv)

diegoliberman:
@ysbaddaden let's change the `Virus type` title to `Virus lineage`.  Let's add the `Batch ID` column. 

I'm not convinced with the Concentration column.  In both Numbers and Spreadsheet, its values get converted in a way I can't figure out.  Perhaps I'm missing some kind of scientific notation?  For instance, the plain text of `10E-1` is being displayed as `10E+00` (????). @leandroradusky how should you show this data?

IMO, we should try to show it as similar as possible to the Sample ID form.  Perhaps we need to show it in 2 separated columns, as there are 2 possible values: the multiplier and the exponent.  And maybe a 3rd column showing the hole formula?

image
image

ysbaddaden:
@diegoliberman Your spreadsheet guessed the concentration is a number, and it's totally right: this is the scientific representation of a float: `1E+0` is the same as `10E-1` which is a mere `1.0` in regular float notation.

I'll do the small changes.

leandroradusky:
As julien said, these numbers are both the same, and I don't know if you can prevent excel from making this change (I've no idea of why it prefer to put the 10 as the n, this you can "move it" with a button in excel). I cannot ensure they won't complain, but for a technician that is working with concentrations and dilutions all day long should not be a problem nor an error inducing thing.

ysbaddaden:
I added `Batch ID`, and renamed `Virus Type` as `Virus Lineage` and took the liberty to rename `Inactivation Type` as `Inactivation Method`.

PR #1674, opened 2022-05-20
Add validation for sample in box context

straight-shoota:
This patch adds a simple validation that a sample has the same institution and site as the box it is contained in.

### DIFF
--- a/app/models/box.rb
+++ b/app/models/box.rb
@@ -34,6 +34,8 @@ class Box < ApplicationRecord
           concentration_number: 1,
           concentration_exponent: exponent,
           replicate: replicate,
+          institution: institution,
+          site: site,
         )
       end
     end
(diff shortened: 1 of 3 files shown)


app/models/box.rb
```diff
@@ -34,6 +34,8 @@ def build_samples(batch, concentration_exponents:, replicates:)
           concentration_number: 1,
           concentration_exponent: exponent,
           replicate: replicate,
+          institution: institution,
+          site: site,
```
ysbaddaden, line 38: It's not enough to rely on `Batch#build_sample` :facepalm:

straight-shoota: Yeah, it would only set the institution, but not the site. And technically, there is no inherent guarantee that the institutions of batch and box match, that's why I'm setting the institution explicitly to the box's here.

PR #1683, opened 2022-05-24
Adapt more forms to use `CdxFormHelper`

### LINKED ISSUE #1065: Improve the UX for missing mandatory fields in all forms

straight-shoota:
Applies the form field helper introduced in https://github.com/instedd/cdx/pull/1662 to more forms. This should finally resolve #1065.
<del>There are a couple of forms left which didn't seem to be trivial to adapt. We'll have to look into those more carefully.</del>
<ins>I changed it for all forms.</ins>

### DIFF
--- a/app/views/device_models/_form.haml
+++ b/app/views/device_models/_form.haml
@@ -7,6 +7,9 @@
         = f.text_field :name
 
       = f.fields_for :manifest do |manifest_form|
+        - # FIXME: This is just to surpress the error message showing up in the
+        - # form (via `form_errors`).
+        - f.field_errors :"manifest.definition"
         = manifest_form.form_field :definition do
           - if !@device_model.new_record? && @device_model.manifest.definition
             .file-uploaded
--- a/app/views/device_models/_form.haml
+++ b/app/views/device_models/_form.haml
@@ -1,90 +1,59 @@
-= validation_errors @device_model
-
-= form_for(@device_model, html: { multipart: true }) do |f|
+= cdx_form_for(@device_model, html: { multipart: true }) do |f|
   .row
     .col
-      .row
-        .col.pe-4
-          = f.label :institution
-        .col
-          .value= f.object.institution
+      = f.form_field :institution, value: f.object.institution
 
-      .row
-        .col.pe-4
-          = f.label :name
-        .col
-          = f.text_field :name
-
-
+      = f.form_field :name do
+        = f.text_field :name
 
       = f.fields_for :manifest do |manifest_form|
-        .row
-          .col.pe-4
-            = manifest_form.label :definition
-          .col
-            - if !@device_model.new_record? && @device_model.manifest.definition
-              .file-uploaded
-                %label.input.on
-                  = link_to @device_model.manifest.filename, manifest_device_model_path(@device_model)
-                  %a.clear-label{:href => "#"}
-                    = image_tag "ic-cross.png"
-            .upload-new-file
-              = manifest_form.file_field :definition, :class => "inputfile"
-              %label{:for => 'device_model_manifest_attributes_definition'}
-                %span
-                  .icon-upload.icon-blue
-                  Choose a file to upload
-                %a.clear-input{:href => "#"}
-                  = image_tag "ic-cross.png"
-
-      .row
-        .col.pe-4
-          = f.label :setup_instructions
-        .col
-          - if !@device_model.new_record? && @device_model.setup_instructions.present?
+        = manifest_form.form_field :definition do
+          - if !@device_model.new_record? && @device_model.manifest.definition
             .file-uploaded
               %label.input.on
-                = link_to "download", @device_model.setup_instructions.url
-                .clear-label
-                  = f.check_box :delete_setup_instructions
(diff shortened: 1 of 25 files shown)


straight-shoota:
There are some issues with nested scopes (`#fields_for`). #1689 fixes that.

app/helpers/cdx_form_helper.rb
```diff
@@ -78,30 +77,77 @@ def form_actions(&block)
...
+      key
+    end.to_sym
   end
 
   def error_messages_to_show
+    nested_prefix = @options[:nested_errors_namespace]
+    nested_prefix = "#{nested_prefix}." if nested_prefix
     Hash[errors_to_show.map do |attribute|
-      messages = @object.errors.full_messages_for(attribute)
+      next if nested_prefix && !attribute.to_s.starts_with?(nested_prefix)
```
ysbaddaden, line 123:
```suggestion
      next if nested_prefix && !attribute.to_s.start_with?(nested_prefix)
```

app/helpers/cdx_form_helper.rb
```diff
@@ -78,30 +77,77 @@ def form_actions(&block)
...
+
+      messages = object_errors.full_messages_for(attribute)
       [attribute, messages] unless messages.empty?
     end.compact]
   end
 
+  def clear_errors_to_show
+    if nested_prefix = @options[:nested_errors_namespace]
+      nested_prefix = "#{nested_prefix}."
+      errors_to_show.delete_if { |key| key.to_s.starts_with?(nested_prefix) }
```
ysbaddaden, line 133:
```suggestion
      errors_to_show.delete_if { |key| key.to_s.start_with?(nested_prefix) }
```

ysbaddaden (approved):

app/helpers/cdx_form_helper.rb
```diff
@@ -78,30 +77,77 @@ def form_actions(&block)
     end
   end
 
+  def fields_for(record_name, record_object = nil, fields_options = {}, &block)
+    fields_options[:nested_errors_namespace] = prefixed_error_key(record_name)
+
+    super(record_name, record_object = nil, objectify_options(fields_options)) do |form|
```
ysbaddaden, line 83:
```suggestion
    super(record_name, record_object, objectify_options(fields_options)) do |form|
```

ysbaddaden (commented):
Hum, I'm not so sure about `#fields_for` and nested keys. I need to check it up.

straight-shoota:
In light of https://github.com/instedd/cdx/pull/1689#pullrequestreview-989976210 I rebased this on main without #1689. It works with stock `fields_for` as well. Worst thing that happens is that errors from associated objects show up as form errors as well (besides directly on the field where they belong). But that's hardly a problem.
In the specs this happened only in a single form and there I added a simple fix to suppress the error for showing up on the parent form.

A proper fix should be implemented for `fields_for`, but we can already port the existing forms to use the form helper. It is clearly an improvement if error messages occasionally show up in two locations than all error messages being placed unrelated to the respective fields.

app/views/device_models/_form.haml
```diff
@@ -1,90 +1,62 @@
...
-                  = image_tag "ic-cross.png"
-
-      .row
-        .col.pe-4
-          = f.label :setup_instructions
-        .col
-          - if !@device_model.new_record? && @device_model.setup_instructions.present?
+        - # FIXME: This is just to surpress the error message showing up in the
+        - # form (via `form_errors`).
+        - f.field_errors :"manifest.definition"
```
ysbaddaden, line 12: What about making this explicit with a nice helper?

```ruby
def FormFieldBuilder.skip_errors(attr_name)
  @object.errors.delete(attr_name.to_sym)
  nil
end
```

ysbaddaden (approved):

app/views/device_models/_form.haml
straight-shoota: Yeah, that's a good idea. But such a skip method should probably delete all errors with that prefix path, not just exact matches. I wanted to keep that complexity out of this PR. Merging this is a great improvement and I wouldn't delay it. We can easily refactor it later (that's why I left the `FIXME`).

straight-shoota: No, we really need to handle nested keys already for use case like this. We should skip all errors starting with the key of the nested form. I'm only explicitly skipping `:"manifest.definition"` for now because that's the only error that shows up in CI. If the manifest object contains other errors (and there are a couple of validations on that model), they would still show up in the parent, but we'd want to suppress them and only show them attached to the nested form.

ysbaddaden:
LGTM!

Issue #1684, opened 2022-05-24
Add creation date in the list of boxes


diegoliberman:
Perhaps it isn't frequent to have multiple boxes created on the same date.  But there is room for adding more columns and at least for the purposes of testing it would be useful to easily identify boxes.  I would add a column with the creation date & time of the box.  Sort should be descending by creation date.  

![image](https://user-images.githubusercontent.com/25181052/170036676-954a6558-67d8-4abc-b42f-b7dde60c34a6.png)

diegoliberman:
@jkicillof please follow this issue and if we get to implement it, update the design.  Thanks!

ysbaddaden:
Good idea, and easy to add. Let's do it.

mmuller:
Working properly in version https://github.com/instedd/cdx/commit/1ace6019d5cccc50cd836e0dbc8c7f0c4462c110

PR #1685, opened 2022-05-24
Add creation date to list of boxes

ysbaddaden:
Capture d'écran du 2022-05-24 15 23 28@2x

cloes #1684

### DIFF
--- a/app/views/boxes/index.haml
+++ b/app/views/boxes/index.haml
@@ -15,10 +15,10 @@
               .icon-trash.btn-icon
           - t.columns do
             %col{:width => "5%"}
-            %col{:width => "30%"}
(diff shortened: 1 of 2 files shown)


straight-shoota (approved):

diegoliberman (changes requested):
Can we add the timestamp in the creation date?

ysbaddaden:
@diegoliberman I'm not sure whether we use a 12 hours or 24 hours clock in CDx? I chose a 12-hours clock since we already use the MM/DD/YYYY americanism for dates.

Capture d'écran du 2022-05-24 18 19 56@2x

diegoliberman (approved):

PR #1687, opened 2022-05-24
Rake: add Box permissions to predefined roles

### LINKED ISSUE #1680: Add the new Box permissions to the predefined institution and site roles
There is a rake task to update the predefined roles  for all institutions and sites (`policy:update_roles`), but it actually replaces the permission set, overriding any previous customization that could have been done by the users.

We should instead have a migration to only add the new Box permissions once, without touching the other ones.

ysbaddaden:
Applies the box related permissions to predefined roles, using the JSON definitions of the roles. Takes to only set the permissions for the roles that still exist (do not re-create them if they have been deleted), and to only inject the box related resources/actions.

This is an alternative to the `policy:update_roles` that actually resets and re-creates the predefined roles.

An alternative would be to make the predefined roles non editable, but still deletable. Users could remove them, we could update them (if they do exist), and that would be fine.

closes #1680

### DIFF
--- a/lib/tasks/policy.rake
+++ b/lib/tasks/policy.rake
@@ -32,6 +32,60 @@ namespace :policy do
     PoliciesMigrator.new.update_roles
   end
 
+  desc "Add Box permissions to predefined roles (one time migration)"
+  task :add_box_permissions => :environment do
+    search = lambda do |statements, resource|
+      statements.find { |s| s["resource"] == resource }
+    end
+
+    copy_box_policies = lambda do |predefined_role, actual_role|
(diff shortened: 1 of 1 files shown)


straight-shoota (commented):
I know barely anything about the policy system. So my review is not worth much, except that it looks fine on a very high level.

I'm wondering though if this could be generalized. What happens when we add another model and need default roles to get permissions for that? Would be nice if we could re-use this tool maybe?

ysbaddaden:
Yes, we could probably make it a little more abstract. Let's see that for the next time we had a resource!

PR #1689, opened 2022-05-24
Add `FormFieldBuilder#fields_for`

straight-shoota:
This adds handling for nested forms to the `FormFieldBuilder`. It needs to track the nested path of the current builder instance and access the error keys by full path.

### DIFF
--- a/app/helpers/cdx_form_helper.rb
+++ b/app/helpers/cdx_form_helper.rb
@@ -35,17 +35,16 @@ class FormFieldBuilder < ActionView::Helpers::FormBuilder
 
   # Renders all error messages for *attribute*.
   def field_errors(attr_name, full_message: true)
-    attr_name = attr_name.to_sym
-
+    error_key = prefixed_error_key(attr_name)
     messages =
       if full_message
-        @object.errors.full_messages_for(attr_name)
+        object_errors.full_messages_for(error_key)
       else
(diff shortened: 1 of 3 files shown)


ysbaddaden (changes requested):
AR will indeed put nested errors on the form's model, but for one-to-many associations it squashes them, and you can't rely on them. To be honest I don't even know why it does that :shrug:

For example:

```ruby
sample = Sample.new
sample.notes.build
sample.notes.build(description: "blabla")

sample.valid?
# => false

sample.errors.keys
# => [..., :"notes.users", :"notes.description", :notes]
```

Only the first Note has an invalid description, but the parent's error squashed the error and we'd print the error for both notes (:bug:).

We shall continue to always use the errors on each `@object` with unprefixed keys to get proper error messages for each record, yet we must clean the prefixed errors from the ancestors'...

**Maybe we don't need it to be automated?** It could be achieved manually with methods like `#skip_error(name)` and `#skip_nested_errors(name)`?

**Maybe we should just skip nested errors?**. It's only correct for one-to-one relations, deceiving/confusing for one-to-many, and duplicates what's already in the errors of each relation record...

Yet, I believe a `#skip_error(name)` method would be nice to have; sometimes we just want to invalidate the model, especially with `#validates_associated`, but printing an error message is pointless.

PR #1690, opened 2022-05-25
Fix login icons to be visible when field has value

### LINKED ISSUE #1670: Login icons should remain visible after input
image
image

straight-shoota:
Resolves #1670


ysbaddaden (approved):
Yeah, less JS!

Issue #840, opened 2016-02-22
All date input fields should be controlled by calendar tool


cisaacs01:
Do not allow users the possibility to mistake international date formats by using a calendar selection tool

nditada:
@anthonyoleary or @tripledub are you working on this already?

tripledub:
Yes we are, although @nditada did mention there is possibly a component for the date picker that Jonathan has devised previously. Either way, the three of us can work it out between us

cisaacs01:
Being handled by UK team

pmallol:
As for Version: 0.13-pre1, only the Date input in the Home page is controlled by a calendar tool. Other Date filters still maintain the 'Previous Month/Week/year' options.
It is expected all Date filters to be controlled by calendar tools.

straight-shoota:
This is still an issue with date fields on the sample form, for example. The fields on the sample filter form have widgets, though.

straight-shoota:
Also, fields like `date_produced` could probably be pre-filled with the current date for easy use.

PR #1691, opened 2022-05-26
Use `date_field` for `date_produced` form field

straight-shoota:
Use a dedicated date input field for date value. This adds a useful widget for selecting the date.

grafik

Note: The date format in the field is no longer defined by cdx. It's delegated to the browser's locale.

Ref: #840


ysbaddaden (approved):
Nice!

PR #1693, opened 2022-05-26
Refactor `date_produced` to always cast to `Time` type

straight-shoota:
Casting `date_produced` to `Time` simplifies the code because we don't have to deal with discerning types (`String` vs `Time` vs `Nil` all the time).
This would be the behaviour for a column-backed field in ActiveRecord. The form models are not tied to the database and have no type information, thus it doesn't work out of the box and we need a custom `date_produced=` setter.  In Rails 6 we can use the Attributes API.

### DIFF
--- a/app/models/batch_form.rb
+++ b/app/models/batch_form.rb
@@ -35,6 +35,20 @@ class BatchForm
   validates_presence_of :date_produced
   validates_numericality_of :samples_quantity, greater_than_or_equal_to: 0, message: "value must be greater or equal to 0", if: :creating_batch?
 
+  if Rails::VERSION::MAJOR >= 6
+    include ActiveModel::Attributes
+    attribute :date_produced, :date
+  else
+    def date_produced=(value)
+      value = value.presence
+      if value.is_a?(Time) || value.nil?
+        @date_produced = value
+      else
+        @date_produced = value.to_time rescue nil
+      end
+    end
+  end
+
   def self.for(batch)
     new.tap do |form|
       form.batch = batch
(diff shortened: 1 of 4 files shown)


app/models/concerns/date_produced.rb
```diff
@@ -1,21 +1,10 @@
...
-  end
-
   def date_produced_description
-    if date_produced.is_a?(Time)
-      return date_produced.strftime(I18n.t('date.input_format.pattern'))
-    end
+    return unless date_produced
 
-    date_produced
+    date_produced.strftime(I18n.t('date.input_format.pattern'))
```
ysbaddaden, line 7: It could even become `I18n.l(date_produced.to_date, format: :traditional)`.

Then, I don't think we'd need the `input_format` setting anymore?

app/models/batch_form.rb
```diff
@@ -28,6 +28,20 @@ def self.human_attribute_name(*args)
...
+  else
+    def date_produced=(value)
+      value = value.presence
+      if value.is_a?(Time) || value.nil?
+        @date_produced = value
+      else
+        @date_produced = value.to_time rescue nil
+      end
+    end
+  end
```
ysbaddaden, line 43: What about moving the setter to the `DateProduced` concern instead of the duplication?

ysbaddaden (approved):

app/models/batch_form.rb
straight-shoota: I briefly considered that, but wasn't sure The setter is only necessary for a lacking attributes API. It wouldn't be necessary when the field is backed by a database column with datetime type. So one could consider it as being specific to the implementing type, not the concern.
However, the form types themselves don't even include `DateProduced`.
And I think by now we can probably get rid of that concern entirely. The only function that it has left is about display logic which is trivial and really doesn't belong into the model.
I'm just waiting for the next PR to resolve this =)

app/models/concerns/date_produced.rb
straight-shoota: Yes, we can get rid of this.

PR #1695, opened 2022-05-27
Allow to disable sign ups via environment variable

matiasgarciaisaia:


straight-shoota (approved):

matiasgarciaisaia:
There's an exception happening after logging in with sign ups disabled.

```
undefined local variable or method `edit_user_registration_path' for #<#<Class:0x007f92e7abf6b8>:0x007f92e0046f80>
```

ysbaddaden:
@matiasgarciaisaia I didn't notice your comment. I pushed a fix and did some refactor to use `Settings.public_registration` instead, that can be controlled with the `PUBLIC_REGISTRATION` environment variable in the same fashion than `SINGLE_TENANT` (i.e. set it to `true` or `false`).

Please note that this is a breaking change, as you must set the env var to `true` to enable public registration!

config/settings.yml
```diff
@@ -9,6 +9,7 @@ google_client_id:
 google_client_secret:
 google_maps_api_key:
 single_tenant: <%= ENV['SINGLE_TENANT'] %>
+public_registration: <%= ENV['PUBLIC_REGISTRATION'] %>
```
straight-shoota, line 12: Can we avoid the breaking change and accept an unset variable as `true`. This makes the registration feature effectively opt-out, which IMO should be okay.

ysbaddaden:
@matiasgarciaisaia cherry-picked to the `release/0.20` branch.

Issue #1611, opened 2022-04-22
Confirm the receipt of a transfer


diegoliberman:
As an Awardee, I want to confirm the receipt of a transfer, so that I can provide feedback to the sender.
As a delivery responsible, I want the Awardee to confirm the receipt, so that I can keep track of the location of the samples that I used to own.

When confirmed, the samples will appear in the recipient's list of samples.  And the box will appear in the recipient's list of boxes.

Within the transfer log detail, the recipient will see a button to confirm the receipt.  One button for each box.  
This button will show a modal.  The recipient will confirm by entering the last 4 characters of the box ID.  

### How we had implemented it for the samples:
<img width="323" alt="image" src="https://user-images.githubusercontent.com/25181052/164751033-38da4842-54c6-4091-bfd9-c35bb1abfc53.png">

Issue #1610, opened 2022-04-22
Allow to transfer boxes


diegoliberman:
As a delivery responsible, I want to easily transfer a Box of samples, so that I can send the materials to the requester institution.

In the same way we are allowing to transfer samples, we can also allow to select a box.  Samples won't be transferred directly, only through a box.

The transfer action moves the samples records & box records to the recipient institution.

### Mock up
![image](https://user-images.githubusercontent.com/25181052/167026572-d308c0cd-a9ec-46d3-976a-da8a208b1203.png)

### Invision
https://projects.invisionapp.com/d/main#/console/4103589/465275290/preview
And see subsequent screens

straight-shoota:
For this we'll need a table to express a many-to-many relationship between `TransferPackage` and `Box`. A `TransferPackage` can contain multiple boxes and a `Box` could be transferred multiple times - at least theoretically. I think it makes sense to use this model just to be future proof. It isn't really more effort than a one-to-many relationship.

Lately, `SampleTransfer` has served this purpose for transfers of individual samples. It doesn't really serve any other purpose any more. It was originally created to keep track of transfers, but that role has been taken over by `TransferPackage`. We can probably drop `SampleTransfer` entirely after the transition to transferring boxes.

straight-shoota:
@diegoliberman How are we going to handle QC samples with box transfers? Do we need to check for QC samples when transferring a box? Or should it already be impossible to add QC samples to a box?
I think the latter might make more sense. There should be no reason to have QC samples in a box, when the purpose of boxes is being transferred, right?

ysbaddaden:
We can't add arbitrary samples to a Box, yet, so it's not a problem _now_.

But when we implement the 'Other' purpose, we'll have to deal with it, yes.

straight-shoota:
Yes, we'll have to implement that *now* in order to migrate transfer packages to transfer boxes instead of samples.

diegoliberman:
We should restrict QC samples from being added to a box.  Whenever it's convenient implementation wise.  Just in case, I added this consideration in #1619.

PR #1698, opened 2022-05-31
Blind/unblind box samples

### LINKED ISSUE #1614: Blind/Unblind samples
As UCSD, I want to blind the samples so that the recipient is freed from bias when running the tests on the samples.
As UCSD, I want to unblind the samples onces the Awardee informs that the results are ready.

- Blinding is done if the user enabled the check box when transferring boxes
- Blinding means that samples that were created as part of a Box, should have some attributes hidden for the recipient of those samples.

1. For LOD, we should blind the `concentration` level and the `replicate`.
2. For Variants, we should blind the `Virus lineage` and `Batch ID`.
3. For Challenge, we should blind the `concentration`, `replicate`, the `virus lineage` and `Batch ID`.

### Blinding
Hidden since the moment it's transferred, everywhere for the recipient of the transfer:
1. In the transfer logs
4. In the sample's list
5. In the sample's detail

The data isn't hidden to the Sender, just to the recipient.  

### Unblinding
The unblinding will be done manually but by the Sender of the VQA Box.  We could add a button in the transfer log that correspond to VQA boxes.  This button will unblind all the samples that had been transferred within that transfer process.

### Printing labels & downloading the inventory CSV
As the labels are printed in a certain order, it might be a problem for blinded data.  For instance, for challenge, we will be always printing the virus samples first.  Same goes for the downloaded CSV.  So, the recipient of a box could know what the virus samples are, just by printing or downloading.

If the data is blinded, we should hide the action of printing.  I think it's the easier approach.  Anyhow, there's no use case for the recipient needing to print the labels.  I think that this logic needs a comment in the code mentioning why we are hiding the action.   PS: An alternative would be to allow printing but changing the print order by UUID.

If the data is blinded, we should allow to download the csv, but the rows should be ordered by UUID.

### Invision
Transfer with blind check box: https://projects.invisionapp.com/d/main#/console/4103589/465275294/preview
Unblind from the transfer log: https://projects.invisionapp.com/d/main#/console/4103589/466664448/preview
Sample detail with blinded data: https://projects.invisionapp.com/d/main#/console/4103589/466778251/preview

### Acceptance criteria
1. When transferring a box, I can choose to blind the box with a check box.
2. The recipient of a blinded box won't see the blinded data anywhere (samples view, sample detail, inventory sheet)
3. The recipient can't print the box & sample labels.
4. The recipient can download the inventory sheet sorted by UUID
5. The sender can unblind a box through the transfer log.

### LINKED ISSUE #1628: Allow to print inventory sheet to the recipient of a transfer
As an Awardee, I want to print an inventory sheet with the samples that I received within a transfer, so that I can make physical controls.

1. If the transfer is a VQA box, the inventory sheet should respect the sample's blinding, and the blinded columns should not be included in the print page.  Alternatively we could offer to download a CSV instead.  The order of the samples in the printed list, should be the same than described in #1609.
2. If the transfer is individual samples, there's no need to blind data.

image

ysbaddaden:
Adds support for blinding samples

- [x] add blinded column to boxes;
- [x] redact samples attributes of blinded boxes (depending on box purpose);
- [x] text copy + design
- [x] checkbox to blind (or not) the box samples when creating a transfer package (~~depends on #1610~~);
- [x] button to unblind received boxes from the transfer package list (sender only, if blinded, depends on #1611).

closes #1614 
closes #1628


db/migrate/20220531133731_add_blinded_to_boxes.rb
```diff
@@ -0,0 +1,11 @@
+class AddBlindedToBoxes < ActiveRecord::Migration
+  def up
+    change_table :boxes do |t|
+      t.boolean :blinded, default: false, null: false
+    end
+  end
+
+  def down
+    remove_column :boxes, :blinded
+  end
```
straight-shoota, line 10: Totally marginal, but couldn't this be a single `change` method?
```suggestion
  def up
    add_column :boxes, :blinded, :boolean, default: false, null: false
  end
```

db/schema.rb
```diff
@@ -168,17 +168,18 @@
...
   create_table "boxes", force: :cascade do |t|
-    t.string   "uuid",           limit: 36,    null: false
+    t.string   "uuid",           limit: 36,                       null: false
     t.integer  "institution_id", limit: 4
     t.integer  "site_id",        limit: 4
     t.string   "site_prefix",    limit: 255
-    t.text     "core_fields",    limit: 65535
-    t.text     "custom_fields",  limit: 65535
+    t.text     "core_fields",    limit: 16777215
+    t.text     "custom_fields",  limit: 16777215
```
straight-shoota, line 176: These changes seem to be unrelated to the migration?

app/models/box.rb
```diff
@@ -81,4 +81,21 @@ def detach_from_context
       sample.save!
     end
   end
+
+  def blind_attributes
```
straight-shoota, line 85: Maybe call this `blinded_attributes`?

app/views/samples/_form.haml
```diff
@@ -37,7 +37,10 @@
           = text_field_tag :specimen_role, Sample.specimen_role_description(f.object.specimen_role), { :class => 'input-x-large', readonly: true }
 
       = f.form_field :virus_lineage do
-        = f.text_field :virus_lineage, class: 'input-x-large', readonly: !@can_update
+        - if blinded_attribute?(@sample_form, :virus_lineage)
+          .value= blinded_value
+        - else
```
straight-shoota, line 42: This pattern is used three times already. Perhaps this would be a good use case for a blind method that yields when the attribute is not blinded?
```rb
  def blinded_attribute(sample, attr_name, value: nil)
    if blinded_attribute?(sample, attr_name)
      value || blinded_value
    else
      value = sample.__send__(attr_name)
      if block_given?
        yield value
      else
        value
      end
    end
  end
```
It's just an idea and I haven't completely thought it out. Would also need revisiting for the presenter option.

spec/support/blueprints.rb
```diff
@@ -208,6 +208,33 @@ def first_or_make_site_unless_manufacturer(institution)
...
+
+Box.blueprint(:Challenge) do
+  institution { object.site.try(&:institution) || Institution.make }
+  purpose { "Challenge" }
+  samples { [
+    Sample.make(institution: object.institution, date_produced: Date.today),
+    Sample.make(institution: object.institution, date_produced: Date.today),
+  ] }
+end
+
```
straight-shoota, line 237: I'm wondering if it makes sense to define purpose specific blueprints when they don't really implement the purpose box? The samples are just stubbed.
You would get pretty much the same result using the generic blueprint with a purpose attribute (`Box.make(purpose: "Challenge)`). Only diff would be the number of samples, but I think we can simply increase that to two in the default blueprint.

straight-shoota (commented):
There should be some specs for blinding happening in `samples#show`.

Blinding currently doesn't discriminate sender vs. receiver. IIUC, blinding should only affect the receiver.

db/migrate/20220531133731_add_blinded_to_boxes.rb
ysbaddaden: You're right. I'm used to write custom up/down methods...

ysbaddaden:
@straight-shoota I addressed most of your comments. I'll rebase tomorrow morning and review the blueprints.

app/presenters/sample_presenter.rb
```diff
@@ -0,0 +1,56 @@
...
+    :virus_lineage,
+  ].each do |attr_name|
+    define_method attr_name do
+      if blinded_attribute?(attr_name)
+        blinded_value
+      else
+        @sample.__send__(attr_name)
+      end
+    end
+  end
```
straight-shoota, line 27: I'd consider this list of affected attributes a smell. It's a duplicate that needs to be kept in sync with the attributes returned by `Box#blind_attributes`.
Two possible solutions:

* Add this wrapper for *all* delegated attributes. This means essentially replacing `delegate_missing_to` with a custom `method_missing`. This is practically already happening here for the affected attributes.
* Refactor `Box#blind_attributes` to a class method that returns all blindable attributes when no purpose is given and use this list here.

PR #1701, opened 2022-06-03
Add `transfer_packages#confirm`

straight-shoota:
Implements the core feature of #1611.
This is just a basic confirmation without any validation step. This will be provided in a follow-up.

### DIFF
--- a/app/controllers/transfer_packages_controller.rb
+++ b/app/controllers/transfer_packages_controller.rb
@@ -38,8 +38,10 @@ class TransferPackagesController < ApplicationController
       .find(params[:id])
 
     @transfer_package = TransferPackagePresenter.new(transfer_package, @navigation_context)
+
     @view_helper = view_helper({ save_back_path: true })
     @can_update = false
+    @can_confirm = !transfer_package.confirmed? && @transfer_package.receiver? && authorize_resource(confirmation_resource(transfer_package), UPDATE_BOX)
   end
 
   def new
@@ -59,7 +61,7 @@ class TransferPackagesController < ApplicationController
     end
 
     if @transfer_package.save
-      redirect_to boxes_path, notice: "Boxes were succesfully sent"
+      redirect_to transfer_packages_path, notice: "Transfer package was successfully sent"
     else
       @view_helper = view_helper
       @can_update = true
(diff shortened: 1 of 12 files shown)


app/controllers/transfer_packages_controller.rb
```diff
@@ -38,8 +38,28 @@ def show
...
+    @can_confirm = !transfer_package.confirmed? && @transfer_package.receiver? && authorize_resource(confirmation_resource(transfer_package), UPDATE_BOX)
+  end
+
+  def confirm
+    transfer_package = TransferPackage
+      .with_receiver(@navigation_context.institution)
+      .find(params[:id])
+    @transfer_package = TransferPackagePresenter.new(transfer_package, @navigation_context)
+
+    return unless authorize_resource(confirmation_resource(transfer_package), UPDATE_BOX)
```
ysbaddaden, line 53: The `#confirmation_resource` method takes a Box, and will pass `box.id` to the policy framework, but you're passing a TransferPackage instead? Same above for `@can_confirm`.

I think you mean to use `box = transfer_package.boxes.take` instead?

app/controllers/transfer_packages_controller.rb
```diff
@@ -38,8 +38,28 @@ def show
...
+
+    @can_confirm = !transfer_package.confirmed?
+
+    if transfer_package.confirm!
```
ysbaddaden, line 57: The method raises on error, so I don't think we need a `if`?

ysbaddaden (changes requested):
Yeah, we're almost there!

I just have one issue with the authorization, then that's about it.

app/controllers/transfer_packages_controller.rb
straight-shoota: Good catch. I believe the policy framework doesn't actually validate the actual resource ID, otherwise this would've likely failed at some point.

ysbaddaden: I believe it tries to search a permission for `box:id` with a fallback to a broader `institution:id` permission, so it works because I don't think there are actual such fine grained permissions (but there could be).

ysbaddaden (approved):

PR #1702, opened 2022-06-03
Drop support for Rails 4.2

ysbaddaden:
Fully migrates the application to Rails 5.0, removing support for Rails 4.2, disabling the next variant (until we start migrating to Rails 5.1), and enabling all the Rails 5.0 settings.


ysbaddaden:
QA: nothing much to review. The only real change with the currently deployed Rails 5.0 is that belongs_to associations are now required by default, which may cause some errors to pop up when validating forms.

Testing for #1065 should be enough to validate this change.

PR #1707, opened 2022-06-09
Box: add other purpose

### LINKED ISSUE #1619: Allow to add in a box: individual samples
When the box's purpose is `Other`, we should allow to add existing samples to the box.

### Story
As UCSD, I want to create a custom box for transferring some specific samples, so that I can keep track of what happened with these samples.

### Acceptance criteria
1. I select the `Other` purpose.
2. I can see a field for entering a sample ID, I'll call this field `Find ID`.
3. If 1 or more non-QC samples match with the **partial** ID of the previous step, I can confirm which one I want to add and it gets added to a list of samples being added to the box.  After confirming, the focus moves back to the `Find ID`.
4. I can continue adding as many samples as I want.
5. If a non-QC sample matches with the **complete** ID of step 2, the sample gets added automatically and the focus is placed again in the `Find ID`.
6. I can save the box.

### Invision
https://projects.invisionapp.com/d/main#/console/4103589/466673800/preview

ysbaddaden:
We can now select the "Other" purpose when creating a Box. It will present an interface to select samples one by one by their UUID. You can type an UUID and it will be autocompleted, or you can paste/scan a QR code.

Notes/Questions:

1. ~~Which sample attributes should be blinded for this purpose?~~ all of them

2. QC samples aren't autocompleted and can't be selected (we can't transfer a box with QC samples). ~~Alternative would be to warn about the sample being QC, and eventually prevent transferring a box with QC samples.~~

Improvement:

- [x] Auto-select the sample when we entered the UUID completely (and got one result) to improve the UX with QR scanners.

closes #1619

### DIFF
--- a/app/assets/javascripts/components/cdx_select_autocomplete.js.jsx
+++ b/app/assets/javascripts/components/cdx_select_autocomplete.js.jsx
@@ -25,6 +25,7 @@ var CdxSelectAutocomplete = React.createClass({
     if (!query) return callback(null, []);
 
     var autoselect = this.props.autoselect;
+    var prepareOptions = this.props.prepareOptions;
     var url = this.props.url;
     url += (url.includes("?") ? "&query=" : "?query=") + encodeURIComponent(query);
 
@@ -33,6 +34,9 @@ var CdxSelectAutocomplete = React.createClass({
       url: url,
 
       success: function (options) {
+        if (prepareOptions) {
+          options = prepareOptions.call(null, options);
+        }
         callback(null, {
           options: options,
           complete: options.size < 10,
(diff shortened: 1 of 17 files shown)


straight-shoota (commented):
LGTM 👍 

There is duplication of the same function in `BoxesController#samples_data` and `autocomplete.json.jbuilder`. This could be consolidated.

straight-shoota (approved):
Wrong button, I meant to approve =)

ysbaddaden:
I tried but failed to avoid the duplication between the `samples/autocomplete` template and the `#samples_data` helper. The template dumps a string (obviously), but the `#react_component` expects ruby objects to dump to JSON along other react props.

I pushed one last commit, that:
- fixes react warnings (bind, key);
- harmonizes samples/autocomplete and samples_data to return uuid & batch_number then have the SamplesSelector component prepare the options for CdxSelect (value, label);

straight-shoota:
How about using the helper method for `samples#autocomplete` instead of the other way around?

PR #1719, opened 2022-06-14
Fix: use direct SQL in migration instead of removed AR association

ysbaddaden:
That should fix the deployment issue in v0.21.0 by avoiding to rely on an AR association (that was removed) and using a direct SQL request.


matiasgarciaisaia (approved):
:shipit:

PR #1727, opened 2022-06-21
Fix: can't register in development (aka can't send registration email)

ysbaddaden:
This was caused by a workaround that isn't needed anymore.


diegoliberman:
@omelao can you please confirm if it works locally and if it does, move the card to the Verified column?

PR #1737, opened 2022-06-29
Replace icomoon font with individual SVG files

ysbaddaden:
**There are numerous issues with the icomoon font**

1. it's tedious to add an icon, when it should just be adding a SVG file to the assets directory;
2. every new developer stumbles on this issue (:face_with_head_bandage:);
3. we can't have 2 pull requests in parallel each adding a new icon (:rage:).

**This patch is an attempt to move away from it**

I extracted the individual SVG icons from the `icomoon.svg` font, and changed the SCSS file to use the individual files instead of icomoon. I'm using CSS filters to colorize the icons, with a SASS mixin to autogenerate the `filter` property for any CSS color. Thank you SASS for all the nice color functions!

The change is supposedly transparent, since it's restricted to CSS, but of course there are some issues:

- [x] `.icon-text-color` and `.btn-icon` rely on `color: inherit`. It could be fixed by using a color class name, and/or with a helper function to embed the icon. For example:

```scss
.btn-icon {
  color: inherit;
  fill: currentcolor;
}
```
```ruby
# app/helpers/icon_helper.rb
module IconHelper
  def icon(name, **attributes)
    attributes[:class] = "icon-svg #{attributes[:class]}"

    content_tag :span, **attributes do
      raw File.read(Rails.root.join("app/assets/images/icons", "#{name}.svg"))
    end
  end
end
```
```html.erb
<%= icon "visibility_off" %>
<%= icon "visibility_off", class: "icon-blue" %>
```

- [x] icons on the settings page aren't centered horizontally: this is caused by `.centered` being defined twice and not working correctly, causing  one centers text (thus horizontally), when the other one uses flexbox... which will align vertically and takes precedence (urgh).


straight-shoota:
I don't follow the reason for `icon-color` helper. Why can't we configure the color directly in the SVG via CSS?

ysbaddaden:
If you embed the SVG in the page you can style them with CSS **but** here I'm referencing the SVG as external images, hence external documents, which can't be styled with CSS directly, hence the `icon-color` mixin.

straight-shoota:
Ah, okay. Makes sense then. I didn't see that you're just using `background-image`.

ysbaddaden:
I added an `icon` helper, fixed usages of `color:inherit` to use it, and wen the extra mile to embed the SVG in all templates... but the React components in `app/assets/javascripts/components`, where I don't think we can use view helpers AFAIK?

app/assets/stylesheets/_barcode_card.scss
```diff
@@ -60,8 +60,8 @@
     .btn-icon {
       margin: -3px 0 0;
       font-size: 24px;
-      background: none !important;
       color: inherit;
+      svg { fill: currentcolor; }
```
straight-shoota, line 64: I'm wondering if this shouldn't be the default for *all* SVG icons? I think you would usually want the icons to be in the text color. That was previously the case with the icon font, so it should not be a change. SVG opens more possibilities for coloured icons, but `currentColor` should be the default.

ysbaddaden: See the current [`_sprites.scss`](https://github.com/instedd/cdx/blob/main/app/assets/stylesheets/_sprites.scss#L27), it sets the icon color to black by default.

straight-shoota: Sigh, it's stupid, but you're right. We should consider changing that some time, though.

PR #1739, opened 2022-06-30
1598 show date last modified date in samples

omelao:
I don't know if we need to consider some conflict between Modified From/To and Modified Dropdown filters. I don't think it would be a problem. So this is it.

### DIFF
--- a/app/assets/javascripts/components.js
+++ b/app/assets/javascripts/components.js
@@ -17,6 +17,19 @@ function cdx_select_on_change(name, callback) {
     callback($(this).val());
   });
 }
+
+/*
+This is kind of a exception solution. Because input date's behavior are different.
+Consider this points:
+  - input date's placeholders doesn't work
+  - even if at some point in the future it works, you cannot know what's the OS date format
+  - CSS selectors like :empty, :valid, [value=''], :placeholder-shown doesn't work
+  - date format's depends of the browser and OS
+  - Google Chrome has specific selector to color the content of this kind of fields, but other browser doesn't
+
+  Finally you can test if something changes on the future on this fiddle:
+    https://jsfiddle.net/omelao/hsk8vyax/15/
+*/
 $(document).on("change", "input[type=date]", function(){
   $(this).val() != '' ? $(this).addClass("filled") : $(this).removeClass("filled");
 });
--- a/app/assets/javascripts/components.js
+++ b/app/assets/javascripts/components.js
@@ -17,3 +17,7 @@ function cdx_select_on_change(name, callback) {
     callback($(this).val());
   });
 }
+
+$(document).on("change", "input[type=date]", function(){
+  $(this).val() != '' ? $(this).addClass("filled") : $(this).removeClass("filled");
+});
--- a/app/assets/javascripts/components.js
+++ b/app/assets/javascripts/components.js
@@ -17,7 +17,9 @@ function cdx_select_on_change(name, callback) {
     callback($(this).val());
   });
 }
-
 $(document).on("change", "input[type=date]", function(){
   $(this).val() != '' ? $(this).addClass("filled") : $(this).removeClass("filled");
+});
+$(function(){
+  $("input[type=date]").change();
 });
--- a/app/assets/stylesheets/_global.scss
+++ b/app/assets/stylesheets/_global.scss
@@ -377,3 +377,9 @@ a[title="Back"] {
 .faded {
   color: $gray3;
 }
+input[type=date].filled::-webkit-datetime-edit { 
+  color: $black;
+}
+input[type=date]::-webkit-datetime-edit {
+    color: $gray3;
+}
--- a/app/assets/stylesheets/_global.scss
+++ b/app/assets/stylesheets/_global.scss
@@ -377,9 +377,9 @@ a[title="Back"] {
 .faded {
   color: $gray3;
 }
-input[type=date].filled::-webkit-datetime-edit { 
+input[type=date].filled { 
   color: $black;
 }
-input[type=date]::-webkit-datetime-edit {
+input[type=date] {
     color: $gray3;
 }
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -15,6 +15,7 @@ class SamplesController < ApplicationController
     @samples = @samples.where("specimen_role = ?", params[:specimen_role]) unless params[:specimen_role].blank?
     @samples = @samples.where("updated_at >= ?", params[:updated_at_from].to_time) unless params[:updated_at_from].blank?
     @samples = @samples.where("updated_at <= ?", params[:updated_at_to].to_time) unless params[:updated_at_to].blank?
+    @samples = @samples.where("updated_at >= ?", params[:modified].to_time) unless params[:modified].blank?
     @samples = @samples.order( params[:sort] + ' DESC') unless params[:sort].blank?
 
     @institutions = Institution
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -16,7 +16,7 @@ class SamplesController < ApplicationController
     @samples = @samples.where("updated_at >= ?", params[:updated_at_from].to_time) unless params[:updated_at_from].blank?
     @samples = @samples.where("updated_at <= ?", params[:updated_at_to].to_time) unless params[:updated_at_to].blank?
     @samples = @samples.where("updated_at >= ?", params[:modified].to_time) unless params[:modified].blank?
-    if @samples.sort_column?( params[:sort] )
+    if params[:sort].present? && @samples.sort_column?( params[:sort] )
       @samples = @samples.order( params[:sort] => :desc) unless params[:sort].blank?
     end
 
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -16,7 +16,7 @@ class SamplesController < ApplicationController
     @samples = @samples.where("updated_at >= ?", params[:updated_at_from].to_time) unless params[:updated_at_from].blank?
     @samples = @samples.where("updated_at <= ?", params[:updated_at_to].to_time) unless params[:updated_at_to].blank?
     @samples = @samples.where("updated_at >= ?", params[:modified].to_time) unless params[:modified].blank?
-    if params[:sort].present? && @samples.sort_column?( params[:sort] )
+    if params[:sort].present? && Sample.sort_column?( params[:sort] )
       @samples = @samples.order( params[:sort] => :desc) unless params[:sort].blank?
     end
 
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -16,7 +16,7 @@ class SamplesController < ApplicationController
     @samples = @samples.where("updated_at >= ?", params[:updated_at_from].to_time) unless params[:updated_at_from].blank?
     @samples = @samples.where("updated_at <= ?", params[:updated_at_to].to_time) unless params[:updated_at_to].blank?
     @samples = @samples.where("updated_at >= ?", params[:modified].to_time) unless params[:modified].blank?
-    @samples = @samples.order( params[:sort] + ' DESC') unless params[:sort].blank?
+    @samples = @samples.order(params[:sort] => :desc) unless params[:sort].blank?
 
     @institutions = Institution
       .where()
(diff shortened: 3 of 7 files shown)


diegoliberman:
@omelao it could be useful to add a couple of screenshots in the PRs.  Showing how it looked before the PR and how it will look after merging.  This will help the reviewer and any watcher to (perhaps) spot issues in advance.  We don't do this for all PRs, perhaps because we forget, or because we think it isn't necessary. :)

omelao:
image

app/views/samples/index.haml
```diff
@@ -20,20 +20,23 @@
...
           - t.thead do
             %tr
               %th
               %th Sample Id
-              %th Isolate
+              %th sample_isolate_name
               %th Inactivation
               %th Specimen Role
               %th Production Date
+              %th= link_to "Modified", ( params["sort"] ? request.query_parameters.except(:sort) : request.query_parameters.merge({:sort => 'updated_at'}) ), :class => "#{params["sort"] ? 'ordered' : ''}"
```
ysbaddaden, line 38: There is a little too much happening on a single line, and we can check the sort parameter's value to be future proof.

```suggestion
              - if params[:sort] == "updated_at"
                %th= link_to "Modified", request.query_parameters.except(:sort), class: "ordered"
              - else
                %th= link_to "Modified", request.query_parameters.merge(sort: "updated_at")
```

app/views/samples/index.haml
```diff
@@ -20,20 +20,23 @@
...
+            %col{:width => "12%"}
+            %col{:width => "26%"}
+            %col{:width => "13%"}
+            %col{:width => "15%"}
           - t.thead do
             %tr
               %th
               %th Sample Id
-              %th Isolate
+              %th sample_isolate_name
```
ysbaddaden, line 34: Why the change?

app/views/samples/_filters.haml
```diff
@@ -25,3 +25,20 @@
...
+            .filter
+              %label.block Modified
+              = cdx_select name: "modified", value: params["modified"] do |modified|
+                - modified.item "", "All"
+                - modified.item Date.today, "Today"
+                - modified.item Date.today - 1.days, "Yesterday"
+                - modified.item Date.today - 7.days, "7 Days"
+                - modified.item Date.today - 30.days, "30 Days"
+                - modified.item Date.today - 60.days, "60 Days"
+                - modified.item Date.today - 365.days, "Last Year"
```
ysbaddaden, line 44: We can be even more cool in Rails, for example: `1.day.ago`, `1.week.ago`, `1.month.ago`, `1.year.ago`

ysbaddaden (changes requested):
Nice job! Just one big issue to fix; the rest is to improve things further.

That being said, I think we should have either a pair of from/to date selectors or a drodpown selector to select a range, but both is too much. What do you think @jkicillof and @diegoliberman?

diegoliberman:
> d have either a pair of from/to date selectors or a drodpown selector to select a rang

I agree.  The `From - To` is the most flexible option, so I would keep that.  Sorry for the confusion, the dropdown was an alternative in case it was much easier to do.

jkicillof:
Do we really need from/to? could be a list of time spans be enough? something like last week, last month, last year. If we need and extra row to fit all the filters we could use something like this: [collapsed](https://projects.invisionapp.com/d/main#/console/4103589/118580897/preview), [expanded](https://projects.invisionapp.com/d/main#/console/4103589/93837630/preview) and leave visible by default the most common ones. I don't know if this is already implemented somewhere.

jkicillof:
A few style comments:
1. Combo text should be black 
image
2. Placeholder text should be grey like the label 
image

diegoliberman:
> Do we really need from/to? could a list of time spans be enough? something like last week, last month, last year.
I prefer to be more flexible right now, and allow the user to select whatever range they need.

app/views/samples/_filters.haml
omelao: good to know....but the days of this dropdown are counted...I removed it :)

diegoliberman:
@omelao you can move ahead with deleting the dropdown.  Let us know if you need anything.  Thanks!

app/assets/stylesheets/_global.scss
```diff
@@ -377,3 +377,9 @@ a[title="Back"] {
 .faded {
   color: $gray3;
 }
+input[type=date].filled::-webkit-datetime-edit {
```
straight-shoota, line 380: What's the reason for this webkit-specific selector? And what's the effect in non-webkit browsers?

app/controllers/samples_controller.rb
```diff
@@ -16,7 +16,7 @@ def index
     @samples = @samples.where("updated_at >= ?", params[:updated_at_from].to_time) unless params[:updated_at_from].blank?
     @samples = @samples.where("updated_at <= ?", params[:updated_at_to].to_time) unless params[:updated_at_to].blank?
     @samples = @samples.where("updated_at >= ?", params[:modified].to_time) unless params[:modified].blank?
-    @samples = @samples.order( params[:sort] + ' DESC') unless params[:sort].blank?
+    @samples = @samples.order(params[:sort] => :desc) unless params[:sort].blank?
```
straight-shoota, line 19: We should prevent injecting order by arbitrary columns in to the SQL query. While not necessarily a security risk (although it *can* leak some otherwise concealed information), it can create performance issues when huge tables are sorted by strange keys that are not meant for sorting.

I'd suggest adding a whitelist for column names that are available for sorting (for now it's probably just `updated_at`, right?).
It would also be a good idea to move this order mechanism into a scope on the `Sample` model.

app/assets/stylesheets/_global.scss
omelao: The input date behaves differently in each type of browser and it doesn't obey the placeholder coloring, is a kind of input that the browser builds (I believe it also depends on the operating system). I found several attempts to solve, I tried only with CSS, but in the end, I needed a little jquery too. CSS selectors to change color like `:empty` or `:placeholder-shown` doesn't work on this type of input. Placeholder doesn't work too. It is possible that this has problems in other browsers as well, I forgot to test it on Firefox and maybe someone could test on Edge. This problem was happening on every input date in the system, not just on this screen.

omelao: I didn't find how to do this on Mozilla and Edge. And I can't use placeholder, it doesn't work in Chrome, and if it works on the other browsers, it's not recommended, because the date format depends of user's operating system. So, if I put DD/MM/YYYY in placeholder and user's system is in English, it will fill MM/DD/YYYY.

omelao: @straight-shoota could I mark this issue as resolved?

straight-shoota (approved):

app/controllers/samples_controller.rb
```diff
@@ -13,6 +13,12 @@ def index
     @samples = @samples.joins("LEFT JOIN batches ON batches.id = samples.batch_id").where("batches.batch_number LIKE concat('%', ?, '%') OR samples.old_batch_number LIKE concat('%', ?, '%')", params[:batch_number], params[:batch_number]) unless params[:batch_number].blank?
     @samples = @samples.where("isolate_name LIKE concat('%', ?, '%')", params[:isolate_name]) unless params[:isolate_name].blank?
     @samples = @samples.where("specimen_role = ?", params[:specimen_role]) unless params[:specimen_role].blank?
+    @samples = @samples.where("updated_at >= ?", params[:updated_at_from].to_time) unless params[:updated_at_from].blank?
+    @samples = @samples.where("updated_at <= ?", params[:updated_at_to].to_time) unless params[:updated_at_to].blank?
+    @samples = @samples.where("updated_at >= ?", params[:modified].to_time) unless params[:modified].blank?
+    if @samples.sort_column?( params[:sort] )
+      @samples = @samples.order( params[:sort] => :desc) unless params[:sort].blank?
+    end
```
ysbaddaden, line 21: Maybe check for blank before sortable and combine them?

```ruby
if params[:sort].present? && Sample.sort_column?(params[:sort])
  # ...
end
```

app/views/samples/_filters.haml
```diff
@@ -25,3 +25,10 @@
             = cdx_select name: "specimen_role", value: params["specimen_role"] do |specimen_role|
               - specimen_role.item "", "All specimen roles"
               - specimen_role.items Sample.specimen_roles.map{|item| {id: item[:id], description: item[:description].truncate(30) }}, :id, :description
+          .filter-group
+            .filter
+              %label.block Modified From
+              %input.input-large{type: "date", name: "updated_at_from", value: params["updated_at_from"]}
+            .filter
+              %label.block Modified To
+              %input.input-large{type: "date", name: "updated_at_to", value: params["updated_at_to"]}
```
ysbaddaden, line 34: If you keep the `filled` class, you could specify it right into the HTML, so you wouldn't need to run a JS function on page load to set it. For example:

```suggestion
              %input.input-large{type: "date", name: "updated_at_to", value: params["updated_at_to"], class: ("filled" if params["updated_at_to"].present?)}
```

app/assets/javascripts/components.js
```diff
@@ -17,3 +17,9 @@ function cdx_select_on_change(name, callback) {
     callback($(this).val());
   });
 }
+$(document).on("change", "input[type=date]", function(){
+  $(this).val() != '' ? $(this).addClass("filled") : $(this).removeClass("filled");
+});
+$(function(){
+  $("input[type=date]").change();
+});
```
ysbaddaden, line 25: When I see JS I'm always wondering if we could achieve the same effect in pure CSS, and quite often it happens that we can. Consider:

```scss
input[type=date]:not([value='']) {
  // style for filled element
}
```

ysbaddaden: I'm also a little wary of this piece of JS, because it will affect _all_ date input types (same for the CSS selector above), and maybe we want it to to only apply to the date selector appearing in filters?

But maybe this case is fine, because we probably want all date inputs to look the same across the site :+1:

ysbaddaden (commented):
I just left a few comments.

app/assets/javascripts/components.js
ysbaddaden: (I also found the `:placeholder-shown` pseudo selector, but it doesn't match the default values of a date type, only inputs with an actual placeholder).

straight-shoota: It would be nice if `[value='']` worked. But the `value` attribute is not dynamically bound. It does not change when editing the field value in the UI. You could use a javascript handler to do that: `onchange="this.setAttribute('value', this.value);"`. But then we're back at JS 🤷 
An alternative would be `:not(:placeholder-shown)`, which requires a `placeholder` attribute (although it can be empty). Unfortunately, this does not work reliably with date inputs. Their widget typically doesn't show a placeholder.

Last option would be `:valid` with form validation attributes, but I think that's probably too much of a hack. Filled doesn't necessarily mean valid.

app/views/samples/_filters.haml
omelao: The JS function is not just for content that loads from server. But when the users clears the field, it will work too. That's why I did it through javascript, and I realized too that this is a change I had to do on every input date. So it works for all.

app/assets/javascripts/components.js
omelao: @ysbaddaden the idea is exactly to make it work on every input date, because it's the default style of a input date, and it was wrong on other pages, like the initial dashboard.

I also found selectors like `:placeholder-shown`, `:valid`, `:empty`. I talked about this to @straight-shoota in somewhere, maybe another PR. They doesn't work. Input date works in a very particular way.

omelao: I spent a long time before deciding to make it works like this. I know that it doesn't look good when you see it at the first time, and I think I should explain it better when I opened my PR, to help you with these doubts. My fault. That's the best way I found, because of many conditions.

app/controllers/samples_controller.rb
omelao: Great! You are right

app/assets/javascripts/components.js
omelao: @ysbaddaden look what happens when I change the value of a input date, using the `[value=''"]` selector:
image

straight-shoota: It's probably a good idea to add a short comment in the code to explain the reasoning for this solution. This will avoid future me and you from having to think about it again.
That's a good practice whenever you need to research a solution, to add a comment how you ended up with that. That makes it easy for a reviewer to understand your reasons and make reviews easier.

omelao: Great idea @straight-shoota you are right. Even I can forget about it all. Should I put this on JS? I think it's the place where we have the most unexpected solution, right?

omelao:
@straight-shoota @ysbaddaden what do you think about this comment?

/*
This is kind of a exception solution. Because input date's behavior are different. 
Consider this points:
  - input date's placeholders doesn't work
  - even if at some point in the future it works, date format's depends of the browser and OS
  - CSS selectors like :empty, :valid, [value=''], :placeholder-shown doesn't work
  - Google Chrome has specific selector to coloring the content of this kind of fields, but other browser doesn't

  Finally you can test if something changes on the future on this fiddle: 
    https://jsfiddle.net/omelao/hsk8vyax/15/
*/

omelao:
Maybe I could put all of this on the fiddle, and just link the fiddle on the comment.

PR #1742, opened 2022-07-01
ActiveRecord RecordNotFound redirect

omelao:
Redirected to home


omelao:
I was testing some possibilities and got the same error when I'm in Transfer Detail screen and change to another institution.

image

Issue #1598, opened 2022-04-12
Show date last modified date in Samples


diegoliberman:
### Use case that triggered this req
A user needs to know when Assays have been uploaded into the Q sample.  Today, the person is accessing each Q sample to see if there are new assays for any.

### Idea
In order to make a general solution, we could add a column in the Sample's listing, showing the last modified date.  The date should be updated when any sample's attribute is modified, including Assays or Notes.

### Acceptance criteria
1. I attach an Assay to a Sample, save, and in the list of samples, it's shown with today's date in a `Modified` column.
2. I make any modification to a sample, and the `Modified` date is updated with today`s date. 
3. I have a way to quickly find the samples that were modified most recently.

### Options for finding samples that were modified recently
1. Allowing to sort by Modified, descending.  Sorting hasn't been implemented anywhere else in the site.
2. Allowing to filter by a start/end date.  This has been implemented in the Batches list.
3. Adding a dropdown filter labeled "Modified" with these options: today, yesterday, last 7 days, last 30 days, last 60 days, last year.

straight-shoota:
I suppose this should also include sorting and/or filtering by modification date? Showing the date on the index page is certainly already an improvement, but when working with lots of samples to keep track of, I imagine it would help a lot to be able to sort and/or filter by date.

diegoliberman:
You're absolutely right, it needs to be implemented along with something else.  I would start with a simple filter with fixed options: modified in the last 7 days, 14 days, 30 days.   I'm not suggesting to start with a sort, because we don't have any other sort in the page, so it might require more effort than adding a filter.

straight-shoota:
The implementation is actually quite similar and any differences should really be neglibible. So I think it should be up to what's most useful.

diegoliberman:
We'll decide when the implementation time comes.  Maybe both? Who knows :)

diegoliberman:
I think the sorting could be more useful.

mmuller:
In version `7100246`: 
-after clicking on the `Modified` header, the samples sorting doesn't change.
-the `Modified` date is not updated after adding an Assay. It happens **randomly**.

diegoliberman:
> In version `7100246`: -after clicking on the `Modified` header, the samples sorting doesn't change. -the `Modified` date is not updated after adding an Assay. It happens **randomly**.

@mmuller we didn't implement sorting when clicking on the header.  Instead, there are 2 new filters to select a date range.

omelao:
[...]

PR #1751, opened 2022-08-09
Fix Sorting Samples Modified Column

omelao:
The list to allow what column should sort was with
variables, changed to string.


diegoliberman:
Related with #1598

omelao:
@diegoliberman  This one is a fix I did on a new branch, because of the ticket was already closed and the branch deleted.

Issue #1771, opened 2022-10-24
Logic for blinded boxes


diegoliberman:
Make sure the code is honoring the logic described in #1755 within the section `Logic for blinded boxes` and make sure the app is working as expected, as described in `Test cases for a blinded box` and `Test cases for an unblinded box`.

sardar-usman:
@leandroradusky It is working as expected.

Issue #1755, opened 2022-08-19
Simplify boxes, transfers and blinding flow


diegoliberman:

ysbaddaden:
The main technical problem I see are permissions (and it may have user facing repercussions):

The permission system is per-resource-type (e.g. boxes, samples) for an institution (and optionally per site), and each resource _directly_ belongs to an institution and site.

With transfers, we but kinda retained it (but had to cheat): the transfer boxes and samples would no longer belonging to anything during the transfer (cheat n°1), yet the transfer resource still had a direct relationship to an institution, erm... actually two institutions (cheat n°2).

With the new changes, transferred boxes and samples shall be accessed from their direct owner —once the transfer is confirmed, the destination becomes their direct institution— but also from indirect relationships: box -> transfer and worse: sample -> box -> transfer!

Add the ability to re-transfer a previously transferred box or sample, and... we got a mess to take care of. Let's say we have institutions A, B, C and D, plus samples S1, S2 and S3.

1. A transfers a box B1 with S1, S2 and S3 to B:
   - B now owns B1 as well as S1, S2 and S3;
   - A can still read B1, S1, S2 and S3 + unblind B1.
2. B transfers a box B2 with S1 and S2 to C:
   - C now owns B2 as well as S1 and S2;
   - S1 and S2 now belongs to both B1 and B2 (and to both transfers);
   - B can still read B2 and S1 and S2 + unblind B2;
   - A can still read B1, S1, S2 and S3 + unblind B1.

We're already starting to lose scope of who can see the samples (despite both A and B not having the samples anymore), but worse are the actions that A can still do (unblind).

This may still be alleviated if we keep the inability to transfer a blinded sample. In the previous example B would only transfer S1 and S2 to C after they're unblinded, and blind them _again_ if they wish: A wouldn't be able to unblind them anymore.

And I'm not talking about re-transferring a (blinded) box...

diegoliberman:
@ysbaddaden transferring more than once doesn't have a use case to justify it.  Just all of this discussion is enough for changing my mind about the feature, let's keep the prohibition of transferring a 2nd time.

omelao:
The problem on this case is when I transfer a box, I can't see it anymore. You asked me once to put the blinded option on Box detail, and it wasn't possible because of this. The sender can't see the box anymore after transfering.

omelao:
[...]

PR #1779, opened 2022-11-03
Simplify boxes transfer blinding flow - Stage 1

### LINKED ISSUE #1768: Recover inventory ordering

leandroradusky:
Reverted the changes of #1746.

Diego and I performed a smoke test about functionality recovery and it's there. The full list of the use cases will be tested after the completion of all the functionalities of the stabilization stage of boxes V2 (#1771). 

Some tests written/modified in the reverted PR are useful and were kept, here is a list of them: 

- **Print floating box shouldn't be accessible to the receiver:** To accomplish this we recover in the specs setup the declaration of the floating box (and the confirmed box as well which is useful for some tests described below). Also, the method `load_box_print` within the boxes controller was kept in order to be able to access the floating and confirmed boxes. 
- **Inventory & print actions should be allowed if can read:** these are done over the confirmed transferred box and not over the box as were implemented in the PR (#1746). 
- **Inventory should be ordered by batch_number, concentration, replicate ASC:** The ordering of the inventory is verified in this test and corresponds with the use case described in (#1755), and is passing, this closes issue (#1768).

### DIFF
--- a/app/assets/stylesheets/_barcode_card.scss
+++ b/app/assets/stylesheets/_barcode_card.scss
@@ -10,7 +10,7 @@
 
     background: $white;
     padding: 5px 7px;
-    margin: 15px 15px;
+    margin: 15px 15px 7px;
 
     .title {
       font-size: 14px;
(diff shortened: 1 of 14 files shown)


ysbaddaden:
You may drop most of the third commit, the specs are testing the behavior you just reverted. I'd only keep the spec for the inventory order, which adds value.

ysbaddaden:
Then, please merge :)

Issue #1777, opened 2022-11-02
Fix bug where the box is blinded even when not enabling the check box


diegoliberman:

leandroradusky:
After removing the blinding checkbox from the transfer detail this incorrect behavior was lost, verify this is corrected when committing #1769 and close both issues in the same PR.

diegoliberman:
Nothing to review here, just move it to QA when merged so that we make sure the bug isn't there.

PR #1780, opened 2022-11-04
Fix/1769 blind checkbox from transfer to box detail

### LINKED ISSUE #1769: Move the blind check box to the box detail

leandroradusky:
Closes #1769 and #1777.

The blind samples checkbox is now in the 'New Box' form and was removed from the transfers.

image

Some comments about the code:
- The `blinded` param was removed from the `TransferPackages` controller and moved into the `Boxes` controller.
- The `BoxForm` build method read the param mentioned above.
- The `TransferPackage` model doesn't have the `attr_accessor :blinded` anymore, this has effects on several tests, where  `make!` was requiring it. New blueprints were created to `make!` transfers of blinded boxes. 
- The `blinded` attribute of the boxes is mandatory, tests creating boxes were modified accordingly to create the boxes with this attribute.
- The test verifying that transferring a box and blinding it was effectively blinding the samples was removed. There is already test for boxes that verify the correct blinding then no new tests were added. 

Functional tests made:
- A box is created unblinded, samples are unblinded for the creator, then is transferred, confirmed and samples remain unblinded for the receiver(#1777 solved!). The automatic blinding was happening.
- A box is created blinded, samples are blinded for the creator, then is transferred and confirmed, samples remain blinded for the receiver, then is unblinded, samples unblind for the receiver.

### DIFF
--- a/app/controllers/boxes_controller.rb
+++ b/app/controllers/boxes_controller.rb
@@ -149,12 +149,12 @@ class BoxesController < ApplicationController
 
   def box_params
     if Rails::VERSION::MAJOR == 5 && Rails::VERSION::MINOR == 0
-      params.require(:box).permit(:purpose, :media).tap do |allowed|
+      params.require(:box).permit(:purpose, :media, :blinded).tap do |allowed|
         allowed[:batch_uuids] = params[:box][:batch_uuids].try(&:permit!)
         allowed[:sample_uuids] = params[:box][:sample_uuids].try(&:permit!)
       end
     else
-      params.require(:box).permit(:purpose, :media, batch_uuids: {}, sample_uuids: [])
+      params.require(:box).permit(:purpose, :media, :blinded, batch_uuids: {}, sample_uuids: [])
     end
   end
 
--- a/app/controllers/transfer_packages_controller.rb
+++ b/app/controllers/transfer_packages_controller.rb
@@ -126,7 +126,6 @@ class TransferPackagesController < ApplicationController
       :receiver_institution_id,
       :recipient,
       :includes_qc_info,
-      :blinded,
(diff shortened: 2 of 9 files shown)


omelao:
I could be wrong in what I'm going to say but...the `blinded` field was already in `box model`. In transfers it was like a fake field, just to set it in the boxes. The information in database was always stored on `boxes`.

> A box is created blinded, samples are blinded for the creator, then is transferred and confirmed, samples remain blinded for the receiver, then is unblinded, samples unblind for the receiver.

The question is, how do you unblind this boxes if the sender doesn't see it anymore? Who is unblinding?

leandroradusky:
Hi @omelao ... Within the transfer there is yet the unblind button, which unblinds the boxes. The code doing this is 

```
  def unblind_boxes!
    boxes.update_all(blinded: false)
  end
```

Called by `unblind` method in the `TransferPackagesController`. As long as I tested it was working ok.

ysbaddaden (approved):
It looks good to me. Permissions aren't an issue until we allow to _edit_ a box. I just left a comment to improve a spec.

Once the conflicts with main are fixed, please merge!

omelao:
These `@floating_transfer` and `@confirmed_transfer` on `box_controller_spec.rb` were created to test the printing buttons on transfers. I think we don't need some specs anymore.  I will check this tomorrow.

omelao:
line 202 - 209 `it "confirmed box should be accessible to sender" do`, we don't need this anymore. The other specs I did are still relevant.

omelao (commented):
I don't know how to mention unchanged lines on my review. But you need to remove `box_resource` function from line 26, 38 and the entire function lines 172-183.

PR #1783, opened 2022-11-07
Fix/1776 printing blinded labels

leandroradusky:
To correctly print the blinded attributes, both in the normal and the bulk print, the `SamplePresenter` object is passed to the PDF generator (instead of `Sample` as it was before, which wasn't blinding. 

Before, incorrectly, blinded attributes were printed, now we can observe they don't: 
 
image

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -117,12 +117,13 @@ class SamplesController < ApplicationController
 
   def print
     @sample = Sample.find(params[:id])
+    @sample_form = SamplePresenter.new(@sample, request.format)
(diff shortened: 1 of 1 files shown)


ysbaddaden (approved):
Neat!

I'm just not fond of `form` in the naming since they're not a `SampleForm` it could be confusing. Maybe we can just avoid naming a variable? For example:

```ruby
locals: { sample: SamplePresenter.new(@sample, request.format) }

locals: { samples: samples.preload(:sample_identifiers).map { |s| SamplePresenter.new(s, request.format) } }
```

PR #1787, opened 2022-11-10
Improve flexibility samples concentration

### LINKED ISSUE #1753: Improve flexibility for defining sample's concentrations
1. In the [sample detail](https://projects.invisionapp.com/d/main#/console/4103589/467087061/comments), the field for entering the concentration should be just an input box expressing the unit `copies/µl`.  The user may enter absolute numbers such as `600` or scientific notation such as `75e+6`.
   1. When creating a sample, the user may enter any of those 2 formats. 
   2. When editing a sample, the system will show the absolute number if it's less than 5 figures.  Else the scientific notation will be used.
2. Concentrations that were saved before these changes: we need to delete the concentration value and exponent, and make sure we store the corresponding absolute number that can be used for correctly displaying as mentioned in item 1.2.
3. Creating boxes should honor the new way of storing concentrations.  Even when we're planning to implement V2 of boxes, let's please implement the concentration changes and adapt other features if necessary.  The effort will be harvested in V2.
4. The concentration is included when downloading the inventory of a box.  Make sure we're populating this column with the absolute value of the concentration.  
5. The concentration is one of the blinded fields.  Make sure the blinding feature is still working as expected in regards to the blinding and unblinding of the concentration.

### Acceptance criteria
1. I can enter a concentration in the sample detail as `500` and `75e+6` ([scientific notation](https://support.microsoft.com/en-us/office/display-numbers-in-scientific-exponential-notation-f85a96c0-18a1-4249-81c3-e934cd2aae25#:~:text=The%20Scientific%20format%20displays%20a,10%20to%20the%2010th%20power.)).
2. I open a sample with a concentration greater than 99,999 and the absolute number is shown.
3. I open a sample with a concentration lower than 100,000 and it's shown in the scientific notation.
4. I create an LOD box and the concentrations are shown (within the sample detail) as described in 3 and 4.
5. Same as 5 for Variants.
6. Same as 5 for Challenge.
7. I can download the inventory of a box and the Concentration column is populated with the absolute number.
8. If I blind a box, and transfer it the recipient can't see the concentration the sample detail and in the inventory.
9. If I don't blind a box and transfer it, the recipient sees the concentration in the sample detail and in the inventory.

leandroradusky:
Close #1753 

Sample concentration is now a unique field accepting scientific notation or raw numbers. The concentration is stored within the DB as an integer value: it stores the number of copies per milliliter which is an integer, given that scientific notation can express as well decimal numbers, if the inputted number is `float` only the integer part is kept. 

**Comments about the code:** 

- Tested input formats: 2.2e4, 2.2E4, 2.2e+4, 2.2E+4, 2.2e+04, 2.2E+04, all of them work. When numbers are big and sample detail is accessed is shown in 2.2e+4 format.

image

- As can be seen in the image above, concentration is now medium-sized instead of small-sized since for scientific notation it was too small. Also, if the number exceeds the 8 figures we're converting it to scientific notation on display to improve readability with `%g` which removes only the trailing zeros in the conversion.

- `:concentration_number` and `:concentration_exponent` were `core_fields` within the sample, which explains why the migration is done this way.

- In the sample model the remaining code to deal with numerical values looks like this now:

```
  # Attribute fields don't convert numerical types.
  %w[replicate].each do |name|
    define_method name do
      core_fields[name].presence.try(&:to_i)
    end
  end

  # Fields accepting exponentiation (e.g. 1e8) first are converted to float then to integer.
  %w[concentration].each do |name|
    define_method name do
      core_fields[name].presence.try(&:to_f).try(&:to_i)
    end
  end
```

It makes sense to do the `each`? (cc @ysbaddaden)

- Tests were modified to keep the same numbers that were used before. 

** Comments on the acceptance criteria of the issue: **

- Concentration on the inventory is now printed as a raw number, we leave on excel the dealing of scientific notation:

image

If we make the column smaller: 

image

- The other cases of acceptance criteria were tested and they work but no comments are needed here :)


app/helpers/sample_helper.rb
```diff
@@ -6,4 +6,12 @@ def loinc_code_description(loinc_code)
       loinc_code.description
     end
   end
+
+  def concentration_formatting(concentration)
+    if concentration.to_s.length > 8
+      ("%g" % concentration.to_s).to_s
```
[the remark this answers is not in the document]
ysbaddaden: Maybe we could also rename the method as `humanize_concentration` to be more rails-like and self explanatory?

db/migrate/20221109105200_change_sample_concentrations_to_single_value.rb
```diff
@@ -0,0 +1,19 @@
+class ChangeSampleConcentrationsToSingleValue < ActiveRecord::Migration
+
+  def up
+    Sample.find_each do |sample|
+      if sample.core_fields.key?("concentration_number") && sample.core_fields.key?("concentration_exponent")
+        sample.core_fields["concentration"] = sample.core_fields["concentration_number"].to_i * (10**(sample.core_fields["concentration_exponent"].to_i))
```
ysbaddaden, line 6: nitpick: we could improve readability by using variables

```suggestion
      number = sample.core_fields.key?("concentration_number").try(&:to_i)
      exponent = sample.core_fields.key?("concentration_exponent").try(&:to_i)

      if number && exponent
        sample.core_fields["concentration"] = number * (10**exponent)
```

db/migrate/20221109105200_change_sample_concentrations_to_single_value.rb
```diff
@@ -0,0 +1,19 @@
...
+        sample.core_fields.delete("concentration_number")
+        sample.core_fields.delete("concentration_exponent")
+        sample.save
+      else
+        sample.core_fields["concentration"] = 0
+        sample.save
```
ysbaddaden, line 12: I don't think we want to set a default concentration for all samples. Let's leave them as `nil`.

spec/controllers/boxes_controller_spec.rb
```diff
@@ -267,18 +267,17 @@
...
       # NOTE: can't where/order in SQL because entity fields...
-      samples = batch.samples.to_a.reject { |s| s.concentration_number != concentration_number }.sort! do |a, b|
-        cmp = a.concentration_exponent <=> b.concentration_exponent
-        cmp = a.replicate <=> b.replicate if cmp == 0
-        cmp
-      end
+      concentrations.each do |c|
+        samples = batch.samples.to_a.reject { |s| s.concentration != c }.sort! do |a, b|
+          cmp = a.replicate <=> b.replicate
+          cmp
```
ysbaddaden, line 275: Let's simplify a bit:

```suggestion
        samples = batch.samples.to_a
          .reject { |s| s.concentration != c }
          .sort_by(&:replicate)
```

spec/models/sample_spec.rb
```diff
@@ -62,8 +62,8 @@
   end
 
   it "concentration" do
```
ysbaddaden, line 64: We don't need this spec anymore.

ysbaddaden (changes requested):
I left a few comments to further cleanup, but :+1: 

Juts one thing to fix: don't set a default concentration of `0` to all samples. It's not a required field, and it should be left to `nil` by default.

ysbaddaden (approved):

Issue #1786, opened 2022-11-09
Print and download samples


diegoliberman:

omelao:
You meant these buttons?
<img width="401" alt="image" src="https://user-images.githubusercontent.com/19227697/201237456-12db1ce1-1d23-4fda-b9e2-7ebfebe30f92.png">

<img width="251" alt="image" src="https://user-images.githubusercontent.com/19227697/201237498-edf5df4e-e52b-4276-a5e6-c0a211c5806d.png">

sardar-usman:
@omelao it is working as expected.

PR #1789, opened 2022-11-14
Simplify boxes transfer blinding flow - Stage 2

### LINKED ISSUE #1775: Downloading and printing boxes
1. When the sender creates a blinded box, they can print the blinded labels, and they can download the unblinded inventory or the blinded inventory.
2. When the sender creates an unblinded box, they can print the unblinded labels and they can download the unblinded inventory.
3. When the recipient access a blinded box, they can print the blinded labels, and they can download the blinded inventory.
4. When the recipient access an unblinded box, they can print the unblinded labels and download the unblinded inventory.

leandroradusky:
Closes #1775 and #1771.

Changes made to meet the acceptance criteria for the remaining issues of the boxes stabilization.

**Issue #1775 (Downloading and printing boxes)**:  there are comments to make only in one of the use cases described:

- `When the sender creates a blinded box, they can print the blinded labels and download the unblinded or the blinded inventory.`

_Respect to labels:_ 
Within the `/views/boxes/_barcode_card.haml` file there was the condition: " `unless @box.blinded?` show the print button",  which prevented showing the print button for blinded boxes. This condition was removed in order to allow the printing of blinded labels. 

_Respect to inventory:_ 
(This is also mentioned in #1755: "The Sender of the box can see an action for downloading an unblinded inventory. The Unblinded inventory action")

After discussing with @diegoliberman and @omelao this behavior was specified in the issue [#1786]( #https://github.com/instedd/cdx/issues/1786) which is actually in `open` state. 

This behavior was implemented and now it looks like this.

image

I don't consider that the mentioned issue is closed because there are changes in style for this and other views considered in the issue that aren't part of this PR. 

With respect to the code, this button is implemented under the condition `@box.blinded && @box.box_transfers.count == 0`. If the box is unblinded is unnecessary and this button should not be accessible to box receivers. The style for the `Download Inventory` buttons was renamed, but kept in the `_barcode_card.scss`. There is no specific styles file for boxes, should I make a new styles file? 

The rest of the cases of #1775 work correctly.

**Issue #1771  and other test cases of #1755**:

- `Transferring more than once: "let's keep the prohibition to avoid risks, as we don't have a use case to justify a 2nd transfer"` 

Thanks to @omelao input, this was restored by adding 

```
      .left_joins(:box_transfers)
      .where(box_transfers: {id: nil})
```
To the `find_box` method on the `TransferPackagesController`. 

-  `The sender should be able to edit a box, but for now, just allow to edit the blind check box`

This behavior is not present actually and after discussion with @diegoliberman this will be implemented on the new increment, since actually when entering a box detail we are viewing them but not editing the boxes.

- `The user might want to manually change the concentration of some individual samples, after creating the box. a) An unblinded box can always be edited. b) A blinded box can be edited only if it doesn't have an associated transfer record.`

As the latter use case, will be implemented in the new increment.

The rest of the use cases were tested and worked correctly.

### DIFF
--- a/app/presenters/sample_presenter.rb
+++ b/app/presenters/sample_presenter.rb
@@ -1,11 +1,11 @@
 class SamplePresenter
   delegate_missing_to :@sample
 
-  def self.map(samples, format, unblind= false)
-    samples.map { |sample| new(sample, format, unblind) }
+  def self.map(samples, format, unblind: false)
+    samples.map { |sample| new(sample, format, unblind: unblind) }
   end
 
-  def initialize(sample, format, unblind= false)
+  def initialize(sample, format, unblind: false)
     @sample = sample
     @format = format
     @unblind = unblind
--- a/app/presenters/sample_presenter.rb
+++ b/app/presenters/sample_presenter.rb
@@ -1,13 +1,14 @@
 class SamplePresenter
   delegate_missing_to :@sample
 
-  def self.map(samples, format)
-    samples.map { |sample| new(sample, format) }
+  def self.map(samples, format, unblind= false)
+    samples.map { |sample| new(sample, format, unblind) }
   end
 
-  def initialize(sample, format)
+  def initialize(sample, format, unblind= false)
     @sample = sample
     @format = format
+    @unblind = unblind
   end
 
   def to_param
@@ -18,7 +19,7 @@ class SamplePresenter
 
   Box.blind_attribute_names.each do |attr_name|
     define_method attr_name do
-      if blinded_attribute?(attr_name)
+      if blinded_attribute?(attr_name) && !@unblind
         blinded_value
       else
         @sample.__send__(attr_name)
--- a/app/views/boxes/show.haml
+++ b/app/views/boxes/show.haml
@@ -30,17 +30,29 @@
       .col
         = render "barcode_card"
 
-        .barcode-after
-          = link_to inventory_box_path(@box, format: "csv"), target: "_blank", class: 'btn-link' do
-            .icon-download.btn-icon
-            %span Download inventory
-
 .row
+  .col.pe-3
+    %label Contents
   .col
     .row
       .col.pe-3
-        %label Contents
-      .col.pe-9
+        .samples-count
+          .title
+            = @box.samples.count.to_s + ' ' + "samples" + (@box.blinded ? ' (blinded)' : '' )
+      .col.pe-3
+        .inventory-button
+          = link_to inventory_box_path(@box, format: "csv"), target: "_blank", class: 'btn-link' do
+            .icon-download.btn-icon
+            %span Download inventory
+      - if @box.blinded && @box.box_transfers.count == 0
+        .col.pe-3
+          .inventory-button
+            = link_to inventory_box_path(@box, format: "csv", unblind: true), target: "_blank", class: 'btn-link' do
+              .icon-download.btn-icon
+              %span Download unblinded inventory
+
+    .row
+      .col
         - @samples.each do |sample|
           .col.pe-7.batches-samples
             .samples-row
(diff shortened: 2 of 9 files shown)


app/presenters/sample_presenter.rb
```diff
@@ -1,13 +1,14 @@
 class SamplePresenter
   delegate_missing_to :@sample
 
-  def self.map(samples, format)
-    samples.map { |sample| new(sample, format) }
+  def self.map(samples, format, unblind= false)
```
ysbaddaden, line 4: I would make it keyword argument: 

```suggestion
  def self.map(samples, format, unblind: false)
```

app/presenters/sample_presenter.rb
```diff
@@ -1,13 +1,14 @@
...
+    samples.map { |sample| new(sample, format, unblind) }
   end
 
-  def initialize(sample, format)
+  def initialize(sample, format, unblind= false)
```
ysbaddaden, line 8: Same here:
```suggestion
  def initialize(sample, format, unblind: false)
```

app/views/boxes/show.haml
```diff
@@ -30,17 +30,29 @@
...
-        .barcode-after
+.row
+  .col.pe-3
+    %label Contents
+  .col
+    .row
+      .col.pe-3
+        .samples-count
+          .title
+            = @box.samples.count.to_s + ' ' + "samples" + (@box.blinded ? ' (blinded)' : '' )
```
ysbaddaden, line 41: Let's interpolate:
```suggestion
            = "#{@box.samples.count} samples" + (@box.blinded ? ' (blinded)' : '')
```

ysbaddaden (commented):
There are 2 potential security issues:

1. The `unblind` parameter is good but... what if the receiver manually adds `unblind=true` to the download URL? They'll download the unblinded inventory (oops). We must verify in `#print` that the box was never transferred _or_ that the current institution is the sender when the `unblind` parameter is present (fail with a 403 forbidden error).

2. I can't find a previously transferred box when creating a transfer package, but I can manipulate the request to pass the `id` of a whatever box I own in the form. There should be a check in TransferPackagesController or a validation in TransferPackage to ensure that.

These can be handled in subsequent pull requests.

Otherwise the PR is good, I just left a few suggestions to improve readability a bit, but I'm also some tests. For example:

- `transfer_packages#find_box` won't find a previously transferred box.
- pass `unblind=true` to `boxes#print`, then match the PDF (it's just text) to see whether we have ` (blinded)` (should not);
- same but don't don't pass `unblind=true` :)

If you can add these tests today, please do so, otherwise let's merge and create issues to work on this ASAP, so we can deploy to staging for review.

app/views/boxes/show.haml
leandroradusky: This was copypasted from the batches' samples screen, I changed it here but didn't modify it for the batches... should do it?

ysbaddaden: Sure, let's :)

leandroradusky:
Applied changes to deal with security issues mentioned by Julien, tests were added as suggested, and comments were solved except for one that throws an error and I commented it above.

**Security issues**

1) To avoid this flaw I added the line 
```
return head :forbidden unless !(params[:unblind] && @box.transferred?)
```
To the `BoxesController` which prevents printing an unblinded inventory if the box is already transferred (`transferred?` method was created in boxes model.

2) To avoid this flaw I added, similarly, a check that the box to be transferred is not `transferred?` in the transfer packages controller.

**Suggested tests**

- The first test suggested already exists, it was called `blocks transfered boxes`, its name is correct but wasn't inferrable, but it checks that the box isn't found if its transferred.
- The second test suggested was added, checking the inventory is unblinded if `unblind` parameter is true.
- The third test already existed, which checks that the inventory is blinded if `unblind` isn't passed.

ysbaddaden (commented):
Neat! Just a couple suggestions to avoid double negations (unless not) and let's merge!

PR #1791, opened 2022-11-14
Upgrade to `mini_racer` gem

### LINKED ISSUE #1790: Replace abandoned 'therubyracer' gem / XML device message parser
The `therubyracer` gem has been abandoned for years, and uses a very old version of V8. We use it compile assets, so we can replace it with anything, but it appears we also use it to run arbitrary code to parse manifest fields (see `ManifestFieldMapping`).

ysbaddaden:
closes #1790 

I implemented a custom XML serializer to expose a trivial data structure to the JavaScript context because existing accessible solutions (e.g. Hash.from_xml()` from ActiveSupport) weren't very reliable (e.g. override attributes with contents), or required to introduce yet another dependency to handle something quite easy to implement in the end, and where we're in full control of the generated data structure (e.g. we can move to another XML library).

### DIFF
--- a/spec/models/manifest_operations_spec.rb
+++ b/spec/models/manifest_operations_spec.rb
@@ -122,10 +122,26 @@ describe Manifest, validate_manifest: false do
       {"sample" => {"custom" => {"fields" => "#{loc.name},#{loc.lat.to_i},#{loc.lng.to_i}"}, "pii" => {}, "core" => {}}}, device
     end
 
-    it "loads xml in javascript" do
+    it "loads xml in javascript (serialized hash)" do
       assert_manifest_application %(
         {
-          "patient.name": {"script": "message.xpath('Patient/@name').first().value"}
+          "patient.name": {"script": "message.children.find(function (node) { return node.name === 'Patient' }).attributes.name"}
+        }
+      ), %(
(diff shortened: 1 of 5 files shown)


spec/models/manifest_operations_spec.rb
```diff
@@ -122,10 +122,26 @@
...
-          "patient.name": {"script": "message.xpath('Patient/@name').first().value"}
+          "patient.name": {"script": "message.children.find(function (node) { return node.name === 'Patient' }).attributes.name"}
+        }
+      ), %(
+        {
+        }
+      ), {xml: %(
+        <Message>
+          <Patient name="Socrates" age="27"/>
+        </Message>
```
ysbaddaden, line 136: For example this XML document is serialized as:

```json
{
  "name": "Message",
  "children": [
    {
      "name": "Patient",
      "attributes": [{"name": "Socrates", "age": "17"}]
    }
  ]
}
```

diegoliberman:
Let's wait until January to merge this, so that we send it to production when Julien is back from vacations.

leandroradusky:
Looks good to me :)

leandroradusky (approved):

Issue #1792, opened 2022-11-15
Upgrade Dependencies


ysbaddaden:
This is a long running epic to keep track of efforts to upgrade the major dependencies of CDx.

The list below is non exhaustive, items will probably be added as we upgrade and investigate the possibility to move on to the next item. It is, however, in the recommended order to follow.

**Preparations**
- [x] ~~upgrade to latest Rails 4.2.x~~ — done in #1511
- [x] ~~enabled dual boot~~ — done in #1549

**Rails 5**
- [x] ~~upgrade to Rails 5.0~~ —done in #1549
- [ ] upgrade to Ruby 2.3 —done in #1781
- [ ] upgrade to Ruby 2.4
- [ ] replace `therubyracer` gem —see #1790
- [ ] upgrade assets (blocks next Rails upgrade, because gems)
- [ ] upgrade to React 0.14 (blocks next Rails upgrade, because of `react-rails`)
- [ ] upgrade to Rails 5.1 —see #1236 
- [ ] upgrade to Ruby 2.5
- [ ] upgrade to Rails 5.2 —see #1237

**Rails 6**
- [ ] upgrade to Rails 6.0 (Maintained) —see #1238
- [ ] upgrade to Ruby 2.6
- [ ] upgrade to Rails 6.1 (Maintained)
- [ ] upgrade to Ruby 2.7 (EOL 2023/04)
- [ ] upgrade to Ruby 3.0 (Maintained)

PR #1794, opened 2022-11-15
Upgrade to Ruby 2.4

### LINKED ISSUE #1723: Improve performance of the print PDF actions
The current implementation required to changed a server timeout.  See this [thread](https://github.com/instedd/cdx/issues/1715).

The goal of this ticket is to improve the performance, if possible, to avoid future problems in scenarios that may have more than 108 samples for printing.  The most likely use cases are:
1. There is a new box's purpose that requires more than 108 samples.
2. The user creates a batch with more than 108 samples and decides to print all by selecting all samples and printing from the batch detail.

ysbaddaden:
- [x] upgrade to Ruby 2.4;
- [x] fix deprecations (Fixnum, OpenSSL);
- [x] upgrade gems (to fix above deprecations);
- [x] upgrade `encryptor` gem (fixing the algorithm to the previous `aes-256-cbc` default);
- [x] re-enable the `poirot-rails` gem (we'll remove it in a dedicated PR);
- [x] re-enable the `aws-sdk` gem (or just `aws-sdk-s3`);
- [x] replace `therubyracer` gem with `mini_racer` (#1791);
- [x] push compatible branch to [instedd/ruby-location_service](https://github.com/instedd/ruby-location_service) and use it;
  - [x] merge https://github.com/instedd/ruby-location_service/pull/1
  - [x] update `Gemfile` to use the merged commit
- [x] bump loofah to 2.19.1 (#1827)
- [x] bump rails-html-sanitizer to 1.4.4 (#1829)
- [x] **Regression**: investigate slow integration tests
- [x] ~~build the `instedd/nginx-rails:2.4` Docker image~~ rework production Dockerfile to be based on `ruby:2.4` instead of `instedd/nginx-rails`.
- [x] **BUG**: Sample label PDF pages are bleeding (the font is a little too big)c closes #1723

refs #1792


diegoliberman:
Let's wait until January to merge this, so that we send it to production when Julien is back from vacations.

ysbaddaden:
Hum, the integration tests on CI seem to be _much_ slower all of a sudden in this branch (`main` is unaffected). I'm investigating to see what's happening.

ysbaddaden:
Selecting options is what's taking the most time in both Capybara & Cucumber specs. Each time it must select an option in a form, the browser hangs for a while, and I see the following message in the geckodriver log. It eventually continues but that inserts many hangs in the automated process that slows everything down (the rest feels normally fast).

```
Marionette      WARN    TimedPromise timed out after 500 ms
```

It may have something to do with the CdxSelect React component.

ysbaddaden:
Further reduced: the hangup occurs when we click the `.Select-placeholder` selector in `CdxSelect#set` (and other methods). Finding the element is quick, but either the click is slow or the action of displaying the options (`.Select-option`) is super slow.

ysbaddaden:
Found it: [`CdxSelect#select_elem`](https://github.com/instedd/cdx/blob/b7119087d10503b6cbf00f9a13214f3658dc060b/features/support/page_objects/cdx_select.rb#L41-L50) is the source of the issue. We're searching for a parent element until `.Select` matches, which is taking over 2 seconds :fearful:

ysbaddaden:
I added an optimization to avoid blindly looking for the element when it's already the element we expect (it matches `.Select`). It fixed the slow issue that appeared with the upgrades (Ruby & dependencies).

There are still lots of cases where the fallback to the slow path is required (which I tried to optimize a bit). Reviewing the matchers in Page Objects to directly attach CdxSelect to `.Select` elements may speed up the integration tests further.

ysbaddaden:
@matiasgarciaisaia I reworked the production `Dockerfile` to be based off `ruby:2.4` directly instead of `instedd/nginx-rails:2.4`.

The changes for deployments are:

- change proxy target port is now 3000 (instead of 80)
- add the `RAILS_SERVE_STATIC_FILES=1` environment variable

This is already done in staging.

Dockerfile
```diff
@@ -17,41 +17,44 @@ RUN curl -L https://github.com/wkhtmltopdf/packaging/releases/download/0.12.6-1/
...
 # Configure NNDD
 RUN /app/docker/config-nndd
 
 # Set permissions for tmp and log directories
 RUN mkdir -p /app/tmp /app/log && chown -R app:app /app/tmp /app/log
 
-# Add config files
-ADD docker/web-run /etc/service/web/run
+EXPOSE 3000
+CMD ["/app/docker/web-run"]
```
ysbaddaden, line 60: The command actually starts `whenever` and `puma` processes. We might want to have a distinct container for `whenever`, like we do for `sidekiq`.

Alternative: use `sidekiq-cron` or `sidekiq-scheduler` instead.

Dockerfile-dev
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.3
+FROM ruby:2.4
```
ysbaddaden, line 1: Yeah, we're moving to Debian Buster (`oldstable`) :muscle:

ysbaddaden:
**BUG**: the PDF is slightly different than on `main`. For example the CDx logo bleeds to other pages.

Generating the PDF relies on the abandoned and far outdated `wkhtmltopdf` binary through the `wicked_pdf` gem. A weird "webkit shrinking" mode is enabled by default and happens to behave differently depending on the `wkhtmltopdf` builds, despite using the exact same HTML source file.

After trying, unsuccessfully, to get a similar rendering, I chose to try out [Prawn](https://prawnpdf.org) which proved as simple to use as ever and allowed to tweak the render to reach almost exactly the same render when zoomed at 400% (important because of dot matrix printer).

PR #1796, opened 2022-11-16
Defend against spam

omelao:
In my opinion the best approach is set the AWS WAF Bot control. But....as we need Mati to implement this, I installed Google Recaptcha V2 on Register form. I tried recpatcha v3 with the same gem, but it doesn't work. `recaptcha_v3` returns `undefined method`.

To test it you must go to `/users/registration/sign_up` page.


omelao:
I tried the last comment of [this](https://github.com/ambethia/recaptcha/issues/296) also....and it doesn't installed the gem.

docker-dev.env
```diff
@@ -10,3 +10,7 @@ SECRET_KEY_BASE=4f13a3e80905baa2e8e35489ebfae43d265cf8ca92bcac80e1ed0d2265874765
 REDIS_URL=redis://redis:6379
 PUMA_OPTIONS=
 PUBLIC_REGISTRATION=true
+RECAPTCHA_SITE_KEY=6LdCLQMjAAAAAOyRvpHlZ7hgHE8JHDdshNYms8zo
+RECAPTCHA_SECRET_KEY=6LdCLQMjAAAAAOy0aoCWAzTs0mo-0PYw9nd4kwyS
+RECAPTCHA_SITE_KEY_V3=6LeLkv0iAAAAAO51qtrpPv1UVVhZYYIsaBqyl6JQ
+RECAPTCHA_SECRET_KEY_V3=6LeLkv0iAAAAAAZi7HvBkxPW-BDcigTDJQtdj0iV
```
ysbaddaden, line 16: Security Rule: NEVER commit credentials, it's unsafe even for private repositories.

Don't forget to invalidate and regenerate these keys.

```suggestion
```

config/settings.yml
```diff
@@ -18,3 +18,6 @@ alert_sms_account_name: "CDx"
 alert_sms_application_name: "CDx-Dev"
 alert_sms_application_password: "cdx123cdx"
 alert_sms_from_tel: "sms://442393162302"
+recaptcha:
+  site_key: <%= ENV['RECAPTCHA_SITE_KEY'] %>
+  secret_key: <%= ENV['RECAPTCHA_SECRET_KEY'] %>
```
ysbaddaden, line 23: Let's add the ability to disable the feature. It should be opt-in (i.e. disabled by default).

```suggestion
recaptcha:
  enabled: <%= ENV["RECAPTCHA_ENABLED"] == "1" %>
  site_key: <%= ENV["RECAPTCHA_SITE_KEY"] %>
  secret_key: <%= ENV["RECAPTCHA_SECRET_KEY"] %>
```

Of course wherever we use recaptcha we should wrap it with `if Settings.recaptcha.enabled`.

config/initializers/recaptcha.rb
```diff
@@ -0,0 +1,4 @@
+Recaptcha.configure do |config|
+  config.site_key  = ENV['RECAPTCHA_SITE_KEY']
+  config.secret_key = ENV['RECAPTCHA_SECRET_KEY']
```
ysbaddaden, line 3: Let's use the settings

```suggestion
  config.site_key = Settings.recaptcha.site_key
  config.secret_key = Settings.recaptcha.secret_key
```

ysbaddaden (changes requested):
Let's make the feature opt-in, so it's disabled by default and we won't need to deal with it in development and tests but also in production when it's not needed.

Then it shall be ready to merge.

ysbaddaden (approved):

PR #1798, opened 2022-11-16
Box Label UI same as the Sample

omelao:
image

### DIFF
--- a/app/views/samples/_barcode_card.haml
+++ b/app/views/samples/_barcode_card.haml
@@ -19,7 +19,7 @@
     .code
       %strong
         Production Date:
-      #{@sample_form.date_produced.try(:strftime, ("%m/%d/%Y"))}
+      #{@sample_form.date_produced.try{ strftime("%m/%d/%Y")}}
     .logo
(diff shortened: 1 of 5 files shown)


app/views/samples/_barcode_card.haml
```diff
@@ -19,7 +19,7 @@
     .code
       %strong
         Production Date:
-      #{@sample_form.date_produced}
+      #{@sample_form.date_produced.try(:strftime, ("%m/%d/%Y"))}
```
ysbaddaden, line 22: You might want to use the block version:
```suggestion
      #{@sample_form.date_produced.try { strftime("%m/%d/%Y") }
```

ysbaddaden (commented):
Can you revert the changes to `db/schema.rb`? The pull request didn't change anything, so we should skip any changes.

By-the-way: I have the same differences, on a fresh database, so we can have a distinct pull request to fix that.

omelao:
@ysbaddaden fixed

ysbaddaden (approved):

PR #1801, opened 2022-11-17
Ensure correct inventory ordering

### LINKED ISSUE #1800: Inventory ordering is broken when the sample concentration is manually modified
1. Create an unblinded box
2. Find 1 of the samples and modify its concentration to whatever
3. Download the inventory and you'll see that the ordering isn't updated, the sample remains at the position it used to be with the previous concentration.

Reminder: unblinded inventory is ordered by batch, concentration and replicate number.

### Test cases for Blinded boxes
It should order by sample UUID

- [ ] LOD box
- [ ] Variants box
- [ ] Challenge box

### Test cases for Unblinded boxes
It should order by batch, concentration, replicate

- [ ] LOD box
- [ ] Variants box
- [ ] Challenge box

### Bonus
Let's take the opportunity to place the columns in a more logic order.  `Concentration` should be to the left of `Replicate`

### Screenshots
image

The challenge shows that the ordering by batch is also broken
image

leandroradusky:
Closes #1800.

Inventory ordering is now ensured. The problem was that the `sort_by` method was called in an incorrect manner, and the sort was made using `nil` values.

Debug info will help to understand, this is within the context of the line 106 of `load_box_samples`: 
```
>> samples[0]
=> #<Sample id: 1479, sensitive_data: "\x04\xC8\x84\xAD\x9BF\x97\xED#\xD27\xCD\xDE\x01\xE2\x9A", institution_id: 12, custom_fields: {}, core_fields: {"date_produced"=>2009-09-09 00:00:00 +0000, "lab_technician"=>"123", "specimen_role"=>"b", "isolate_name"=>"zzzx", "inactivation_method"=>"Formaldehyde", "concentration"=>"102e1", "replicate"=>"1", "media"=>"Culture media"}, created_at: "2022-11-17 09:07:22", updated_at: "2022-11-17 09:07:41", patient_id: nil, encounter_id: nil, is_phantom: true, deleted_at: nil, batch_id: 18, isolate_name: "zzzx", site_id: nil, site_prefix: nil, specimen_role: "b", old_batch_number: nil, qc_info_id: nil, box_id: 39>
>> samples[0][:concentration]
=> nil
>> samples[0].concentration
=> 1020
```

The spec checking correct ordering was modified accordingly.

 **Bonus:** As the issue specify, the `concentration` column of the inventory was moved next to the replicate, this required as well to modify a spec.

### DIFF
--- a/app/controllers/boxes_controller.rb
+++ b/app/controllers/boxes_controller.rb
@@ -103,7 +103,7 @@ class BoxesController < ApplicationController
 
   def load_box_samples
     samples = @box.samples.preload(:batch, :sample_identifiers)
-    samples = if @box.blinded?
+    samples = if @box.blinded? && !params[:unblind] 
       samples.sort_by{ |sample|  [ sample.uuid ] }
     else
       samples.sort_by{ |sample|  [ sample.batch_number , sample.concentration , sample.replicate ] }
--- a/app/controllers/boxes_controller.rb
+++ b/app/controllers/boxes_controller.rb
@@ -104,7 +104,7 @@ class BoxesController < ApplicationController
   def load_box_samples
     samples = @box.samples.preload(:batch, :sample_identifiers)
     samples = if @box.blinded? && !params[:unblind] 
-      samples.sort_by{ |sample|  [ sample.uuid ] }
+      samples.scrambled
     else
       samples.sort_by{ |sample|  [ sample.batch_number , sample.concentration , sample.replicate ] }
     end 
--- a/app/controllers/boxes_controller.rb
+++ b/app/controllers/boxes_controller.rb
@@ -103,7 +103,11 @@ class BoxesController < ApplicationController
 
   def load_box_samples
     samples = @box.samples.preload(:batch, :sample_identifiers)
-    samples = samples.scrambled if @box.blinded?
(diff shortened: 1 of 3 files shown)


app/controllers/boxes_controller.rb
```diff
@@ -103,7 +103,11 @@ def load_box
 
   def load_box_samples
     samples = @box.samples.preload(:batch, :sample_identifiers)
-    samples = samples.scrambled if @box.blinded?
+    samples = if @box.blinded? && !params[:unblind] 
+      samples.sort_by{ |sample|  [ sample.uuid ] }
+    else
+      samples.sort_by{ |sample|  [ sample.batch_number , sample.concentration , sample.replicate ] }
+    end
```
ysbaddaden, line 110: It's better to order in SQL rather than having to do it in Ruby, so let's keep the `scrambled` scope, and fallback to `#sort_by` because of they're entity fields.

```suggestion
    samples = if @box.blinded? && !params[:unblind] 
      samples.scrambled
    else
      samples.sort_by{ |sample|  [ sample.batch_number , sample.concentration , sample.replicate ] }
    end 
```

ysbaddaden (approved):

PR #1803, opened 2022-11-17
Pattern to validate input of concentration field

### LINKED ISSUE #1795: Validate the input of the concentration field in the Sample
When entering values such as `1e+30`, there is a probability that the user enters an incorrect value.  I have tested entering `ijadsijdsidsj` and the sample was saved without errors.  When coming back to the sample it would show the concentration with a value of zero.

Let's at least validate that the allowed values are:
1. Any number
2. The letter `e` and the letter `E`
3. The plus sign `+`
4. The decimal separator `.`

If the user inputs any other character, we should show an error message informing an example of an acceptable value in the scientific notation.

We should also validate that:
1. There is only 1 consonant.
2. The consonant is followed by the plus sign
3. No negatives

leandroradusky:
Closes #1795 

Added a pattern to `concentration` field to allow only for positive numbers allowing scientific notation or raw numbers.

image

### DIFF
--- a/app/views/samples/_form.haml
+++ b/app/views/samples/_form.haml
@@ -65,7 +65,7 @@
       = f.form_field :concentration do
         - @sample_form.blinded_attribute(:concentration) do
           .row.input-unit
-            = f.text_field :concentration, min:0, readonly: !@can_update, :class=> "input-medium", :value => humanize_concentration(@sample_form.concentration), :pattern => "[\\d.]+(?:e-?\\d+)?"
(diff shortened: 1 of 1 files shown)


diegoliberman:
@leandroradusky can we show an example of a valid scientific notation?  Perhaps the error message could be: `Please use a valid scientific format, ie: 3.14e+24`

leandroradusky:
To avoid defining javascript custom messages and keep everything the most HTML as possible (which I leart from Julien is a good practice <3 )  I've defined the `title` of the field, which will look like this. Let me know if this is not good.

image

ysbaddaden (approved):

PR #1804, opened 2022-11-18
Add (Un)Blind Samples Button on Box Detail

omelao:
Just for Owners/Creators, if a box is transfered it will not show the button.

image

image

### DIFF
--- a/app/controllers/boxes_controller.rb
+++ b/app/controllers/boxes_controller.rb
@@ -17,7 +17,7 @@ class BoxesController < ApplicationController
   def show
     return unless authorize_resource(@box, READ_BOX)
     @can_delete = has_access?(@box, DELETE_BOX)
-    @can_blind = @box.box_transfers.count == 0
+    @can_blind = !@box.transferred?
 
     @samples = load_box_samples
   end
@@ -93,7 +93,7 @@ class BoxesController < ApplicationController
   end
 
   def unblind
-    return unless authorize_resource(@box, UPDATE_BOX)
+    return head :forbidden if @box.transferred?
     @box.unblind!
 
     redirect_to box_path(@box), notice: "Samples were successfully unblinded."
--- a/app/controllers/boxes_controller.rb
+++ b/app/controllers/boxes_controller.rb
@@ -100,8 +100,7 @@ class BoxesController < ApplicationController
   end
 
   def blind
-    return unless authorize_resource(@box, UPDATE_BOX)
-
+    return head :forbidden if @box.transferred?
     @box.blind!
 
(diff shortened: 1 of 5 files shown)


app/models/box.rb
```diff
@@ -94,4 +94,12 @@ def blind_attribute?(attr_name)
   def transferred?
     box_transfers.exists?
   end
+
+  def unblind!
+    self.update(blinded: false)
```
ysbaddaden, line 99: `self` is implicit in Ruby, you don't have to specify it (same in `#blind!`). You may also want to use [#update_columns](https://api.rubyonrails.org/v6.1/classes/ActiveRecord/Persistence.html#method-i-update_columns) which would be a bit faster (no need to re-validate the box, we only change one value):

```suggestion
    update_columns(blinded: false)
```

Same for `#blind!`

app/controllers/boxes_controller.rb
```diff
@@ -17,6 +17,7 @@ def index
   def show
     return unless authorize_resource(@box, READ_BOX)
     @can_delete = has_access?(@box, DELETE_BOX)
+    @can_blind = @box.box_transfers.count == 0
```
ysbaddaden, line 20: We have a method for that (it's a little faster as it won't even run a COUNT in SQL):

```suggestion
    @can_blind = !@box.transferred?
```

app/controllers/boxes_controller.rb
```diff
@@ -91,6 +92,21 @@ def bulk_destroy
     redirect_to boxes_path, notice: "Boxes were successfully deleted."
   end
 
+  def unblind
+    return unless authorize_resource(@box, UPDATE_BOX)
```
ysbaddaden, line 96: Sadly, this won't do. Once the box is added to a transfer we won't show the blind/unblind button, but once the transfer is confirmed, the receiver institution will own the box, and thus be authorized to blind/unblind by forging the request.

A basic `return head :forbidden if @box.transferred?` should work for now.

ysbaddaden (changes requested):
Nice! Just a few readability enhancements, but most importantly the authorization to improve.

ysbaddaden:
@omelao Can I bother you to add controller specs for the `boxes#blind` and `boxes#unblind` actions, to ensure that they work as expected?

- context: before transfer
  - owner institution can blind box
  - owner institution can unblind box

- context: after transfer
  - owner institution can't blind box
  - owner institution can't unblind box
  - sender institution can unblind box after transfer

Those specs are important to ensure that we don't inadvertently break the authorizations later!

omelao:
Yeah. Sure.

omelao:
@ysbaddaden done

PR #1818, opened 2022-12-07
Fix for concentration migration

### LINKED ISSUE #1817: Bug when migrating the concentration
This concentration:
image

Was migrated to `1000` when the correct value is `10`.

This concentration
image

Was migrated to `10330000` when the correct value is `10.33`.

### Game plan for Demo
Consider that the data has already been migrated there.  I think we don't need to worry about this.  I don't care what's the final concentration for those 2 test cases I created. 

### Game plan for Local
Consider we're dealing with data and we need to be careful. You should setup a local environment with the current production version, create a few samples, and then upgrade + run the migrations.  And check that the concentrations were correctly transformed.  What I don't know, is what happens with the current buggy migration: do we have to create a 2nd migration that fixes the error of the first one? Do we create a new migration that replaces the buggy one? 

### Game plan for Production
We'll have to talk with the team. Do we add the fix into the current release? Do we create a new release?

leandroradusky:
Closes #1817 

Bug fix in concentrations migrations from two fields(`concentration_number` & `concentration_exponent`) into one field (`concentration`). Sign of the exponent was badly expressed as positive in the existant migration.

### DIFF
--- a/db/migrate/20221109105200_change_sample_concentrations_to_single_value.rb
+++ b/db/migrate/20221109105200_change_sample_concentrations_to_single_value.rb
@@ -5,7 +5,7 @@ class ChangeSampleConcentrationsToSingleValue < ActiveRecord::Migration
       exponent = sample.core_fields["concentration_exponent"].try(&:to_i)
 
       if number && exponent
-        sample.core_fields["concentration"] = number * (10**exponent)
+        sample.core_fields["concentration"] = number * (10**-exponent)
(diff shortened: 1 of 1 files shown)


ysbaddaden (approved):

diegoliberman:
This PR should be merged and included in a new released based on https://github.com/instedd/cdx/releases/tag/0.21.5.

PR #1819, opened 2022-12-07
Truncating institution name on box label

### LINKED ISSUE #1618: Allow to upload measured concentrations with a CSV file
In the samples menu, add a new option named: Measurement Reports. For the scope of this ticket, this new page will only have an option for uploading the CSV file with the measured concentrations.

The CSV will have 3 columns: `<sampleId>, <measuredConcentration>, <measurementResult>`.

For each one of the sample Ids, we will store the received values in new attributes of the Sample entity.  These 2 fields can be added in the Sample detail form, but in read mode only.

### Considerations
1. If the sample already contains a value for these fields, they shouldn't be overwritten, and an error feedback should be informed to the user.
2. If the sample doesn't exist, an error should be informed to the user.

### invision
1. After uploading, we show the measured concentrations / results as a column.  See the [invision](https://projects.invisionapp.com/d/main#/console/4103589/470303882/preview).
2. [Errors](https://projects.invisionapp.com/d/main#/console/4103589/470303881/preview)

### LINKED ISSUE #1816: Box label is broken in the PDF
In order to make the content fit within the label, let's do the following: keep the QR and the ID.  Below that:

`<Institution name>` (without label, truncate to 22 characters)
`Purpose: xxxx (nn samples)` 

Example:
University of Californ...
Purpose: LOD (28 samples)

image

leandroradusky:
Closes #1816 

Institution name is now truncated for institution name to avoid a new line that broke the label.

Before:

image

Now:

image

Given that an ellipsis is used for truncating, there are only 6 characters displayed when the institution name exceeds the 9 characters, which is relatively short, maybe `Institution` can be shortened? (cc @diegoliberman)

### DIFF
--- a/app/views/boxes/barcode.pdf.haml
+++ b/app/views/boxes/barcode.pdf.haml
@@ -8,7 +8,7 @@
     .code
       Box: #{box.samples.count} samples
     .code
-      Institution: #{box.institution.name}
+      Institution: #{(box.institution.name).truncate(9)}
     .code
       Purpose: #{box.purpose}
     .logo
--- a/app/views/boxes/barcode.pdf.haml
+++ b/app/views/boxes/barcode.pdf.haml
@@ -6,11 +6,9 @@
       .uuid
         #{box.uuid}
     .code
-      Box: #{box.samples.count} samples
+      #{(box.institution.name).truncate(22)}
     .code
-      Institution: #{(box.institution.name).truncate(9)}
-    .code
-      Purpose: #{box.purpose}
+      #{box.purpose} (#{box.samples.count} samples)
     .logo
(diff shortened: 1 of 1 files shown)


ysbaddaden:
@leandroradusky please read the issue again, it says "without label, truncate to 22 characters" :)

diegoliberman:
This PR should be merged and included in a new release based on https://github.com/instedd/cdx/releases/tag/0.21.5

leandroradusky:
Sorry for the bad reading on the original issue, now is properly displayed.
Also, after talking with Diego, we eliminated the `Purpose:` label and display the purpose and quantity of samples in the second line as follows:

image

ysbaddaden (approved):

PR #1824, opened 2022-12-12
GitHub Codespaces

ysbaddaden:
Creates the basic devcontainer configuration for GitHub Codespaces.

It starts a devcontainer (required by codespaces) with docker in docker and compose support. Once started it pulls and builds the docker images required by CDx, and runs a few commands to setup the dev environment (i.e. ruby dependencies, db setup).

Once the codespace is started, you can run `docker compose up web` on a VScode terminal, and port 3000 should automatically be exposed.

TODO: we should use a [browserless image](https://hub.docker.com/r/browserless/chrome) that exposes a webpage with a VNC server (or something like that) instead of the custom `selenium` images, that is only useful for CI / running system tests quickly, but bad for writing tests (you can't see what's happening). For local dev, we redirect to a geckodriver/chromedriver running on the host, but we can't do that in a codespace, which runs remotely.

NOTE: VScode web is _very_ sluggish (at least in Firefox). Be it the code editor, the Git or Terminal panes, everything is slow. It probably fares better when running in a local VScode desktop with the GitHub codespaces extension.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,12 +1,28 @@
 {
-  "dockerComposeFile": "../docker-compose.yml",
-  "service": "web",
-  "runServices": ["web", "db", "elasticsearch", "redis", "sidekiq", "selenium"],
-
-  "settings": {
-    "terminal.integrated.shell.linux": "/bin/bash"
-  },
-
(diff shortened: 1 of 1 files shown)


ysbaddaden:
Let's merge so we can try it.

PR #1828, opened 2022-12-14
Box reports for LOD boxes

### LINKED ISSUE #1820: LOD visualization

leandroradusky:
Closes #1820 .

It looks like this:

(1/2)
image
(2/2)
image

There are some differences with respect to the [mockup](https://projects.invisionapp.com/d/main#/console/4103589/470974659/preview):

- Measurement data points are shown for the LOD plot. 
- No threshold is needed for the LOD boxes
- Computed LOD is not displayed following the university requirements
- Box purpose is indicated in the summary
- If there are samples without measurements they are informed (if measurements are uploaded for those samples, the report will update when re-displaying)

image

- Confusion matrix labels for totals were corrected to display the data they are totalizing.
 
**Code comments**

- New React components were created to display the plots. The `BoxReportsBarChart` was based on the existing `BarChart` adding flexibility to display specific variables stored within the dictionary `data` parameter. The `BarChart` react component wasn't modified to don't affect existing reports' stacking philosophy (even if they aren't used).
- Despite the last item, colors in CSS were just updated on the existing styles to reuse the existing code and respect the mockup, this could only have an impact on old reports' coloring.
- The Confusion Matrix is a view, its data is collected with a helper method.


app/controllers/samples_reports_controller.rb
```diff
@@ -0,0 +1,158 @@
...
+      redirect_to samples_reports_path, notice: 'Box report was successfully created.'
+    else
+      render action: 'new'
+    end
+  end
+
+  def show
+    @samples_report = SamplesReport.find_by_id(params[:id])
+    @reports_data = measured_signal_data(@samples_report)
+    @samples_without_results_count = @samples_report.samples.where("core_fields NOT LIKE '%measured_signal%'").count
```
leandroradusky, line 77: Is this OK?

ysbaddaden: We might be able to. We should be able to use the following:

```ruby
.where('core_fields->>"$.measured_signals" IS NOT NULL')
```

To be efficient we'd have to create an ugly [virtual generated column to create an index](https://dev.mysql.com/doc/refman/5.7/en/create-table-secondary-indexes.html#json-column-indirect-index), but given how many samples there are a sequential scan might not be an issue.

ysbaddaden: I said :poop: the core fields are... encoded in YAML :sob:

PR #1831, opened 2022-12-16
Automatic unblinding for samples with uploaded measurements

### LINKED ISSUE #1735: Automatic unblinding
1. When the measured concentrations are uploaded (#1618), all blinded data for the samples need to be automatically unblinded.
2. The manual action for unblinding, that is enabled to the sender in the transfer log detail, needs to be removed.

leandroradusky:
Closes #1735 .

Sample data is unblinded when `measured_signal` field is defined. Also, is shown as read-only in the sample's form when it is displayed.

image

With respect to the issue specification: unblind action was not removed since it applies for boxes and this implementation applies for individual samples. **Small consistency issue: theoretically, if almost all results are uploaded for samples of a box, the rest could be "inferred" if the contents of how the box are made are known, while I don't think this is really an issue to take care of, I mention it**. (cc @diegoliberman).

### DIFF
--- a/app/presenters/sample_presenter.rb
+++ b/app/presenters/sample_presenter.rb
@@ -33,7 +33,7 @@ class SamplePresenter
 
   def blinded_attribute?(attr_name)
     if box = @sample.box
-      box.blinded? && !!box.blind_attribute?(attr_name)
+      (box.blinded? && !!box.blind_attribute?(attr_name)) && !@sample.measured_signal
     else
       false
     end
--- a/app/views/samples/_form.haml
+++ b/app/views/samples/_form.haml
@@ -82,6 +82,10 @@
         - else
           = f.text_field :media, readonly: true
 
+      - if @sample_form.measured_signal
+        = f.form_field :measured_signal do
+          = f.text_field :measured_signal, readonly: true
+
   .col
     = render (@can_update ? 'form_assays' : 'show_assays'), f: f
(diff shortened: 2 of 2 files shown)


diegoliberman:
@leandroradusky which fields are shown as read-only after the results are uploaded?

leandroradusky:
> @leandroradusky which fields are shown as read-only after the results are uploaded?

The read-only "forced" applies (now) just for the `measured_signal`, the others' "read-onlyness" are linked to the property `can_update`... I should check the transferred boxes have this property `can_update` set to false, no?

Other detail @diegoliberman: this will unblind as well blinded samples within the inventory: 
- this is ok?
- should we add `measured_signal` in the inventory if applies?

diegoliberman:
1. Unblinding applies everywhere, it isn't defined per element.  So it's fine to unblind the inventory as well.
2. We don't need to add the `measured_signal` in the inventory.
3. OK with `measured_signal` field read-only.  I think we might change our mind in the future, but it's ok for now.

leandroradusky:
Then, if the PR is approved we're good to merge :)

ysbaddaden (approved):
Looks good, as long as the `Sample#measured_signal` attribute exists on `main`?

sardar-usman:
@leandroradusky It's unblinding the sample data. Working as expected.

Issue #1808, opened 2022-11-23
Show custom confirm on sample upload


leandroradusky:

sardar-usman:
@leandroradusky it's working as expected.

PR #1834, opened 2023-01-03
Feature/1807 custom confirm modal

### LINKED ISSUE #1807: Create custom confirm component
Create a new component to display confirmation custom messages like the ones defined [in this mockup](https://projects.invisionapp.com/d/main#/console/4103589/470842165/preview) or [in this one](https://projects.invisionapp.com/d/main#/console/4103589/470792646/preview). 

The existing component `ConfirmationModal` serves as a good template but bases all it's logic on the actions of confirming and confirming deletion to display its title and apply the corresponding styling, doesn't allow nor to display custom messages on the buttons neither define actions for them, which are needed by the reporting feature requirements.

leandroradusky:
Closes #1807 & #1808. 

When a file containing invalid IDs is uploaded a modal to confirm the upload is shown as in the [mockup](https://projects.invisionapp.com/d/main#/console/4103589/470842165/preview):

image

Also, a warning message is shown in a modal to confirm the measurements uploading, as in [this mockup](https://projects.invisionapp.com/d/main#/console/4103589/470792646/preview)

image

Some notes:
- An extra warning before uploading was added with respect to the message in the mockup.
- A unique class to color the contents of the modal was added (`red`) and this same class modify the title and the text of the modal. Maybe is better to add two separate props to the modal? The name was chosen to make explicit that modifying other than the color will make the modal look not ok since the title and the content will apply the same class.
- There were introduced two props to the `ConfirmModal` class to call callback functions and use this component as a regular modal and not just as a confirm submission one (thanks Julien <3). 
- The `confirmation_modal.js.jsx` file was really old and I removed the tabs to indent with spaces as in the rest of the project.

### DIFF
--- a/app/assets/javascripts/components/confirmation_modal.js.jsx
+++ b/app/assets/javascripts/components/confirmation_modal.js.jsx
@@ -51,8 +51,8 @@ var ConfirmationModal = React.createClass({
     return this.props.deletion ? "btn-primary btn-danger" : "btn-primary";
   },
   
-  titleClass: function() {
-    return this.props.titleClass ? this.props.titleClass : "";
+  colorClass: function() {
+    return this.props.colorClass ? this.props.colorClass : "";
   },
   
   showCancelButton: function() {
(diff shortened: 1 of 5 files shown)


ysbaddaden (approved):
Oh my, the JS on this project is really :see_no_evil: :hear_no_evil: :speak_no_evil: 

I didn't get the `window[this.props.funcName]()` at first but that's because the modal is a react component, whereas the upload results is jQuery injected into the HTML.

Oh, well.

leandroradusky:
Yeah!! That's what has made me struggle with how to do it!! And then after talking with you I realized this was the way... But is a "sophisticated" ─to use a nice word :)─ wiring of the functionalities

PR #1841, opened 2023-01-16
Drop `poirot_rails` gem

### LINKED ISSUE #1234: Disable poirot_rails in Staging
The poirot_rails gem prints hundreds of logs.  In Staging, it's difficulting debugging efforts, and it doesn't provide any value.  We should disable this gem from the Staging environment.

ysbaddaden:
closes #1234


matiasgarciaisaia (approved):
LGTM!

Does this bring the logs back to non-JSON form? I didn't see where that's written.

ysbaddaden:
@matiasgarciaisaia That's a good question! I guess it returns logs to be regular Rails logs, hence very verbose but readable? I guess we'll see that on staging.

We'll may want to try [lograge](https://github.com/roidrage/lograge/) for even nicer, yet semantic, logs.

PR #1844, opened 2023-01-18
Box reports merge into main

### LINKED ISSUE #1734: Allow to create reports for the measured concentrations
Create an entity for the Samples\Measurement Reports page. 

When creating a report, the user will be able to:
1. Define a name for the report
1. Select a list of boxes or a list of samples
1. Define a threshold (numeric, float)
1. Save the report
1. See the 3 graphics

The system will use the measured concentrations (fields added in #1618) for the selected samples/boxes and it will show 3 graphics:

1. True positives and negatives (confusion matrix): Specificity, Sensitivity
5. LOD
6. ROC curve

### Note
- Samples with `null` values in the measured concentrations need to be ignored for the purpose of these graphics. 
- Add an informative message saying the number of samples that are being considered for the graphic, and the number of samples that were ignored.

### Measurement reports - Columns & sorting
Sort by Creation date, descending. 

1. Creation date
2. Name
3. Threshold

### Measurement reports - Filters
Allow to filter by Name, partial matching.

### Acceptance criteria
1. I can see the list of created reports with the most recents at the top
2. I can create new reports
3. I can modify exiting reports
4. I can delete exiting reports
5. I can see the graphics generated for each report
6. I can filter reports by name

leandroradusky:
Closes #1734.

Reports feature merged into main. Some comments & small fixes that were pending:

- When samples results are uploaded a `notice` informing it is displayed
- The `bulk_delete` button is enabled now only when `sample_reports` are selected in the index view.
- In the two branches (boxes v2 and reports) the styles to display samples in batches/in boxes/list of uploaded files, etc grow separately as `samples-row*` and `items*`, I have merged them choosing the second option (and checked all the places where they are used.
- The `batches_selector` component had a bug: `distractor` and `instructions` were defined in the samples when they were added into the batch, meaning that if the user uploaded first the information of the `replicates`/`copies` and after the `distractor`/`instructions` this info was lost. I have changed it to be added to the samples at the moment of adding a new batch into the component.

This is the merge of the whole feature within main, so maybe I should point out the specific places of the code I'm referring to in these small comments, given that the rest was already reviewed, please let me know Julien if you want that I do it.


ysbaddaden:
> maybe I should point out the specific places of the code I'm referring

Yes, you can review/comment your own pull requests. I do it all the time when I need to say something related to context.

app/helpers/samples_reports_helper.rb
```diff
@@ -0,0 +1,52 @@
+module SamplesReportsHelper
+  def get_rates(samples, signal)
+    confusion_matrix = confusion_matrix(samples, signal)
+
+    if confusion_matrix[:actual_positive] > 0 && confusion_matrix[:actual_negative] > 0 
+      tpr=confusion_matrix[:true_positive].to_f/confusion_matrix[:actual_positive].to_f
+      fpr =confusion_matrix[:false_positive].to_f/confusion_matrix[:actual_negative].to_f
+      [fpr, tpr]
+    else
+      [0,0]
```
ysbaddaden, line 10: This is returning an Array<Integer> while the other branch returns an Array<Float>, let's always return floats?
```suggestion
      [0.0, 0.0]
```

ysbaddaden (approved):
One tiny comment :+1:

PR #1849, opened 2023-01-19
Change samples source empty selector component in box creation

### LINKED ISSUE #1778: Changing the sources empties the contents
If the samples source is change to `Create new samples` or `Select from inventory`, the system shows a warning message and after confirming, the contents are emptied.

https://projects.invisionapp.com/d/#/console/4103589/470324049/comments/149800275

leandroradusky:
Closes #1778.

When the source of samples in box creation is changed, a warning message is shown:

image

To accomplish this, react selector components reset to empty state by clicking a hidden `<a>`. It's the best way I found to apply changes within React components from plain javascript. Not sure it's the best way. 

The javascript is commented because it has some complexity (should I put it in a separate file? don't know if it's a rule to do it or not). At least, some jquery is gone :)

### DIFF
--- a/app/assets/javascripts/components/batches_selector.js.jsx
+++ b/app/assets/javascripts/components/batches_selector.js.jsx
@@ -10,7 +10,7 @@ var BatchesSelector = React.createClass({
       list: [],
     };
   },
-  reset : function() {
+  reset: function() {
     this.setState({
       batches: [],
       samples: this.props.samples,
@@ -23,7 +23,6 @@ var BatchesSelector = React.createClass({
   },
   render: function () {
     let button;
-    let clearButton = <a className="clear-batches" href="#" onClick={this.reset}></a>;
     if (this.state.batches.length > 0) {
       button = '';
     } else {
@@ -36,7 +35,7 @@ var BatchesSelector = React.createClass({
     }
 
     return (<div className="batches-selector">
-      {clearButton}
+      <a className="clear-batches" href="#" onClick={this.reset}></a>
       <div className="items-count">
         <div className="title">{this.state.list.length}&nbsp;{this.state.list.length == 1 ? "batch" : "batches"}</div>
       </div>
--- a/app/assets/javascripts/components/batches_selector.js.jsx
+++ b/app/assets/javascripts/components/batches_selector.js.jsx
@@ -10,9 +10,20 @@ var BatchesSelector = React.createClass({
       list: [],
     };
   },
-
+  reset : function() {
+    this.setState({
+      batches: [],
+      samples: this.props.samples,
+      concentration: null,
+      replicate: null,
+      distractor: null,
+      instruction: null,
+      list: [],
+    });
+  },
   render: function () {
     let button;
+    let clearButton = <a className="clear-batches" href="#" onClick={this.reset}></a>;
     if (this.state.batches.length > 0) {
       button = '';
     } else {
@@ -25,6 +36,7 @@ var BatchesSelector = React.createClass({
     }
 
     return (<div className="batches-selector">
+      {clearButton}
       <div className="items-count">
         <div className="title">{this.state.list.length}&nbsp;{this.state.list.length == 1 ? "batch" : "batches"}</div>
       </div>
--- a/app/assets/javascripts/components/samples_selector.js.jsx
+++ b/app/assets/javascripts/components/samples_selector.js.jsx
@@ -4,14 +4,12 @@ var SamplesSelector = React.createClass({
       samples: this.props.samples,
     };
   },
-  reset : function() {
+  reset: function() {
     this.setState({ samples: [] });
   },
   render: function () {
-    let clearButton = <a className="clear-samples" href="#" onClick={this.reset}></a>;
-    
     return (<div className="samples-selector">
-      {clearButton}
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/components/samples_selector.js.jsx
```diff
@@ -4,9 +4,14 @@ var SamplesSelector = React.createClass({
       samples: this.props.samples,
     };
   },
-
+  reset : function() {
```
ysbaddaden, line 7:
```suggestion
  reset: function() {
```

app/assets/javascripts/components/batches_selector.js.jsx
```diff
@@ -10,9 +10,20 @@ var BatchesSelector = React.createClass({
       list: [],
     };
   },
-
+  reset : function() {
```
ysbaddaden, line 13:
```suggestion
  reset: function() {
```

app/assets/javascripts/components/batches_selector.js.jsx
```diff
@@ -10,9 +10,20 @@ var BatchesSelector = React.createClass({
...
+      concentration: null,
+      replicate: null,
+      distractor: null,
+      instruction: null,
+      list: [],
+    });
+  },
   render: function () {
     let button;
+    let clearButton = <a className="clear-batches" href="#" onClick={this.reset}></a>;
```
ysbaddaden, line 26: I think you can just put the `<a>` directly into the template below.

app/assets/javascripts/components/samples_selector.js.jsx
```diff
@@ -4,9 +4,14 @@ var SamplesSelector = React.createClass({
...
+    this.setState({ samples: [] });
+  },
   render: function () {
+    let clearButton = <a className="clear-samples" href="#" onClick={this.reset}></a>;
```
ysbaddaden, line 11: Same here.

app/views/boxes/_form.haml
```diff
@@ -43,3 +43,64 @@
...
+  var cancelSampleSourceChange = function() {}
+
+  var sourcesRadio = document.querySelectorAll(".radiotoggle input");
+  var oldSource = null;
+
+  var applySampleSourceChange = function (displayField) {
+    // Hide all the fields and disable them
+    sourcesRadio.forEach(function(otherRadio) {
+      document.querySelector(`#${otherRadio.value}`).style.display = 'none';
+      document.querySelector(`#${otherRadio.value}`).setAttribute("disabled", true);
```
ysbaddaden, line 70: Maybe?
```suggestion
      document.getElementById(otherRadio.value).style.display = 'none';
      document.getElementById(otherRadio.value).setAttribute("disabled", true);
```

That would also apply below.

app/views/boxes/_form.haml
```diff
@@ -43,3 +43,64 @@
...
+
+      if (oldSource) {
+        // If the user changed the source, show the modal
+        document.querySelector(`.sample-source-change-modal-container`).style.visibility = 'visible';
+      } else {
+        // For the first time, simulate that the user clicked on the confirm button
+        confirmSampleSourceChange();
+      }
+    });
+  });
```
ysbaddaden, line 105: I think we can use global functions. We can take advantage that the functions will only ever be invoked _after_ showing the confirm modal (or faking a click), so we don't need them to be noop by default.

```js
function confirmSampleSourceChange() {
  applySampleSourceChange(this);
  // ...
}

function cancelSampleSourceChange() {
  // ...
}

sourcesRadio.forEach(function(radio) {
  radio.addEventListener("change", function (evt) {
    if (oldSource) {
      // ..
    } else {
      // here is the trick to have a specific `this`:
      confirmSampleSourceChange.call(this)
    }
  });
});
```

ysbaddaden (commented):
Nice! I think we can improve the JS glue a bit :)

app/views/boxes/_form.haml
leandroradusky: This won't work as it is: remember that `confirmSampleSourceChange` is also passed to the custom confirm modal as a function to be called when the confirm button is pressed, and there the `this` parameter will be the window. I need to redefine the `confirmSampleSourceChange` function (as a var) each time the `change` event is triggered.

leandroradusky: This won't work as it is: remember that `confirmSampleSourceChange` is also passed to the custom confirm modal as a function to be called when the confirm button is pressed, and there the `this` parameter will be the window. I need to redefine the `confirmSampleSourceChange` function (as a var) each time the `change` event is triggered.

ysbaddaden: Oh right :thinking: Then what about the following?

```js
var confirmSampleSourceChange;
function confirmSampleSourceChangeFn() {}
function cancelSampleSourceChange() {}

sourcesRadio.forEach(function(radio) {
  radio.addEventListener("change", function (evt) {
    if (oldSource) {
      confirmSampleSourceChange = confirmSampleSourceChangeFn.bind(this)
      // ...
    } else {
      confirmSampleSourceChangeFn.call(this)
    }
  });
});
```

We could also have `showConfirmModal()` and `hideConfirmModal()` helpers to improve readability.

ysbaddaden (commented):
LGTM

ysbaddaden (approved):

PR #1861, opened 2023-01-24
Fix: rename SamplesReport as BoxReport

### LINKED ISSUE #1860: Rename SamplesReport resource as BoxReport
The displayed named for the feature is "Box Report" but the internal resource name is SamplesReport, this is leading to a dichotomy (external & internal names are different) and is a source of confusion when editing roles, because the box report policies are for the SamplesReport entity.

Let's settle on a single name: BoxReport.

ysbaddaden:
fixes #1860


leandroradusky:
Hi  @ysbaddaden,

The PR looks good, but I think we should wait for the approval of @diegoliberman to move the merge forward: there was mentioned by the university the possibility of making reports of individual samples (i.e. composing samples coming from different boxes), that's why I named the entity like that. 

This is also true "productwise": even if not implemented now, it could be good to have the ability to make reports for whatever samples we want. Some examples: make a report comprising the results of all the boxes sent to an awardee, or one specific technology/machine, or made of individual samples sent separately, etc. 

That's why i.e. I limited the `box_selector` to one box, but everything should work if we allow the box selector allows to add `n` boxes. For individual samples, with minor logic tweaks using a `samples_selector` should allow us to create these reports I'm mentioning.

diegoliberman:
I prefer to think it as a report of samples. The box is just a facilitator for filtering the samples included in the report. But Julien is making a good point. Perhaps we can change the UI titles? Something like "Samples reports from boxes" could work? The menu entry within samples could be just "Reporting". What do you think?

Issue #1858, opened 2023-01-23
Showing wrong count for "Boxes".


sardar-usman:
**Steps to reproduce:**

1. Go to the Boxes page.
2. Showing the wrong count for created boxes.

**Expected Result**
Boxes count should be equal to created boxes.
 

<img width="1493" alt="Screenshot 2023-01-23 at 4 47 09 PM" src="https://user-images.githubusercontent.com/45853972/214033134-11bc82af-ae0e-4731-9d42-b4faf47cf1ad.png">

**My setup**

- Device: Laptop
- OS: macOS
- Browser: Chrome

ysbaddaden:
I just reproduce the issue on my local. I'm investigating.

My first impression is that the `Boxes.count_samples` is dropping a box... maybe it skips boxes without samples?

ysbaddaden:
Yeah, `count_samples` does an INNER JOIN on samples, but what if a box doesn't have any sample?

The root cause may actually be because of #1859 that allows to create to a box without samples.

ysbaddaden:
**QA information**: if staging had boxes without samples, then the bug will appear as not fixed: the counter will always count them but they won't show up. We only fixed the cause for the bug (being able to create boxes without samples).

The proper fix would be to delete those boxes from the database, but there are only a few in staging, and it's probably not worth the hassle.

sardar-usman:
@ysbaddaden I would be better if you remove the boxes from database for staging.

ysbaddaden:
@sardar-usman Alright, I just did the cleanup.

sardar-usman:
@ysbaddaden It is working fine now.

PR #1867, opened 2023-01-24
Improve UX of batch selection in boxes#new

### LINKED ISSUE #1845: Keep uploaded batches information if SamplesReport creation throw error
Example: When creating a Challenge box where, minimally, we need to have a virus and a distractor batch. If we don't include them, then we`re losing all the uploaded batches/samples/concentrations info and I have to go with creating new samples from scratch:

image

Keeping the information of the uploaded batches will avoid having to redefine them.

### LINKED ISSUE #1846: Allow to reopen batch details when creating a box
Now we cannot "reopen details" for defined batches to check if we made a mistake when defining them. 
In the picture below, all the definitions for batch 12 remain hidden and we cannot see them. Let's suppose that we don't recall if we made it a distractor or not, or, that we forget to make it a distractor and we want to change that: once we press "Add" there is no step back nor to change neither to check... Given that boxes will be quite sophisticated can be improved:

image

### LINKED ISSUE #1859: Should not allow to add Batch with empty "Contcentrations" .
1. Go to 'Boxes' page.
2. Filled the Fields.
3. Check the "Create new samples" radio button.
4. Click on 'Add Batch' button.
5. Enter Batch Id, it opens the screen to add batch.
6. Click on Add button with empty Concentrations fields.
7. It will add the Batch.

https://user-images.githubusercontent.com/45853972/214074117-fac18921-c868-4cad-a4d2-17ab676e1e4d.mov

**Expected**

It should not allow to add Batch  with empty Concentrations fields.

**My setup**

- Device: Laptop
- OS: macOS
- Browser: Chrome

ysbaddaden:
Overall refactor of the React component and submitted params. BoxBatchesForm is roughly equivalent to BatchesSelector, but with an extracted BoxBatchForm sub-component and a somewhat simplified params, the component generates inputs to the form that is eventually sent by the browser (the forms are always rendered, but hidden with `display:none`).

No need to click the "ADD BATCH" or "ADD SAMPLE" buttons after choosing to create new samples or selecting. You can immediately go to select your first batch or sample.

No need to manually validate each concentration anymore: just enter valid values, otherwise the OK button is disabled. I removed the CANCEL button as the behavior was confusing to implement. it was currently remove batch, when it feels closer to undo (which didn't felt trivial to implement).

Capture d’écran de 2023-01-26 23-25-33@2x

The backend correctly verifies & reports errors for each purpose (verified with integration tests). On error, the form is re-populated. This works for both batches & samples. It should no longer be possible to create a box without samples, hence avoiding #1858. This is validated by integration tests.

Capture d’écran de 2023-01-26 23-26-51@2x

Last but not least, I added a "down chevron" to allow reopening a batch.

Capture d’écran de 2023-01-26 23-39-21@2x

closes #1845
closes #1846
closes #1859

Potential improvements:
- [ ] mark the inputs as red with a tooltip when the replicate/concentration field is invalid (on input blur only).
- [ ] validate the whole form and disable the SAVE button when something is invalid. Not easy because we're mixing vanilla JS with react components! Maybe we need a BoxForm component.


ysbaddaden:
I found another bug: the form is re-populated on error, but the batch UUID isn't properly set, so the batch is overlooked :sob: 

I also broke some boxes controller test units.

spec/controllers/boxes_controller_spec.rb
```diff
@@ -349,63 +350,66 @@ def expect_samples(batch, concentrations: [1], replicates:)
...
-        expect(box.samples.count).to eq(6)
+        expect(box.samples.count).to eq(7)
         expect(box.samples.where(batch: virus_batch).count).to eq(2)
         expect(box.samples.where(batch: virus_batch).map(&:distractor).uniq).to eq([false])
         expect(box.samples.where(batch: virus_batch).map(&:concentration).uniq).to eq([1])
-        expect(box.samples.where(batch: distractor_batch).count).to eq(4)
+        expect(box.samples.where(batch: distractor_batch).count).to eq(5)
         expect(box.samples.where(batch: distractor_batch).map(&:distractor).uniq).to eq([true])
-        expect(box.samples.where(batch: distractor_batch).map(&:concentration).uniq.sort).to eq([40000,400000])
-        expect(box.blinded).to eq true
```
leandroradusky, line 396: check is blinded doesn't hurt :)

spec/controllers/boxes_controller_spec.rb
```diff
@@ -349,63 +350,66 @@ def expect_samples(batch, concentrations: [1], replicates:)
...
     end
 
     describe "with existing samples" do
       it "creates with samples" do
         samples = Sample.make! 2, :filled, institution: institution, specimen_role: "b"
 
         post :create, params: { box: box_plan.merge(
-                                    media: "Saliva", 
+                                    media: "Saliva",
                                     blinded: true,
```
leandroradusky, line 411: Already in the `box_plan`

leandroradusky:
Just brillant!!

I've pulled and tested and it's working neat.

The only important thing to add IMO, and I think this wasn't before the refactor Julien so it's not your fault, is the same validations for the samples creation than for the batches creation: 

- Variant boxes created from samples should contain samples coming at least from 2 different batches.
- Challenge boxes created from samples should at least contain one distractor and one non-distractor sample. 
- LOD & other are good with what it's there. 

I'm going through the code in detail and it looks great, let me take another look before approving ─the PR is quite long and some stuff is new to me, I'm learning while reading some cool stuff :)─

app/assets/javascripts/components/box_batches_form.js.jsx
```diff
@@ -0,0 +1,360 @@
...
+
+  selectBatch: function (_, options) {
+    if (options.length == 0) return;
+
+    var batches = this.state.batches;
+    var genRandomKey = this.__genRandomKey;
+
+    var newBatches = options.map(function (option) {
+      return {
+        key: genRandomKey(batches),
```
ysbaddaden, line 90: I can't use the batch UUID, because the current form allows to set the same batch twice (or more). One valid reason is that there could be different instructions. I thus use a randomly generated key to make sure that React will correctly diff/patch/update the DOM.

app/models/box_form.rb
```diff
@@ -17,10 +19,11 @@ def self.build(navigation_context, params = {})
 
   def initialize(box, params)
     @box = box
-    @option = params[:samples].presence
-    @concentrations = params[:concentrations].presence
+    @params = params
```
ysbaddaden, line 22:
```suggestion
```

leandroradusky (approved):
I went over the code and functionality and it looks good to me :)

PR #1871, opened 2023-01-26
Don't allow to create Box Report without sample with results

### LINKED ISSUE #1870: Should not allow to create Report with empty result.
1. Go to 'Boxes' page.
2. Filled the Fields.
3. Check the "Select samples from inventory" radio button.
4. Click on 'Add Sample' button.
5. Enter Sample Id, and save the box.
6. Go to "Box Report" page.
7. Generate the Report and it will create empty results report.

**Expected**

It should not allow to create report with empty results.

**My setup**

- Device: Laptop
- OS: macOS
- Browser: Chrome
Screenshot 2023-01-26 at 5 54 45 PM

leandroradusky:
Closes #1870.

Don't allow creating box reports if there isn't at least one sample with results on it. 
Is a fix to a validation implemented incorrectly, without it, plots won't work ok.

### DIFF
--- a/app/models/samples_report.rb
+++ b/app/models/samples_report.rb
@@ -46,8 +46,8 @@ class SamplesReport < ApplicationRecord
   private
 
   def there_are_samples
-    samples_with_results = samples_report_samples.select { |srs| !srs.sample.measured_signal.nil? }
-    errors.add(:base, "Please select a box containing samples with results") if samples_with_results.empty?
+    samples_with_measurements = samples_report_samples.select { |srs| !srs.sample.measured_signal.nil? }
+    errors.add(:base, "The selected box should contain samples with uploaded measurements") if samples_with_measurements.empty?
   end
 
 end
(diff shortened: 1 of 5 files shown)


app/views/samples_reports/_form.haml
```diff
@@ -11,7 +11,7 @@
           maxBoxes: 1,
           caller: "samples_reports",
           boxes: boxes_data(@boxes))
-        = f.field_errors :"samples_report_samples "
+        = f.field_errors :"samples_report_samples"
```
ysbaddaden, line 14: That would be enough :)
```suggestion
        = f.field_errors :samples_report_samples
```

app/models/samples_report.rb
```diff
@@ -46,7 +46,8 @@ def self.entity_scope
   private
 
   def there_are_samples
-    errors.add(:base, "Please select a box containing samples with results") if samples_report_samples.empty?
+    samples_with_results = samples_report_samples.select { |srs| !srs.sample.measured_signal.nil? }
+    errors.add(:base, "Please select a box containing samples with results") if samples_with_results.empty?
```
ysbaddaden, line 50: Maybe explain that we need samples with measurements?

ysbaddaden (approved):

PR #1875, opened 2023-01-27
Ignoring blank lines when uploading sample results csvs

### LINKED ISSUE #1874: Error occurs while uploading CSV Results.
1. Go to 'Samples' page.
2. Click on Upload Results button.
3. On Upload Results page click on "Add File" button.
4. Upload the CSV file of created samples from the system.
5. Click on "Upload Result" button.
6. A pop message appeared, Click on Continue button.
7. It throughs an error.

**Expected**

It should allow to upload samples result successfully without any error.

https://user-images.githubusercontent.com/45853972/215056545-6ab92bd2-0234-40f3-bd6f-c1ab0805ac20.mov

**My setup**

- Device: Laptop
- OS: macOS
- Browser: Chrome

leandroradusky:
Close #1874 

When the uploaded CSV for sample results have blank lines, they produce an error since there are no fields to be matched with the `uuid` regex. Now, if a blank line is present it is ignored.

### DIFF
--- a/app/controllers/samples_controller.rb
+++ b/app/controllers/samples_controller.rb
@@ -208,7 +208,7 @@ class SamplesController < ApplicationController
       CSV.open(csv_file.path) do |csv_stream|
         csv_stream.each do |row|
           sample_id, measured_signal = row[0], row[1]
-          next if !sample_id.match(uuid_regex) # non uuids are ignored
+          next if !sample_id || !sample_id.match(uuid_regex) # non uuids are ignored
           sample = Sample.find_by_uuid(sample_id)
           unless sample.nil?
             sample.measured_signal ||= Float(measured_signal) if measured_signal.present?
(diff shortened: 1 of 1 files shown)


app/controllers/samples_controller.rb
```diff
@@ -208,7 +208,7 @@ def bulk_process_csv
       CSV.open(csv_file.path) do |csv_stream|
         csv_stream.each do |row|
           sample_id, measured_signal = row[0], row[1]
-          next if !sample_id.match(uuid_regex) # non uuids are ignored
+          next if !sample_id || !sample_id.match(uuid_regex) # non uuids are ignored
```
ysbaddaden, line 211: FYI, the following should do (`&.` is the `lonely operator`, similar to `?.` in JS):

```suggestion
          next unless sample_id&.match(uuid_regex) # non uuids are ignored
```

And below could be (Crystal idiom):

```ruby
if sample = Sample.find_by_uuid(sample_id)
  # ...
end
```

In Ruby (and Crystal) only `nil` and `false` can ever be falsy values, everything else is truthy (even `0`, `""`, ...), so we usually don't have to explicitly test for `.nil?`. In Crystal terms this is a nilable (`Sample | Nil` union) and a simple `if` is enough to skip nils!

ysbaddaden (commented):
This is fine to merge as-is. I just left a comment for your learning.

PR #1876, opened 2023-01-27
Speed up login for feature specs

ysbaddaden:
I got bored with how slow each integration is to start, and spend half a hour to implement a dirty token authentication for users, only available to the test environment.

Instead of loading the login page, fill the form, submit, be redirected to the dashboard, just to eventually open _another page_... it now authenticates in a single request, returning a mere 200 OK with a blank page.

Tests now start almost instantly, and the integration test suite is now ~25% faster on my local!

### DIFF
--- a/spec/support/feature_spec_helpers.rb
+++ b/spec/support/feature_spec_helpers.rb
@@ -68,7 +68,7 @@ module FeatureSpecHelpers
 
   def goto_page(klass, args = {})
     page = klass.new
-    load_page_with_retries(page, args)
+    load_page_with_retries(page, args, attempts: 3)
     yield page if block_given?
   end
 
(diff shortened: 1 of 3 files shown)


diegoliberman:
@leandroradusky do you feel comfortable reviewing this PR?

ysbaddaden:
Rebased from main.

spec/support/feature_spec_helpers.rb
```diff
@@ -65,6 +83,22 @@ def expect_page(klass)
   def snapshot
     screenshot_and_open_image
   end
+
+  private
+
+  def load_page_with_retries(page, args, attempts: 0)
```
leandroradusky, line 89: Attempts as name & being equal to `0` is misleading, one expects that it indicates the number of attempts to be made. Also, is anti-intuitive if someone call the method with `attempts=2` which will lead to `3-2` attempts.

leandroradusky:
I've tested it locally and it works great :)

Except for a small comment, looks good to me.

leandroradusky (approved):

spec/support/feature_spec_helpers.rb
ysbaddaden: Oh my, it's definitely confusing. I changed it to be the number of attempts!

PR #1887, opened 2023-02-01
Fixed deleted confirmation message for boxes

### LINKED ISSUE #1881: Wrong "Delete Confirmation" message for Boxes.
1. Go to 'Boxes' page.
2. Click on any Box from the list.
3. Click on "Delete Box button"
4. A pop up screen is appeared.
5. Wrong message for deleting the box.

**Expected**

It should show the message for deleting the Box while for now it showing the message for batches..

**My setup**

- Device: Laptop
- OS: macOS
- Browser: Chrome
Screenshot 2023-01-30 at 10 02 38 PM

leandroradusky:
Closes #1881.

Now when deleting boxes the message shown is correct

image

To avoid future copy/paste errors I rely on the `controller_name`, and changed it accordingly in the other views (not in `SamplesReports` being a special case).

### DIFF
--- a/app/views/batches/index.haml
+++ b/app/views/batches/index.haml
@@ -11,7 +11,7 @@
       =form_tag bulk_action_batches_path, :id => 'batches_form' , :method => :post do
         = cdx_table title: pluralize(@total, "batch") do |t|
           - t.actions do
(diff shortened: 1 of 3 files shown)


ysbaddaden (commented):
Hum, I still prefer verbatim text. It just happens to work with English, but that will cause issues if we ever restart localizing strings.

ysbaddaden (approved):

PR #1888, opened 2023-02-01
Fix: remove concentrations button in box form

### LINKED ISSUE #1885: Concentration's field data should not be removed.
1. Go to 'Boxes' page.
2. Filled the Fields.
3. Check the "Create new Samples" radio button.
4. Click on 'Add Sample' button.
5. Enter Batch Id and open the batch.
6. Filled the ACCOMPLISHED CONCENTRATIONS fields.
7. Add the second Concentration and did not filled the fields.
8. Now removed the second  Concentration, it automatically remove the data of first  Concentration.

**Expected**

It should not Remove the Data of first concentration.

**My setup**

- Device: Laptop
- OS: macOS
- Browser: Chrome

https://user-images.githubusercontent.com/45853972/216059837-4a8c3c2c-dfd2-45be-be84-a23dcd455a5a.mov

ysbaddaden:
We can't use the array index to identify as a react key when the array is dynamic.

fixes #1885

### DIFF
--- a/app/assets/javascripts/random_key.js
+++ b/app/assets/javascripts/random_key.js
@@ -1,0 +1,19 @@
+function cdxGenRandomKey(items) {
+  var keys;
+  var key;
+
+  if (items) {
+    keys = items.map(function (item) { return item.key });
+  } else {
+    keys = [];
+  }
+
+  do {
+    key = "xxxxxxxxxxxxxxxx".replace(/[x]/g, function (c) {
+      const r = Math.floor(Math.random() * 16);
+      return r.toString(16);
+    });
+  } while (keys.indexOf(key) != -1);
+
+  return key;
+}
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/random_key.js
```diff
@@ -0,0 +1,19 @@
+function cdxGenRandomKey(items) {
+  var keys;
+  var key;
+
+  if (items) {
+    keys = items.map(function (item) { return item.key });
+  } else {
+    keys = [];
+  }
```
leandroradusky, line 9: This would doit in one line but I'm not sure if it's preferred. 

```
(items||[]).map(function (item) { return item.key });
```

leandroradusky (approved):

app/assets/javascripts/random_key.js
ysbaddaden: Thanks, this is much better!

PR #1894, opened 2023-02-03
Feature/1877 validate samples when creating boxes

### LINKED ISSUE #1877: Validate samples when creating boxes
When creating a box from selected samples, we should still do some validation based on the purpose:

1. Variant boxes created from samples should contain samples coming at least from 2 different batches;
2. Challenge boxes created from samples should at least contain one distractor and one non-distractor sample.
3. LOD and Other boxes created from samples don't need to more validation.

Follow up to #1867

leandroradusky:
Closes #1877.

Before the PR itself, two small fixes: 
- In Samples form, both `distractor` and `instruction` properties were not saved if modified because they were not included in the `samples_controller` as attributes.
- In `boxes_controller`, on the `add_samples` scenario, the samples could not belong to any batch, making the `sort_by` for the samples list to fail. 

For the PR: 
- Added validations for the box as described in the issue.
- Added new specs to validate this behavior.

### DIFF
--- a/app/controllers/boxes_controller.rb
+++ b/app/controllers/boxes_controller.rb
@@ -113,7 +113,7 @@ class BoxesController < ApplicationController
     samples = if @box.blinded? && !params[:unblind]
       samples.scrambled
     else
-      samples.sort_by{ |sample|  [ sample.batch_number , sample.concentration , sample.replicate ] }
+      samples.sort_by{ |sample|  [ sample.batch_number ? sample.batch_number : "" , sample.concentration , sample.replicate ] }
     end
     SamplePresenter.map(samples, request.format, unblind: params[:unblind])
   end
--- a/spec/support/blueprints.rb
+++ b/spec/support/blueprints.rb
@@ -177,6 +177,7 @@ Sample.blueprint(:filled) do
   specimen_role { "p" }
   date_produced { Faker::Time.backward }
   measured_signal { 10.0 }
+  batch { object.encounter.try(:batch) || Batch.make! }
 end
 
 Sample.blueprint(:batch) do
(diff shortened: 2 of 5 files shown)


leandroradusky:
Hi! I need help on this (cc @ysbaddaden):

Tests will fail in the three commits that I've made, but they don't fail to me locally and I don't understand why. I'm getting the message:

```
expected to find text "You must select at least one non-distractor sample" in "You must select at least one sample"
```

While I'm adding samples to the boxes. Since this doesn't fail in my local I'm lost even on how can I do to debug it. Any help is welcome!

app/controllers/boxes_controller.rb
```diff
@@ -113,7 +113,7 @@ def load_box_samples
     samples = if @box.blinded? && !params[:unblind]
       samples.scrambled
     else
-      samples.sort_by{ |sample|  [ sample.batch_number , sample.concentration , sample.replicate ] }
+      samples.sort_by{ |sample|  [ sample.batch_number ? sample.batch_number : "" , sample.concentration , sample.replicate ] }
```
ysbaddaden, line 116:
```suggestion
      samples.sort_by{ |sample|  [ sample.batch_number.to_s, sample.concentration , sample.replicate ] }
```

spec/support/blueprints.rb
```diff
@@ -177,6 +177,7 @@ def first_or_make_site_unless_manufacturer(institution)
   specimen_role { "p" }
   date_produced { Faker::Time.backward }
   measured_signal { 10.0 }
+  batch { object.encounter.try(:batch) || Batch.make! }
```
ysbaddaden, line 180: This is wrong: a sample doesn't need to come from a batch. You should manually associate the sample to a batch in your tests (or use the `:batch` trait below). For example:

`Sample.make!(:filled, batch: Batch.make)`

ysbaddaden (approved):

PR #1895, opened 2023-02-03
Warning about duplicated uuids on sample results upload

### LINKED ISSUE #1891: "Upload results informing wrong not-found-uuid count if repeated samples in csv file"

leandroradusky:
Closes #1891.

Upon uploading results files for samples, if there were duplicated uuids, there was a wrong behavior on the count of found and not found uuids.
As was discussed in the text of the issue, now, if there are invalid or duplicated sample uuids, this modal is shown:

image

And then, inform invalid uuids (nonexistent within CDx user), the warning is shown just as before:

image

But for valid duplicated samples, the system now just ignores its second appearance, as in this file that contains twice the same sample:

image

### DIFF
--- a/app/views/samples/_upload_results_js.haml
+++ b/app/views/samples/_upload_results_js.haml
@@ -58,10 +58,11 @@
                 var samples = [] // samples uuids
                 var duplicates = false
                 lines.forEach(function(line) { 
-                    if (line != '' && isUUID(line.split(',')[0])){
-                        if (samples.indexOf(line.split(',')[0]) == -1){
-                            samples_param.push( 'uuids[]=' + line.split(',')[0] );
-                            samples.push( line.split(',')[0] );
+                    var row = line.split(',');
+                    if (line != '' && isUUID(row[0])){
+                        if (samples.indexOf(row[0]) == -1){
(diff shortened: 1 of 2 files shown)


app/views/samples/_upload_results_js.haml
```diff
@@ -56,11 +56,16 @@
...
                 var samples = [] // samples uuids
-
+                var duplicates = false
                 lines.forEach(function(line) { 
                     if (line != '' && isUUID(line.split(',')[0])){
-                        samples_param.push( 'uuids[]=' + line.split(',')[0] );
-                        samples.push( line.split(',')[0] );
+                        if (samples.indexOf(line.split(',')[0]) == -1){
+                            samples_param.push( 'uuids[]=' + line.split(',')[0] );
+                            samples.push( line.split(',')[0] );
```
ysbaddaden, line 64: Maybe memoize `var row = line.split(',')` so we don't split & allocate the same array 4 times (starting from line 61)?

ysbaddaden (commented):
Sounds good for now. We might want to improve the handling of errors in a future refinement (I'll write an issue about it).

ysbaddaden (approved):

PR #1898, opened 2023-02-06
Fix to messages on box creation

### LINKED ISSUE #1877: Validate samples when creating boxes

leandroradusky:
Closes #1877 

Fixed error messages when creating batches.


ysbaddaden (commented):
@leandroradusky the integration specs will fail :)

ysbaddaden (approved):

PR #1903, opened 2023-02-07
Feature/1814 upload csv with new box

### LINKED ISSUE #1814: Upload a CSV with the contents of a new box

leandroradusky:
Closes #1814 

Now a box can be created from a CSV file. To agree with the excel files that they use in order to define box samples concentrations, they should look like this.

image

This is the file that you can download as a template file when you choose the new option in the boxes form:

image

If the file is correct, it will respect the mockups done for the sample's upload results page:

image

If, within the file, some rows have inexistent batches, they will be informed:

image

While this message will be shown if no samples have been parsed (because they are not, or because the file has a wrong file format, etc):

image

Comments on the code:
- The way the upload works is, within `box_form`, the file is parsed and the same parameters as if the user has chosen the option `Create new samples` options are fulfilled and saved (`initialize_csv_box` method)
-  The upload CSV is done with templates and plain JS, as for the upload samples results. Should be good to generalize, or, maybe better, create a React component that does this.
- On Capybara, only the creation of boxes is tested, while the functioning of the validations for this new `add_csv` option is tested on the controller (as previously discussed).

### DIFF
--- a/app/controllers/boxes_controller.rb
+++ b/app/controllers/boxes_controller.rb
@@ -133,14 +133,14 @@ class BoxesController < ApplicationController
 
   def box_params
     if Rails::VERSION::MAJOR == 5 && Rails::VERSION::MINOR == 0
-      params.require(:box).permit(:purpose, :media, :blinded, :option).tap do |allowed|
+      params.require(:box).permit(:purpose, :media, :blinded, :option, :csv_box).tap do |allowed|
         allowed[:batches] = params[:box][:batches].try(&:permit!)
         allowed[:sample_uuids] = params[:box][:sample_uuids].try(&:permit!)
         allowed[:concentrations] = params[:box][:concentrations].try(&:permit!)
       end
     else
       params.require(:box)
-        .permit(:purpose, :media, :blinded, :option, sample_uuids: [], batches: [
+        .permit(:purpose, :media, :blinded, :option, :csv_box, sample_uuids: [], batches: [
           :batch_uuid, :distractor, :instruction, concentrations: [:replicate, :concentration]
         ])
     end
--- a/app/models/box_form.rb
+++ b/app/models/box_form.rb
@@ -22,6 +22,8 @@ class BoxForm
     @option = params[:option]
     @media = params[:media].presence
     @batches_data = params[:batches].presence.to_h
+    @csv_box = params[:csv_box].presence
+    initialize_csv_box if @csv_box
     @batch_uuids = @batches_data.transform_values { |v| v[:batch_uuid] }
     @sample_uuids = params[:sample_uuids].presence.to_h
   end
@@ -71,7 +73,7 @@ class BoxForm
 
   def build_samples
     case @option
-    when "add_batches"
+    when "add_batches", "add_csv"
       @batches_data.each do |batch_key, b|
         next if b[:batch_uuid].blank?
 
@@ -139,6 +141,16 @@ class BoxForm
         @box.errors.add(:base, "A virus batch is required") unless have_virus_batch?
         @box.errors.add(:base, "You must select at least one distractor batch") unless have_distractor_batch?
       end
+    when "add_csv"
+      case @box.purpose
+      when "LOD", "Other"
+        @box.errors.add(:base, "You must include at least one sample") unless unique_batch_count >= 1
+      when "Variants"
+        @box.errors.add(:base, "You must include samples from at least two different batches") unless unique_batch_count >= 2
+      when "Challenge"
+        @box.errors.add(:base, "You must include at least one non-distractor sample") unless have_virus_batch?
+        @box.errors.add(:base, "You must include at least one distractor sample") unless have_distractor_batch?
+      end
     when "add_samples"
       if @samples.empty?
         @box.errors.add(:base, "You must select at least one sample")
@@ -184,5 +196,26 @@ class BoxForm
 
   def have_distractor_sample?
     @samples.any? { |_, sample| sample.distractor }
+  end
+
+  def initialize_csv_box
+    CSV.open(@csv_box.path) do |csv_stream|
+      i = 0
+      csv_stream.each do |row|
+        batch_number, concentration, distractor, instruction = row[0..3]
+        batch_uuid = Batch.find_by(batch_number: batch_number)&.uuid
+        next if batch_uuid.blank?
+        @batches_data[i] = {
+          batch_uuid: batch_uuid,
+          distractor: distractor.downcase == "yes",
+          instruction: instruction,
+          concentrations: {i: {
+            replicate: 1,
+            concentration: Integer(Float(concentration)),
+          }},
+        }
+        i += 1
+      end
+    end
   end
 end
--- a/app/views/boxes/_form.haml
+++ b/app/views/boxes/_form.haml
@@ -1,4 +1,4 @@
-= cdx_form_for(@box_form) do |f|
+= cdx_form_for(@box_form, html: { multipart: true }) do |f|
   .row
     .col
       .row.form-field
@@ -20,6 +20,8 @@
           = f.label :option, "Create new samples", value: "add_batches"
           = f.radio_button :option, "add_samples"
           = f.label :option, "Select samples from inventory", value: "add_samples"
+          = f.radio_button :option, "add_csv"
+          = f.label :option, "Create from CSV file", value: "add_csv"
 
       .row.form-field
         .col.form-field__label
@@ -33,7 +35,25 @@
             = react_component "SamplesSelector", { url: autocomplete_samples_path(format: "json", context: params[:context], qc: 0),
               name: "box[sample_uuids]", placeholder: "Enter sample id", className: "input-block",
               samples: @box_form.samples_data }
+          
+          %fieldset#add_csv{ :disabled => @box_form.option != "add_csv", :style => (@box_form.option == "add_csv" ? "" : "display: none") }
+            #uploaded-files
 
+            .row
+              = button_tag id: 'add-box-file', class: 'btn-link', type: "button" do
+                .icon-circle-plus.icon-blue.icon-margin
+                %span.btn-upload
+                  Add file
+            .row
+              =link_to '/templates/upload_box.csv', class: 'btn-link', target: "_blank" do
+                .icon-download.icon-gray
+                %span.btn-download
+                  CSV template
+            .row.errors.hidden#csv-file-error
+              No samples were found in the CSV file or it has an invalid format. 
+              %br
+              Please check the template and try again.
+              
   .row
     .col
       = f.check_box :blinded
@@ -54,6 +74,8 @@
                       title: "Warning",
                       message: "If you change the sample source, the box contents will be emptied. ")
 
+= render 'upload_csv_box'
+
 :javascript
   var sourcesRadio = document.querySelectorAll(".radiotoggle input");
   var oldSource = null;
@@ -73,6 +95,9 @@
     // Clear the selectors contents
     document.querySelector(".clear-batches").click()
     document.querySelector(".clear-samples").click()
+    // Clear the CSV file and re-enable the add button
+    document.getElementById("uploaded-files").innerHTML = "";
+    document.getElementById(`add-box-file`).style.display = 'block';
   }
 
   function showConfirmModal() {
--- a/app/views/boxes/_upload_csv_box.haml
+++ b/app/views/boxes/_upload_csv_box.haml
@@ -1,0 +1,146 @@
+
+%template{:id => "csvFileRow"}
+  .list-items.upload_info
+    .items-row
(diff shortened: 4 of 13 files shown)


leandroradusky:
I need further help with this (cc @ysbaddaden)

In Capybara, even if being able to attach the file (adding the `template` element that contains the file `input`) to the input element, it does not appear as a param in the box creation, these are the params on `create` in the controller of boxes:

`{"utf8"=>"✓", "box"=>{"purpose"=>"LOD", "media"=>"Culture media", "option"=>"add_csv", "blinded"=>"0"}, "commit"=>"Save", "context"=>"d90409a6-3997-c33a-958c-55a3e62b64a5", "controller"=>"boxes", "action"=>"create"}`

Some lead (that could be not the good one) to make the template to be included, I have to click the `Add file` button that will make the dialog appear, but I can't do nothing in capybara with this dialog. So maybe this is keeping the file input empty?

diegoliberman:
If this gets complicated, we'll leave it on a pause.

leandroradusky:
It's just the Capybara tests, besides that works well (and the controller its tested).

app/models/box_form.rb
```diff
@@ -185,4 +197,25 @@ def have_virus_sample?
...
+        batch_uuid = Batch.find_by(batch_number: batch_number)&.uuid
+        next if batch_uuid.blank?
+        @batches_data[i] = {
+          batch_uuid: batch_uuid,
+          distractor: distractor.downcase == "yes",
+          instruction: instruction,
+          concentrations: {i: {
+            replicate: 1,
+            concentration: Integer(Float(concentration)),
+          }},
```
ysbaddaden, line 215: Note that you're actually creating a `{ :i => {} }` hash (symbol), when I think you mean `{ i => {} }`, but actually you could just go with `{ "0" => {} }`, since there will always be only one entry, unless you rework to group by batch UUID, which might be a good idea.

ysbaddaden:
@leandroradusky what about uploading the CSV & validate it on the server and let it render the HTML to inject into the page? That would help rid of most of the JS.

leandroradusky:
Hi @ysbaddaden (cc @diegoliberman)

IMO: there is _at least_ this way to uniquely identify batches on their mind: 
- batch number (which is not a number)
- the batch uuid
- the taxonomy id
- the who_label (really similar to the batch number, but not the same)
- the pango_lineage (really similar to the batch number, but not the same)

for me in this case is quite important to inform the batches which were not found to guide them (the actual validations don't do that)

Also, this file is made to be as similar as possible to the one that they use to compute the concentrations:
- which is an XLS (high change of uploading the bad format _even_ if its a CSV file)
- with more columns / different columns / different order)

ysbaddaden:
@leandroradusky I'm suggesting that we do all these validations but on the backend transparently (in the background, no page reload).

You select a CSV file to upload. JS will display a "validating (spinner)" message, upload the CSV to `/boxes/validate_csv` (for example), validate that the CSV is as expected, and eventually render the validation as the response body, that the JS can merely inject into the page.

You currently have to ask the backend if each batch exists, which results in N requests, but with the above approach we'd only have one request _and_ avoid most of the JS, while still doing all the validation (win-win-win).

leandroradusky:
Great! I didn't get it before, its a great hint thanks <3

diegoliberman:
@ysbaddaden @leandroradusky could it be possible to merge this one as is, and create a new issue for doing the validations in the backend?

ysbaddaden:
@diegoliberman Is the feature that important to have _now_? I'd prefer to avoid pushing tech-debt if possible.

app/views/boxes/_upload_csv_box.haml
```diff
@@ -0,0 +1,146 @@
...
+  // Upon pressing the "Add file" button we will create a file input and click it
+  document.getElementById(`add-box-file`).addEventListener('click', function (e) {
+    var template = document.getElementById(`csvInputFile`).content;
+    var fragment = template.cloneNode(true);
+    var fileInput = fragment.querySelector('.csv_file');
+    document.getElementById(`uploaded-files`).append(fragment)
+    
+    fileInput.addEventListener('change', handleFileSelect, false);
+    fileInput.click();
+  })
```
ysbaddaden, line 146: I understand that we can only select one CSV file to upload (not many)... what about rendering the file input directly into the main HTML and skip the `add file` button? We'd avoid this whole block of JS, the template, and could directly call `handleFileSelect`:

```haml
#uploaded-files
  = file_field_tag "box[csv_box]", class: "csv_file", accept: "text/csv", onchange: "handleFileSelect"
```

app/views/boxes/_upload_csv_box.haml
```diff
@@ -0,0 +1,146 @@
...
+          fragment.querySelector(`.file-name`).textContent = `${f.name}`;
+          
+          // Bind click event to remove button
+          var removeFileRow = function(evt) {
+              evt.currentTarget.upload_info.remove();
+              change_event.target.remove();
+              document.getElementById(`add-box-file`).style.display = 'block';
+          }
+          fragment.querySelector(".remove_file").upload_info = fragment.querySelector(`.upload_info`)
+          fragment.querySelector(".remove_file").addEventListener('click', removeFileRow, false)
```
ysbaddaden, line 108: I feel like we could make the `removeFileRow` global, especially if we always render the file input, we don't need `change_event.target.remove()`. We could then call it directly from the HTML:

```haml
.remove_file{onclick: "removeFileRow"}
```

I'm not sure we need to assign the `upload_info` attribute above. Since there are only one row, it could just:
```js
document.querySelector("#uploaded-files .upload-info").remove()
```

Same for `showToolTip` below.

ysbaddaden (commented):
We can probably live with the JS, but the `handleFileSelect` is too big and nests too many closures.

I believe we can skip the "add files" button and always render the upload CSV file input instead. We could then simplify a bunch of things in the JS, extract methods out of the `handleFileSelect` callback.

For example:
- extract the part that parses & validates the CSV into functions (pass it `load_event.target.result` and let it return `samples`, `found_batches` & `not_found_batches`);
- extract the rendering of the row template;
- make `removeFileRow` global (not closured);
- since we're only rendering one row: we can assume the DOM elements to be there, and don't need to pass references anymore;
- ...

app/models/box_form.rb
```diff
@@ -185,4 +197,25 @@ def have_virus_sample?
...
+
+  def initialize_csv_box
+    CSV.open(@csv_box.path) do |csv_stream|
+      i = 0
+      csv_stream.each do |row|
+        batch_number, concentration, distractor, instruction = row[0..3]
+        batch_uuid = Batch.find_by(batch_number: batch_number)&.uuid
+        next if batch_uuid.blank?
+        @batches_data[i] = {
+          batch_uuid: batch_uuid,
```
ysbaddaden, line 209: Maybe:
```suggestion
        batch = Batch.find_by(batch_number: batch_number)
        next unless batch

        @batches_data[i] = {
          batch_uuid: batch.uuid,
```

bolom:
Voila. I've move the validation in backend. Let me know what do you think @leandroradusky @ysbaddaden

leandroradusky:
Despite the integration tests to be fixed, I've tested the functionality locally to demo it with UCSD and it works as expected :)

app/controllers/csv_validations_controller.rb
```diff
@@ -0,0 +1,19 @@
+class CsvValidationsController < ApplicationController
+  def create
+    uploaded_file = params[:csv_file]
+    if uploaded_file.content_type == "text/csv"
+      csv_text = uploaded_file.read
+      institution = @navigation_context.institution
+      csv_table = CSV.parse(csv_text, headers: true)
+      batch_values = csv_table.map { |row| row["Batch"] }
+      found_batches = institution.batches.where(batch_number: batch_values).pluck(:batch_values)
```
ysbaddaden, line 9: Where does the `:batch_values` column come from?

app/views/boxes/_form.haml
```diff
@@ -35,6 +37,28 @@
               name: "box[sample_uuids]", placeholder: "Enter sample id", className: "input-block",
               samples: @box_form.samples_data }
 
+          %fieldset#add_csv{ :disabled => @box_form.option != "add_csv", :style => (@box_form.option == "add_csv" ? "" : "display: none") }
+            #uploaded-files
```
ysbaddaden, line 41: The other two cases (`add_batches` and `add_samples`) rely on a React component and the HTML & the whole jQuery-like JS really feels wrong.

Would you be courageous enough to try and refactor to such a basic React component? The two other components are quite basic and merely generate a FORM that will eventually be submitted.

Alternatively, if all we do is select a CSV file and upload it... maybe we could just have the user submit the form, we validate on the backend and render the form again on failure (with the expected error messages). That would avoid an endpoint to validate the CSV, all the custom JS, and feel much more like Rails.

app/controllers/csv_validations_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      not_found_batches = (batch_values - found_batches).uniq
+
+      render json: { found_batches: found_batches,
+                    not_found_batches: not_found_batches,
+                    batches_nbr: batch_values.size }
+    else
+      render json: { error: "Invalid file format. Please upload a CSV file." }, status: :unprocessable_entity
+    end
+  end
+end
```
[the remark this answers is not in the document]
bolom: Totally see where you're coming from! However, I'm a tad concerned about introducing non-CRUD methods and creating a hefty controller.

Here's a thought - how about we incorporate the CSV validation directly into the Boxes#create method? This way, we can dodge the double upload and still maintain a streamlined controller design. Does that sound good to you?

app/views/boxes/_form.haml
bolom: Sure, I really like your alternative idea! Shifting the CSV validation to the backend aligns nicely with my preferred style and it simplifies things. It's great that we can avoid having a specialized controller like `csv_validations_controller`. Let's go for it!

bolom:
hey @leandroradusky @ysbaddaden 
 I have removed the unnecessary JavaScript validation and related features from the codebase.

In addition, I've made the Box#create method more consistent and concise. I removed the redundant valid? check for BoxForm, as the BoxForm.save method already performs the necessary validations.

What do you think?

app/controllers/boxes_controller.rb
```diff
@@ -51,21 +51,16 @@ def new
 
   def create
     return unless authorize_resource(@navigation_context.institution, CREATE_INSTITUTION_BOX)
-    @box_form = BoxForm.build(@navigation_context, box_params)
 
+    @box_form = BoxForm.build(@navigation_context, box_params)
     @box_form.batches = check_access(load_batches, READ_BATCH)
-    @box_form.samples = check_access(load_samples, READ_SAMPLE)
```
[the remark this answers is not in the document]
bolom: I'm little confused, what I should I do here? Or you already fixed that in your last commit

app/controllers/boxes_controller.rb
```diff
@@ -51,21 +51,16 @@ def new
...
-    if @box_form.valid?
+    if @box_form.save
       @box_form.build_samples
-
-      if @box_form.save
-        redirect_to boxes_path, notice: "Box was successfully created."
-        return
-      end
+      redirect_to boxes_path, notice: "Box was successfully created."
+    else
```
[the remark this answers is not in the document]
bolom: > we save the box which will create a transaction and create the Box and its Samples in a transaction, so we can never have a Box without Samples (it caused bugs previously) or a Box with only some of the expected Samples (worst case).

I was thinking, we could move validation Box instead of BoxForm. What do you think?

```
class Box < ApplicationRecord
  has_many :samples
  validate :ensure_samples_validity

  private

  def ensure_samples_validity
    errors.add(:base, "Box must have at least one sample") if samples.empty?
    errors.add(:base, "You can't select a QC sample") if samples.any?(&:is_quality_control?)
  end
end


```

ysbaddaden:
@bolom yes, I fixed it ysterday as @leandroradusky wanted to have it today in staging.

app/controllers/boxes_controller.rb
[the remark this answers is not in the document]
ysbaddaden: Hum, having data integrity right into Box is tempting...

But validations are different depending on the selected option to create the box (there are too many ways to create a box).

- The "can't select QC sample" would be fine, since it can only happen when we manually select a sample.
- The "at least one sample message", on the other hand, would be wrong for the `add_batches` option (we only care about the selected batches), but also for the `add_csv` option (the wording is slightly different).

To discriminate, we'd need to push `@option` to Box, which would defeat the point of BoxForm (keep the creation logic from the HTML form distinct from the Box object).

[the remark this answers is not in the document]
bolom: make sense, thanks for the clarification

PR #1914, opened 2023-02-09
Fixes issues with boxes#new form

### LINKED ISSUE #1910: Server error when trying to create Box without a purpose
*Steps to reproduce:*

Try to create a box without selecting a purpose and the server will respond with an INTERNAL SERVER ERROR message.

*Culprits*

1. we don't display the error to the user alongside the form input (aka 'unhandled form error');
2. the mechanism to report this warning to sentry crashes, we thus end-up reporting the crash to sentry instead of the unhandled form parameters (oops).

Log:

> Unhandled form errors in Box: {:purpose=>["Purpose can't be blank", "Purpose is not included in the list"]}

Sentry issue [RADX-CDX-10](https://instedd.sentry.io/issues/3925071421/?project=6087729)

### LINKED ISSUE #1913: Can create box without selecting to create or select samples
Try to create a box, select a purpose but don't select an option to create new samples or select existing sample and submit the form. **The box will be created** and won't show up in the list of boxes yet be counted (triggering bug #1858).

We must validate than a valid option (`add_batches` or `add_samples`) was selected in BoxForm.

ysbaddaden:
closes #1910 
closes #1913

### DIFF
--- a/app/models/box.rb
+++ b/app/models/box.rb
@@ -30,7 +30,7 @@ class Box < ApplicationRecord
   attribute_field :purpose, copy: true
 
   validates_presence_of :purpose
-  validates_inclusion_of :purpose, in: ->(_) { Box.purposes }
+  validates_inclusion_of :purpose, in: ->(_) { Box.purposes }, allow_blank: true
   validates_associated :samples
 
   scope :count_samples, -> {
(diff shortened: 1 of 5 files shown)


app/models/box.rb
```diff
@@ -30,7 +30,7 @@ def self.entity_scope
   attribute_field :purpose, copy: true
 
   validates_presence_of :purpose
-  validates_inclusion_of :purpose, in: ->(_) { Box.purposes }
+  validates_inclusion_of :purpose, in: ->(_) { Box.purposes }, allow_blank: true
```
ysbaddaden, line 33: This is to avoid a duplicate error message when the purpose is missing:

- purpose can't be blank
- purpose is not included in the list

leandroradusky (approved):
LGTM :)

PR #1918, opened 2023-02-10
WIP: some advances to export PDF in firefox

### LINKED ISSUE #1900: Did not allowing to download report on Firefox browser.

leandroradusky:
Closes #1900 

The `html2pdf` library used to export PDF reports have different behaviors depending on the browser. The solution mentioned [here](https://github.com/niklasvh/html2canvas/issues/1578) was applied to create the function `onclone` which makes the report work.

Said this:
- The `{:hidden => true}` has to be removed from the parent div, which was used to hide the rendering of the report. If maintained, prevents the plots to be placed within the PDF file.
- Even if executed through a chain of `then` promises, the `window.close` closes the window before the PDF is saved.
- The `margin` property was removed from the `html2pdf` because it was not supported in firefox, margins are handled with CSSs.

### DIFF
--- a/app/assets/stylesheets/_samples_report.scss
+++ b/app/assets/stylesheets/_samples_report.scss
@@ -2,6 +2,10 @@
   @include flexbox();
   @include flex-direction(column);
   
+  .report-page {
+    margin:15px 15px 30px 30px;
+  }
+
   .report-summary-content {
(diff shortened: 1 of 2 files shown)


ysbaddaden:
Closing in favor to #1982

PR #1926, opened 2023-04-05
use bundle exec for rake command so it doesn't fail on a version mismatch

nthiad:
This is a very small patch to the instructions so that the db rake tasks succeed. When I tried this step, it failed with a rake version mismatch and the error message indicated to prefix the command with `bundle exec`. Other than this, the instructions worked great to set up a working dev env.


matiasgarciaisaia (approved):
👍 Great!

Yes, we've faced similar issues in the project already.

PR #1933, opened 2023-04-28
Setup Standard and Rubocop

bolom:
Add configuration files for [Standard](https://github.com/standardrb/standard) and [Rubocop](https://github.com/rubocop/rubocop) gems


ysbaddaden (commented):
That looks very nice :+1: 

A GitHub action to run Rubocop and it would be perfect.

Any cop that is merely about style could be enabled right now and autocorrected. As long as we have two distinct commits for the configuration _then_ the autocorrection, we can fix the pending pull requests (cherry-pick the configuration, autocorrect, commit).

.github/workflows/test.yml
ysbaddaden, line 1: A couple comments:

- We don't need to setup the database to run rubocop;
- But you must call the command in the docker container:
```suggestion
        run: docker compose run --rm --no-deps web bin/rubocop --parallel
```

.github/workflows/test.yml
```diff
@@ -94,3 +94,14 @@ jobs:
...
+  lint:
+    runs-on: ubuntu-latest
+    needs: setup
+    steps:
+      - uses: actions/checkout@v3
+      - run: docker compose build web
+      - uses: ./.github/actions/gems-cache
+
+      - name: Lint Ruby files
+        run: docker compose run --rm --no-deps web bin/rubocop --parallel
```
ysbaddaden, line 107:
```suggestion
        run: docker compose run --rm --no-deps web bundle exec rubocop --parallel
```

ysbaddaden (approved):
Looks perfect! I think we can squash & merge this PR and open a new one to enable at least the standard rules (syntax)?

PR #1934, opened 2023-04-28
Setup Brakeman

bolom:
Brakeman doesn't support Ruby 2.4, as suggested by the gem's author in [this conversation](https://github.com/presidentbeef/brakeman/issues/1649#issuecomment-964642785), I've used Brakeman-lib


ysbaddaden:
@bolom Then how do you invoke Brakeman?

Another solution would be to stick to Brakeman 5.0. We don't need the latest version when we use old Ruby and Rails versions. As long as it supports Rails 5.0 this is fine. We'll upgrade it along with other dependencies as we continue to upgrade.

bolom:
> @bolom Then how do you invoke Brakeman?
> 

We could maybe add a new action in Github action
```

 - name: Security audit application code
        run: bin/brakeman -q -w2
```

ysbaddaden:
Yeah, let's get Rubocop merged, and we can run Brakeman right after it.

bolom:
@ysbaddaden not sure what I done wrong. every thing is red now :(

ysbaddaden:
Looking at the CI logs, it complains about Nokogiri::HTML4 being undefined. Looking at Nokogiri's documentation:

> 💡 Before v1.12.0, Nokogiri::HTML4 did not exist, and Nokogiri::HTML was the module/namespace for parsing HTML.

There is an invalid dependency. I see Loofah got upgraded and Nokogiri's stuck at 1.10. We don't want to upgrade Nokogiri without continuing with the Ruby/Rails upgrade (CDx uses Nokogiri directly for a XML user feature).

Issue #1785, opened 2022-11-08
Fix UI glitches in the settings of roles and policies


diegoliberman:
#Steps to reproduce the glitch
1. WIth an admin user, go tot he settings menu
2. Go to the Roles section
3. Access the role detail page for any role
4. Click on a policy
5. Scroll down to the bottom --> GLITCH! You will be automatically scrolled back up to top of the page.

https://user-images.githubusercontent.com/25181052/219039468-0b276800-92c7-4f0d-9da9-71567432a7e8.mov

ysbaddaden:
@diegoliberman could you add some details?

diegoliberman:
> @diegoliberman could you add some details?

Done

ysbaddaden:
Oooh right. That rings a bell.

sardar-usman:
@bolom It looks good, working as expected.

PR #1937, opened 2023-05-10
Fix UI glitches in the settings of roles and policies #1785 

bolom:
This PR  fix UI glitches in the settings of roles and policies described in #1785 

**How to test ?**

1. WIth an admin user, go tot he settings menu
2. Go to the Roles section
3. Access the role detail page for any role
4. Click on a policy
5.  Scroll down to the bottom  without any problem


leandroradusky (approved):
I've tested and works well! 

:)

PR #1682, opened 2022-05-24
Adapt form error UI to design

straight-shoota:
Adjusts the UI for form errors according to https://invis.io/Z7OD63XAWKF#/467087061_New_Sample_-_Erros

Depends on #1689

PR #1939, opened 2023-05-17
Adapt form error UI to design

bolom:
Adjusts the UI for form errors according to https://invis.io/Z7OD63XAWKF#/467087061_New_Sample_-_Erros

### DIFF
--- a/app/helpers/cdx_form_helper.rb
+++ b/app/helpers/cdx_form_helper.rb
@@ -102,7 +102,7 @@ class FormFieldBuilder < ActionView::Helpers::FormBuilder
 
   # Returns `true` if the object has any errors.
   def has_errors?
-    !object_errors.empty?
+    !@object.errors.empty?
   end
 
   protected
--- a/app/helpers/cdx_form_helper.rb
+++ b/app/helpers/cdx_form_helper.rb
@@ -5,8 +5,6 @@ module CdxFormHelper
     form_for(record_or_name_or_array, *(args << options.merge(builder: FormFieldBuilder))) do |form|
       yield form
 
-      # `form_errors` could be called in the block, but that doesn't matter because a
-      # second call is a no-op.
       concat form.form_errors
     end
   end
@@ -55,6 +53,9 @@ class FormFieldBuilder < ActionView::Helpers::FormBuilder
 
   # Renders form errors for fields that have not been handled by a dedicated
   # `#field` errors.
+  #
+  # This method is called in `cdx_form_for` and should not be called explicitly
+  # in the form body.
   def form_errors(options = {})
     rendered = @template.render partial: "form_builder/form_errors", locals: {
       form: self,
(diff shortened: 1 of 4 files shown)


leandroradusky:
I was performing a smoke test, and it seems to work OK. **But**, there is an odd change on the date pickers:

Before (`main` branch): 

image

After:

image

Same is happening with all the date pickers.

bolom:
@leandroradusky  I've updated the css

leandroradusky (approved):
I went through the code & tested it locally and it looks good!

app/helpers/cdx_form_helper.rb
```diff
@@ -79,53 +79,80 @@ def form_actions(&block)
...
 
-    Rails.logger.info "Unhandled form errors in #{@object.model_name}: #{unhandled_errors}"
-    Raven.capture_message("Unhandled form errors", extra: {
-      form: @object.model_name,
+    Rails.logger.info message
+    Raven.capture_message("Unhandled form errors",
+      form: object_name,
       errors: unhandled_errors,
-    })
+    )
```
ysbaddaden, line 156: I don't know how the new Sentry gem is working, but I just verified the legacy Raven source code, and we must specify the `extra:` key (see `Raven::Event.new`).

```ruby
    Raven.capture_message("Unhandled form errors", extra: {
      form: object_name,
      errors: unhandled_errors,
    })
```

app/helpers/cdx_form_helper.rb
```diff
@@ -79,53 +79,80 @@ def form_actions(&block)
...
+    else
+      errors_to_show.clear
+    end
+  end
+
   def log_unhandled_errors
     unhandled_errors = error_messages_to_show.reject { |k, _| k == :base }
     return if unhandled_errors.empty?
 
+    message = "Unhandled form errors in #{object_name}: #{unhandled_errors}"
```
ysbaddaden, line 147: Where does the `object_name` variable come from?

ysbaddaden (changes requested):
Thanks @bolom but I'm kinda lost: why bring 28de0127e55d8ead250ea8b8ce5fd8e12bc9fcac from #1689 which we closed as too complex in favor to #1706? The only commit we want in is f0b6086ebea6970c99bce700e10ab681af2a560b from #1682, rebased from `main` and without the above commit from #1689.

Related slack discussion: 
https://manas.slack.com/archives/C034BR64J/p1684274918617989?thread_ts=1684268802.208069&cid=C034BR64J

I just tried a `git cherry-pick f0b6086ebea6970c99bce700e10ab681af2a560b` and the commit applies cleanly on top of main. Once `CdxFormHelper#has_errors?` is fixed, the design is in.

bolom:
@ysbaddaden voila, pushed my commit with your requests.

ysbaddaden (approved):
Perfect!

Issue #1863, opened 2023-01-24
New fields for Batches entity


leandroradusky:
The `Batch` entity within CDx is missing some data to be complete in regard to the data that needs to be uploaded to NIH.

These new columns need to be added to batches:
- reference_gene: all the calculations made are based on `copies` of the virus (i.e. `concentration` where its units is `copies/ml`) but what this count on reality is not copies of the virus, but based on a reference gene that can be counted experimentally. This depends on the methods used, in our scenario will be almost always equal to `N` [the nucleocapsid protein](https://en.wikipedia.org/wiki/Coronavirus_nucleocapsid_protein) but this, for other organisms could be whatever gene.
- target_organism_taxonomy_id: ID given to the organism in molecular biology databases like NCBI
- pango_lineage:  stands for Phylogenetic Assignment of Named Global Outbreak Lineages, a database with reference names widely used in this scenario
- who_label: name that the world health organization gives to the virus

Field labels:
1. Reference gene
2. Target organism taxonomy ID
3. Pango Lineage
4. Virus World Health Organization's label

All of them are combo free text fields:
1. The field should be autocomplete, but allow entering values that aren't in the list.
2. The list should include all values that were entered in previous batches for the current institution.
3. The field needs to be in the Batch and in the Sample. 
4. The list of options for Batches and Samples, should be the same.

leandroradusky:
I will do it in the main text of the issue when we have the full details... But as an advance: these two mentioned above are names that the viruses within the batches have in global reference databases: `WHO_label` is the name that the world health organization gives to the virus, similarly, `PANGO` stands for **P**hylogenetic **A**ssignment of **N**amed **G**lobal **O**utbreak Lineages, a database with reference names widely used in this scenario.

ysbaddaden:
Shall the fields be free text, or would we like to have a database of all these names? The former is prone to human typos, while the second means we must keep the databases updated, which we probably don't want to.

diegoliberman:
@leandroradusky what do you say if we make these fields dropdowns/autocomplete? Plus an "other" option for entering the value manually if not found?
Ideally, in the future the system should have some kind of support for adding new options into the tables, but they would also have the "other" option for not getting blocked.  
In the future, if they find a frequent need of adding options, a feature could be implemented for administering the values from the UI.

leandroradusky:
[...]

Issue #1381, opened 2021-12-09
Index JSON column that stores the core fields


diegoliberman:
The JSON columns in MySQL can be indexed.  See [related article](https://vladmihalcea.com/index-json-columns-mysql/).

By doing this change, we will save time in the future when implementing filters.  Today, when implementing a new filter we need to create an extra column that will have a redundant value of the field being filtered.  Redundant because it will be stored in the json and in the independent column.  But if we get to filter directly using the json column, we would avoid having to create the redundant column, and it would save us the effort of implementing 2 migrations.

ysbaddaden:
Problem: the `core_fields` column isn't serialized as JSON but as unsafe YAML. This will need a significant refactor of the entity framework to:

- migrate the text column (YAML) into a JSON column (de-serialize & re-serialize);
- serialize the `core_fields` as JSON;
- type cast entity attributes from/to the registered type ---currently the YAML is saved as-is, and the entity framework expects the unsafe YAML serializer to store and re-create the exact Ruby classes.

ysbaddaden:
A WIP is available in 505f2ebdb284ab679c3a5ab1ba0013405c01ff04. Untested, and we must verify that the migration won't break the databases in production!

PR #1940, opened 2023-05-23
#1863 AutoComplete Feature for Batch Model

bolom:
This PR implements the autocomplete feature for specific fields in the Batch entity.

Changes Made :

-   Added new fields to the Batch entity: reference_gene, target_organism_taxonomy_id, pango_lineage, and who_label.
-   Implemented the AutocompleteValue model, allowing the storage of unique autocomplete values for each field and institution.
-   Updated the BatchForm class to save autocomplete values when saving a batch.
-   Created the Autocomplete component in React to provide the necessary functionality for the autocomplete feature.


bolom:
The PR is a draft, I want to be sure I'm good direction :)

leandroradusky:
Hi!

When trying to save new batches I get:

image

When trying to access existing ones I get:

image

(is the same error I guess)

I checked out and run `rake db:migrate`... Fields are created... Should I have done something else?

app/assets/javascripts/components/autocomplete.js.jsx
```diff
@@ -0,0 +1,46 @@
+var Autocomplete = React.createClass({
+  getInitialState: function() {
+    return {
+      data: JSON.parse(this.props.data),
```
ysbaddaden, line 4: You shouldn't have to serialize to JSON in the view and deserialize in the React component. This is already taken care of by the view component library.

ysbaddaden (commented):
We already have autocomplete implemented with the [CdxSelectAutocomplete](https://github.com/instedd/cdx/blob/main/app/assets/javascripts/components/cdx_select_autocomplete.js.jsx) component, which can directly make a HTTP request to the backend.

Now, I'm not so sure with the direction. Why duplicate the values in the database? It requires to synchronize the states which is always complex (need to create, update & destroy). It would also be profitable to not duplicate key/value pairs (e.g. maybe keep a counter to know how many references we have).

It would make sense if we encounter performance issues by querying the main table directly, or... if the columns had been added as core fields (impossible to query until we get #1381 in) that now makes me wonder if these new attributes shouldn't be core fields? (see the [Entity Framework](https://github.com/instedd/cdx/wiki/Entity-Concern-Framework), something specific to CDx) @leandroradusky @devduarte? In which case we should implement the autocomplete table/model/controller after all, because we'd need it.

bolom:
> We already have autocomplete implemented with the [CdxSelectAutocomplete](https://github.com/instedd/cdx/blob/main/app/assets/javascripts/components/cdx_select_autocomplete.js.jsx) component, which can directly make a HTTP request to the backend.

 I skipped using this component because it doesn't allow us to input our own value, which is a requirement for the behavior,

ysbaddaden:
@bolom CdxSelectAutocomplete does allow to type text to autocomplete, but it indeed didn't send any free text we'd type with the query. Here is a small patch to allow this behavior. You can then use the component with:

```haml
= react_component "CdxSelectAutocomplete",
    { name: "resource[value]", value: params[:resource][:value], url: "" }
```



```diff
diff --git a/app/assets/javascripts/components/cdx_select_autocomplete.js.jsx b/app/assets/javascripts/components/cdx_select_autocomplete.js.jsx
index b3025571..d8f86c71 100644
--- a/app/assets/javascripts/components/cdx_select_autocomplete.js.jsx
+++ b/app/assets/javascripts/components/cdx_select_autocomplete.js.jsx
@@ -5,6 +5,16 @@ var CdxSelectAutocomplete = React.createClass({
     };
   },

+  // Since <Select> uses a different INPUT we must copy the value when the
+  // user is typing from the visible to the hidden INPUT, otherwise when the
+  // user enters free-text of validates with an invalid value, the backend won't
+  // know about the value and silently skip it.
+  copyValue: function (newValue) {
+    // we can't use setValue() because it has too many side effects, we only
+    // want to set the hidden input's value:
+    this.selectRef.setState({ value: newValue });
+  },
+
   render: function () {
     return (<Select
       ref={this.setSelectRef}
@@ -17,6 +27,7 @@ var CdxSelectAutocomplete = React.createClass({
       asyncOptions={this.asyncOptions}
       value={this.props.value || ""}
       onChange={this.onChange}
+      onInputChange={this.copyValue}
     />);
   },

```

bolom:
Voila. I've used CdxSelectAutocomplete for Batch and Model. Let me know what do you think now guys @leandroradusky @ysbaddaden

leandroradusky:
Hi @bolom ,

I think there's something missing in the commit, because I get 

```
undefined method `autocomplete_values_for' for #<Institution:0x00007fc35e34fde0>
```

when I try to add or edit batches.

bolom:
@leandroradusky should work now

leandroradusky:
Hi @bolom ...

I have been making a smoke test. It works ok, but I have two commentaries:

1) These new fields should be placed after the `Volume` field, not at the beginning of the form:  

image

2) `Reference Gene`, `PANGO lineage` & `WHO label` work OK, but `Target organism taxonomy` is broken, it does not save any values.

config/routes.rb
```diff
@@ -62,6 +62,7 @@
       get :no_data_allowed
       get :new_from_invite_data
     end
+    get 'autocompletes/:field_name/', to: 'autocompletes#index', as: 'autocomplete'
```
ysbaddaden, line 65: Hum, I'm not sure whether we should pass the `institution_id` in the URL. The whole project doesn't work that way, but instead passes the `context` param (institution + all sites or a single site) :thinking:

I'm wondering what we actually want, here. Shall we search within the whole institution (better to avoid duplicates at the global institution level)? Or shall we respect the currently selected site for autocompletes?

Poke @leandroradusky

app/views/samples/_form.haml
```diff
@@ -13,6 +13,15 @@
...
           = icon_tag "copy", class: "btn-icon"
 
+      = f.form_field :reference_gene do
+        = react_component "CdxSelectAutocomplete", { name: "sample[reference_gene]", value: f.object.reference_gene, url: institution_autocomplete_path(institution_id: @navigation_context.institution.id, field_name: "reference_gene") }
+      = f.form_field :target_organism_taxonomy do
+        = react_component "CdxSelectAutocomplete", { name: "sample[target_organism_taxonomy_id]", value: f.object.target_organism_taxonomy_id, url: institution_autocomplete_path(institution_id: @navigation_context.institution.id, field_name: "target_organism_taxonomy_id") }
+      = f.form_field :pango_lineage do
+        = react_component "CdxSelectAutocomplete", { name: "sample[pango_lineage]", value: f.object.pango_lineage, url: institution_autocomplete_path(institution_id: @navigation_context.institution.id, field_name: "pango_lineage") }
+      = f.form_field :who_label do
+        = react_component "CdxSelectAutocomplete", { name: "sample[who_label]", value: f.object.who_label, url: institution_autocomplete_path(institution_id: @navigation_context.institution.id, field_name: "who_label") }
```
ysbaddaden, line 23: What about a simple helper to create the react component? For example `autocomplete_field(f, :who_label)`

app/controllers/autocompletes_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+    render json: @unique_values.map { |option| { value: option, label: option } }
+  end
+
+  private
+
+  def unique_values_for(field_name, query)
+    batch_values = check_access(@institution.batches, READ_BATCH).collect(&field_name.to_sym).uniq
+    sample_values = check_access(@institution.samples, READ_BATCH).collect(&field_name.to_sym).uniq
+    (batch_values + sample_values).uniq.compact.select { |value| value.to_s.downcase.include?(query.downcase) }.split(',')
+  end
```
ysbaddaden, line 15: There is a typo: you should use `READ_SAMPLE` for samples.

ysbaddaden: We should restrict `params[:field_name]` to a list of acceptable field names to autocomplete. Here, any valid method on batch and Sample will be accepted.

ysbaddaden: We could probably optimize a little to avoid too many allocations, for example collect into a Set instead of building large arrays, uniq, building another array, uniq, etc. **But that's not the worst performance issue**, we have to load all batches and samples into ruby memory (because of entity framework), which probably won't do well with millions of samples, repeated on each character we type on the frontend (:fearful:).

Maybe we can load in batch, to avoid loading everything at once?

```ruby
matcher = Regexp.new(Regexp.escape(query), Regexp::IGNORECASE)
values = Set.new

check_access(@institution.batches, READ_BATCH).select(:core_fields).find_each do |batch|
  value = sample.__send__(field_name)
  values << value if value =~ matcher
end

check_access(@institution.samples, READ_SAMPLE).select(:core_fields).find_each do |sample|
  value = sample.__send__(field_name)
  values << value if value =~ matcher
end
```

Note:
- I'm using a regular expression to avoid `value.downcase` to avoid a String allocation/transform but I didn't benchmark to check which is actually faster.
- I assume that the method receives a symbol for `field_name` so it doesn't have to do cast everywhere.

ysbaddaden: Ideally, we should get #1381 in, so we could do that right in MySQL, but that might do in the meantime.

Your idea of an `autocompletes` table was very good, too.

ysbaddaden (commented):
Looks good for a first pass.

I'm not sure about performance with lots of samples and batches. We might be able to do better.

I'm also missing a few tests for the `autocompletes#index` action.

bolom:
I have made the necessary changes to the code based on your feedback. 

@leandroradusky  I was not able to reproduce the error, I successfully updated the `target_organism_taxonomy_id` as you requested. However, 

Regarding the tests, I intentionally didn't write them yet. My aim is to complete the logic first before working on the tests.

leandroradusky:
After correct the syntax error in my local, the four fields are properly saved as if were text fields, but existing values are not found and displayed in the list... 

Have an existing batch with `asd` value in the four fields:

image

But when creating a new one and inputting `asd` I get no results:

image

bolom:
@leandroradusky I was not able to reproduce the erro

This Sample
Capture d’écran 2023-06-20 à 11 40 26

When I'm trying to create a new sample

Capture d’écran 2023-06-20 à 11 41 36

From Batch Form

Capture d’écran 2023-06-20 à 11 42 52

leandroradusky:
I've found the exact conditions where the autocomplete doesn't work for me: it uploads existing values in the new fields if they are defined in `Sample`, but does not load existing values in the `Batch` model. I've reproduced what is happening in the rails console:

Having `Batch.select(:id, :core_fields, :uuid, :custom_fields).last` equal to:

```
#<Batch id: 54, uuid: "34c56061-141f-3998-0328-62949cbafd81", core_fields: {"batch_number"=>"asdf", "date_produced"=>2023-06-21 00:00:00 +0000, "lab_technician"=>"123", "specimen_role"=>"b", "isolate_name"=>"123", "inactivation_method"=>"Formaldehyde", "volume"=>"1", "virus_lineage"=>"Alpha", "reference_gene"=>"asd", "target_organism_taxonomy_id"=>"csd", "pango_lineage"=>"asd", "who_label"=>"asd"}, custom_fields: {}>
```

And `Sample.select(:id, :core_fields, :custom_fields).last``equal to:

```
#<Sample id: 2762, custom_fields: {}, core_fields: {"date_produced"=>2023-06-21 00:00:00 +0000, "lab_technician"=>"LeaR", "specimen_role"=>"b", "isolate_name"=>"asd", "inactivation_method"=>"Formaldehyde", "volume"=>"1", "virus_lineage"=>"12", "concentration"=>"1", "replicate"=>"1", "media"=>"Culture media", "reference_gene"=>"asd", "target_organism_taxonomy_id"=>"bsd", "pango_lineage"=>"asd", "who_label"=>"asd"}>
```

I get `nil` when I execute: `Batch.select(:id, :core_fields, :uuid, :custom_fields).last.try("who_label")`

But I get `asd` when I execute: `Sample.select(:id, :core_fields, :custom_fields).last.try("who_label")`

But if even more strange: If I do the try with `"inactivation_method"`, both `Batch` and `Sample` will return `Formaldehyde`.

Can you reproduce this?

bolom:
@leandroradusky are you sure Sample.last == Sample(id). 
Can you just try?
`Sample.find(2762).who_label`

leandroradusky:
Sure! Here are the outputs:

```
irb(main):012:0> Sample.find(2762) == Sample.last
  Sample Load (0.3ms)  SELECT  `samples`.* FROM `samples` WHERE `samples`.`deleted_at` IS NULL AND `samples`.`id` = 2762 LIMIT 1
  Sample Load (0.2ms)  SELECT  `samples`.* FROM `samples` WHERE `samples`.`deleted_at` IS NULL ORDER BY `samples`.`id` DESC LIMIT 1
=> true
```

```
irb(main):015:0> Sample.find(2762).who_label
  Sample Load (0.3ms)  SELECT  `samples`.* FROM `samples` WHERE `samples`.`deleted_at` IS NULL AND `samples`.`id` = 2762 LIMIT 1
NoMethodError: undefined method `who_label' for #<Sample:0x0000559836c4fd00>
        from /usr/local/bundle/gems/activemodel-5.0.7.2/lib/active_model/attribute_methods.rb:433:in `method_missing'
        from (irb):15
        from /usr/local/bundle/gems/railties-5.0.7.2/lib/rails/commands/console.rb:65:in `start'
        from /usr/local/bundle/gems/railties-5.0.7.2/lib/rails/commands/console_helper.rb:9:in `start'
        from /usr/local/bundle/gems/railties-5.0.7.2/lib/rails/commands/commands_tasks.rb:78:in `console'
        from /usr/local/bundle/gems/railties-5.0.7.2/lib/rails/commands/commands_tasks.rb:49:in `run_command!'
        from /usr/local/bundle/gems/railties-5.0.7.2/lib/rails/commands.rb:18:in `<top (required)>'
        from bin/rails:4:in `require'
        from bin/rails:4:in `<main>'
```

ysbaddaden (approved):
Nice!

@leandroradusky please make one last round of tests, see if you can reproduce any bug, then LGTM @bolom !

leandroradusky:
For me is good to go!

leandroradusky (approved):

bolom:
@ysbaddaden poke let me know if it's good to go for you

ysbaddaden (approved):
LGTM

PR #1951, opened 2023-06-02
Display errors on invalid files when uploading boxes from CSV

leandroradusky:
Fixes error find while testing #1814.

The created `CsvValidation` controller was responding with an unprocessable entity when there was an error but nothing was done with that, then the row wasn't displayed for the uploaded file, misleading the users about what was happening.
Now, when the file is invalid, it looks like this:

image

### DIFF
--- a/app/controllers/csv_validations_controller.rb
+++ b/app/controllers/csv_validations_controller.rb
@@ -7,10 +7,7 @@ class CsvValidationsController < ApplicationController
       csv_table = CSV.parse(csv_text, headers: true)
 
       unless csv_table.headers == ["Batch", "Concentration", "Distractor", "Instructions"]
-        render json: { found_batches: [],
-                       not_found_batches: [],
-                       samples_nbr: 0, 
-                       error_message: "Invalid file format." }
+        render_invalid_format
         return
       end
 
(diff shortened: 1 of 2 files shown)


app/controllers/csv_validations_controller.rb
```diff
@@ -5,15 +5,30 @@ def create
...
+                     samples_nbr: csv_table.size,
+                     error_message: "" }
     else
-      render json: { error: "Invalid file format. Please upload a CSV file." }, status: :unprocessable_entity
+      unless csv_table.headers == ["Batch", "Concentration", "Distractor", "Instructions"]
+        render json: { found_batches: [],
+                       not_found_batches: [],
+                       samples_nbr: 0, 
+                       error_message: "Invalid file format." }
+      end
```
ysbaddaden, line 31: This is identical to the above. Maybe extract a method for rendering the error json? and I believe we don't need to check the csv headers if the file isn't a text/csv file?

leandroradusky: >  and I believe we don't need to check the csv headers if the file isn't a text/csv file?
:facepalm:

ysbaddaden (approved):

PR #1953, opened 2023-06-16
Add Lod and Lob Caculation

bolom:
This PR include caculation of Limit of Detection and Limit of blank. I've used the gem  [rumale](https://github.com/yoshoku/rumale)


leandroradusky:
Hi @bolom, besides the small comments I've left in the code, one thing to fix is that the new fields make the page structure of the PDF report to brake: 

image

The fields should be organized into columns, please take as a reference the PDF for the boxes of purpose = Challenge, where more fields are informed and the confusion matrix fits completely into the first page:

image

app/models/samples_report.rb
```diff
@@ -27,27 +29,58 @@ def self.entity_scope
...
         .group(:samples_report_id) unless batch_number.blank?
-  }
+      }
 
+      def calculate_lod_and_lob
+        concentrations = samples.map(&:concentration)
+        signals = Numo::DFloat[*samples.map(&:measured_signal)]
+        self.lod, _, self.lob, _ = calculate_lod(concentrations, signals)
+        self.save
+      end
```
ysbaddaden, line 52: That method ought to have some unit tests.

app/models/samples_report.rb
```diff
@@ -27,27 +29,58 @@ def self.entity_scope
...
+          lob = nil
+        end
+
+        gray_zone = lob.nil? ? nil : (lob..lod)
+        return lod.round(3), rmse, lob.round(3), gray_zone
+      end
 
-end
\ No newline at end of file
+    end
```
ysbaddaden, line 86: nitpick: indentation looks wrong in the whole file? it seems it broke around the nested scope>joins above.

ysbaddaden (commented):
Nice!

I'd love to see an unit test for the `SamplesReport#calculate_lod_and_lob` method.

It seems we can just import `rumale-linear_model`, instead of pulling `rumale` that pulls far too many libraries we don't use (yet?)

bolom:
et voila

ysbaddaden (commented):
Thanks, that really cleans things up!

@leandroradusky Do you think you could help @bolom with some data to write a test: given a set of samples with these measured signals, when creating a sample report with these settings, it should calculate these LOD and LOB.

We could validate that the calculation is correct, and prevent further regressions.

bolom:
@leandroradusky helped me last week to validate the calculations using some data. We could use the same data to create the test for calculating LOD and LOB.
```
concentrations [0, 10, 10, 10, 20, 0]
signal [10, 100.12, 193.5, 100.12, 193.5, 3.3]
Blank Sample size : 2
Limit of Detection: 17.651812199460668
Root Mean Square Error: 34.65889649273338
Limit of Blank: 14.443377388847534
Gray Zone: 14.443377388847534..17.651812199460668
```

leandroradusky:
These numbers are ok to make a test :+1: (to test the function you will have to round them to three decimal places since we're returning the rounded number). 

Still think that`rmse` & `gray zone`  should be removed from the function and `calculate_lod` & `calculate_lod_and_lob` should be merged into one function or at least renamed properly to reflect what they do before merging this PR.

leandroradusky:
I continued the testing, is almost ready to merge. I've detected two things:

- The indentation of the spec file is broken.
- When creating a report, could be that there are samples of the box that doesn't have uploaded `measured_signal` and these samples will be ignored, as pointed out in the report creation screen:

image

When this happens, and we try to create a report, the following error is thrown:

image

nil values should be ignored in the mapping to float.

leandroradusky (approved):
For me is good to go! :D 

:rocket:

spec/controllers/samples_reports_controller_spec.rb
```diff
@@ -9,6 +9,14 @@
     @other_institution = Institution.make! user: @other_user
 
     @box = Box.make!(:overfilled, institution: @institution, purpose: "LOD")
+
+
+    concentrations = @box.samples.map(&:concentration)
+    signals = Numo::DFloat[*@box.samples.map(&:measured_signal)]
+
+    puts concentrations.inspect
+    puts signals.inspect
```
ysbaddaden, line 18: manual debug leftovers

bolom: @ysbaddaden ah zut. Should I do a revert or a new branch?

ysbaddaden: No need for a revert, we'd end up with 3 commits in total (old commit, revert commit, new commit) :smile:

Create another branch and squash it so we'll have 2 commits (commit, fixup commit)

PR #1954, opened 2023-06-20
Allow to trigger the download NIH report

### LINKED ISSUE #1917: Allow to trigger the download NIH report
In the Sample Report detail:
1. Add a button to download the NIH files
2. The button should open a warning modal saying "After downloading these files, you will need to fill specific columns.  Make sure you read the instructions to be aware of these considerations, in the file `Instructions.txt`.".

leandroradusky:
Closes #1917.

Now, in the detail of the `SamplesReport`, a new button is displayed to download the NIH files:

image

As the issue specifies, when clicking the button the following modal is shown:

image

This triggers the download of a zip file, which at the moment has only the file `Instructions.txt` (the other files will be generated in further issues). This file is empty and its contents will be defined together with the client (it needs to be uploaded to a specific portal and we need to collect this info). 

**NOTE:**
In the `ConfirmationModal`, the `X` button to close it, calls the `hide()` method of the `<Modal>` component that it renders, but this is not how the `ConfirmationModal` is used: it has a `hidden` class that is added or removed as needed. To avoid problems with the other uses of `<Modal>` in the project, a new option for the confirmation modal was added, `showCloseButton`, settled as false in all the uses. This fixes a bug, since when `ConfirmationModal` was used and closed with the `X` and not with the `Cancel` button, it was breaking its functioning.


app/controllers/samples_reports_controller.rb
```diff
@@ -135,6 +137,29 @@ def update_threshold
     render json: { threshold: threshold, confusion_matrix: confusion_matrix }
   end
 
+  def download_nih_tables
```
bolom, line 140: You know, I'm not a big fan of non-CRUD methods, to be honest. But I've got a proposal that I think you might find interesting

```
class NIHTablesController < ApplicationController
  before_action :set_samples_report

  def show
    samples_report = SamplesReport.find(params[:samples_report_id])
    return unless authorize_resource(samples_report, READ_SAMPLES_REPORT)
    
    purpose = samples_report.samples[0].box.purpose
    zip_file = create_zip_file(samples_report.name)

    if purpose == "LOD"
      add_lod_table(zip_file)
    elsif purpose == "Challenge"
      add_challenge_table(zip_file)
    end

    zip_file.close

    send_zip_file(zip_file, samples_report.name)
  end

  private

  def create_zip_file(filename)
    zip_file = Tempfile.new("#{filename}_nih_tables.zip")
    Zip::File.open(zip_file.path, Zip::File::CREATE) do |zip|
      zip.add("Instructions.txt", Rails.root.join('public/templates/Instructions.txt'))
    end
    zip_file
  end

  def add_lod_table(zip_file)
    # TODO: Add the LOD table to the zip file
  end

  def add_challenge_table(zip_file)
    # TODO: Add the Challenge table to the zip file
  end

  def send_zip_file(zip_file, filename)
    send_file zip_file.path, type: 'application/zip', filename: "#{filename}_nih_tables.zip"
  end
end

```

routes

```
 resources :samples_reports, only: [] do
    member do
      get 'nih_tables', to: 'nih_tables#show'
    end
  end
```

bolom (approved):

app/controllers/samples_reports_controller.rb
bolom: change my mind. CREATE is better than Show you're creating a new download each time, but it comes with its own set of trade-offs 

`resources :nih_tables, only: [:create]`

ysbaddaden: Yes, here it makes sense to create another controller!

You may even nest it under the :samples_reports resource:

```ruby
resources :samples_reports do
  resources :nih_tables, only: [:create]
end
```

So you get an explicit route name, and the explicit `:samples_report_id` param in your controller, instead of the plain `:id` that could give the impression that "nih tables" is a resource.

```ruby
samples_reports_nih_tables_path(samples_report)
# => /samples_reports/:samples_report_id/nih_tables
```

ysbaddaden: Note: it might not be easy to use `#create` instead of `#show` because we link from a React component where it might be more complex than necessary to create a form with the CSRF token :/

app/assets/javascripts/components/modal.js.jsx.erb
```diff
@@ -1,6 +1,7 @@
 var Modal = React.createClass({
   getInitialState: function() {
-    return { show: false };
+    return { show: false,
+            showCloseButton : this.props.showCloseButton === undefined ? true : this.props.showCloseButton};
```
ysbaddaden, line 4: We may want to loosely check against `null` to accept both `null` and `undefined`:

```js
showCloseButton : this.props.showCloseButton == null ? true : this.props.showCloseButton};
```

Or just reverse the logic:

```js
showCloseButton: this.props.showCloseButton !== false
```

app/views/samples_reports/show.haml
```diff
@@ -49,6 +49,10 @@
                 .subtitle{id:'print-as-pdf'}
                   = icon_tag "as-pdf", class: "btn-icon icon-text-color"
                   Download PDF
+            .action
+              =button_tag id:'print-as-csv', class: 'subtitle btn-transparent' do
```
ysbaddaden, line 53: A DOM ID must be unique on a page. You want a CLASS or a DATA attribute that doesn't have to be unique.

leandroradusky: But this element is unique in the DOM, indeed there is not other `print-as-csv` elements in the whole project. I'm getting it wrong?

app/controllers/samples_reports_controller.rb
leandroradusky: By nesting I'm asked for a `respond_to` a format and the route becomes a post... But I'm doing a `GET` from js, not from react. Like it is, is working ok with the `#create` method (and also with the `#show` method).

`/views/samples_reports/show.haml:293`:
```
// Open the download_nih_tables path in a new tab
    var context = getUrlParameter('context');
    var url = '/samples_reports/'+#{@samples_report.id}+'/nih_tables.html?context='+context+"";
    window.open(url, '_blank');
```

bolom (approved):

app/views/samples_reports/show.haml
```diff
@@ -258,4 +274,30 @@
       document.getElementById("samples-report").classList.remove("hidden");
     });
   }
-  
+
+  // For NIH download button
+
+  var nihDownloadButton = document.querySelectorAll("#print-as-csv");
```
ysbaddaden, line 280: Oh, since the element is unique, you can `button = document.getElementById("print-as-csv")` then directly `button.addEventListener(...)`.

ysbaddaden (commented):
Just a couple things to fix, then LGTM:

1. You didn't remove the `SamplesReports#download_nih_tables` method
2. If you're using a GET then maybe the method should be `#show` instead of `#create` for which we'd expect a POST "execute an action" vs GET "fetch something".

Last but not least: a simple test for the action that runs the action and verifies that we get a ZIP file would make the PR perfect.

bolom (approved):

PR #1956, opened 2023-06-27
General nih tables exporting

### LINKED ISSUE #1905: General NIH tables
The columns `target_organism_name` and `target_analyte_type` will have dynamic data depending on the dropdown selection (see #1917).  

The general tables for any box purpose to be exported are two: Results and Samples.

**Samples**: The template for this file has the following fields:

- sample_id: self-explanatory
- vqa_box_id: the `uuid` of the box
- sample_group: the `purpose` of the box + `"-panel"`
- sample_media: _to be completed by the awardee_
- sample_media_id: _to be completed by the awardee_
- sample_media_source: _to be completed by the awardee_
- sample_media_source_url: _to be completed by the awardee_
- sample_concentration: self-explanatory
- sample_concentration_unit: fixed to `copies/ml`
- sample_concentration_reference_gene: _new field in batch entity, see #1863_
- virus_sample_id: same as `sample_id`
- virus_batch_number: self-explanatory
- target_analyte_type: fixed to `inactivated virus`
- target_organism_name: fixed to `SARS-CoV-2`
- target_organism_taxonomy_id: _new field in batch entity, see #1863_
- pango_lineage: _new field in batch entity, see #1863_
- who_label: _new field in batch entity, see #1863_
- virus_sample_inactivation_method: self-explanatory

**Results**: The template for this file is the following: 

image

The fields that should be fulfilled by CDx are:
- sample_id: self-explanatory
- vqa_box_id: the `uuid` of the box
- sample_group: the `purpose` of the box + `"-panel"`
- assay_readout: measured signal

The other columns should be there be populated by the Awardee.

leandroradusky:
Closes #1905.

The two general tables (samples & results) specified in the issue are generated as described in the text of the issue and added to the zip file downloaded when the user selects `Download NIH files`. 

The only commentary for the code I think is that the following line

`batch = Batch.find_by(batch_number: sample.batch_number)`

is used to load the batch instead of using `sample.batch` to allow loading fields for deleted batched.

### DIFF
--- a/app/controllers/nih_tables_controller.rb
+++ b/app/controllers/nih_tables_controller.rb
@@ -1,5 +1,3 @@
-require 'csv'
-
 class NihTablesController < ApplicationController
   def show
     samples_report = SamplesReport.find(params[:id])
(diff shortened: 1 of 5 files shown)


spec/controllers/nih_tables_controller_spec.rb
```diff
@@ -38,6 +38,37 @@
...
+      expect(Zip::File.open_buffer(response.body).entries.map(&:name)).to include("#{@samples_report.name}_samples.csv")
+      expect(Zip::File.open_buffer(response.body).entries.map(&:name)).to include("#{@samples_report.name}_results.csv")
+    end
+
+    it "should contain the samples table with the correct data" do
+      get :show, params: { id: @samples_report.id }
+
+      samples_table = CSV.parse(Zip::File.open_buffer(response.body).entries.find{|e| e.name == "Test_samples.csv"}.get_input_stream.read, headers: true)
+      
+      #expect the samples table to contain the same number of rows as the samples report
```
ysbaddaden, line 53: No need to comment every expectation. The expectations are enough to understand what's being tested.

ysbaddaden (commented):
Nice, a pull requests with tests! Yay!

Maybe we can extract the CSV templates into actual templates to clean up the controller?

CDx already depends on the [csv_builder](https://github.com/gtd/csv_builder) gem, so maybe we can extract a CSV template for each file, that we can then render with [#render_to_string](https://api.rubyonrails.org/v5.0/classes/ActionController/Rendering.html#method-i-render_to_string) which is identical to `#render` but returns a string instead of writing the response's body.

Then, the controller would be more about the logic (build a ZIP file), rather than details (what's in the files).

app/controllers/nih_tables_controller.rb
```diff
@@ -18,119 +16,27 @@ def create_zip_file(samples_report)
...
-      end
-    end
-  
-    zip_file.add("#{samples_report.name}_results.csv", csv_file.path)
-  end
+  def add_nih_table(samples_report, table_name, zip_file)
+    csv_file = Tempfile.new("#{samples_report.name}_#{table_name}.csv")
+    
+    csv_file.write(
+      render_to_string(:file => 'samples_reports/nih_'+table_name+'.csv.csvbuilder', :locals => { :samples_report => samples_report })
```
ysbaddaden, line 35: You can set `samples_report` as an ivar so you won't have to pass it to the method nor the template. I believe you can also simplify the call to let rails pick the correct `controller/template.format.type`:

```ruby
render_to_string("nih_#{table_name}", formats: :csv)
```

app/controllers/nih_tables_controller.rb
```diff
@@ -18,119 +16,27 @@ def create_zip_file(samples_report)
...
+    csv_file.close
 
-  def add_lod_table(samples_report, zip_file)
-    # TODO: Add the LOD table to the zip file
-  end
-  
-  def add_challenge_table(samples_report, zip_file)
-    # TODO: Add the Challenge table to the zip file
+    zip_file.add("#{samples_report.name}_#{table_name}.csv", csv_file.path)
   end
```
ysbaddaden, line 40: You'll want to delete (unlink) the tempfile after it has been added to the ZIP file.

ysbaddaden (commented):
A few tiny thingies, then :+1:

app/controllers/nih_tables_controller.rb
leandroradusky: I've tried this myself before (delete and unlink) but I get:

image

if I do 

```
    csv_file.close
    zip_file.add("#{@samples_report.name}_#{table_name}.csv", csv_file.path)
    csv_file.unlink
```

I'm doing it wrongly?

leandroradusky: I saw that there is an alternative [using streams](https://stackoverflow.com/questions/27577838/using-tempfile-to-create-a-zip-file-in-rails) but also, from the docs:

> When a Tempfile object is garbage collected, or when the Ruby interpreter exits, its associated temporary file is automatically deleted.

How do we should proceed here?

ysbaddaden: Argh, I hoped the Zip builder would generate the Zip file as we add files to it (stream by default), but no, it keeps a list of files to add and eventually generates the Zip file by reading the files to add. Damn.

You may want to try streaming. That may end up simpler as we could skip tempfiles... unless we expect larger CSV, like thousands of samples?

At worst, we're leaking file descriptors, and Ruby will run a GC collect when there are none available, but let's try to avoid that.

ysbaddaden: Same for the tempfile of the generated Zip, it should be deleted after being sent to the user.

ysbaddaden (approved):

PR #1958, opened 2023-07-04
Rake task to update new fields for existing batches

### LINKED ISSUE #1922: Update new fields for the existing batches
The attached Excel contains the new fields data for the existing batches.  We need to update the records on production with this data.
1. The production target institution is UCSD, id=18.  Please double check this with Mati.  If the institution has more than 1k samples, then it's the one.
2. We need to loop the Excel batches, find the corresponding one in CDx, and update the 4 fields.
3. For each batch that was found, we need to search the samples associated to the batch and also update the 4 fields.
4. We also need to search for Samples associated to the batch but were transferred to another institution, and update the 4 fields.

This is 1 time script.  But, it could be useful for the future.  If later we add even more fields to the batch, maybe we need to go through the same.  It would be worth keeping this script in the repository. 

[BatchIDs_2023-02-14.xlsx](https://github.com/instedd/cdx/files/10741784/BatchIDs_2023-02-14.xlsx)

leandroradusky:
Close #1922.

Created a new `rake` task to update the new fields (`reference_gene`, `target_organism_taxonomy_id`, `pango_lineage`, `who_label`) in `Batches`. This columns have to exist as headers into the csv file and also the `batch_number` column.

The file provided by the client was added into the `public/` folder. It will update only existing batches. Tested locally and appears to work correctly.


ysbaddaden (commented):
Maybe we don't need to commit the actual CSV file, but then we'd still need to share it somewhere for @matiasgarciaisaia to be able to apply it on radx servers..?

ysbaddaden (approved):

leandroradusky:
Letting the final file here to don't lose it in the eter.

[batches_label_data.csv](https://github.com/instedd/cdx/files/11980707/batches_label_data.csv)

Issue #1907, opened 2023-02-08
Specific Challenge results table


leandroradusky:
The columns `target_organism_name` and `target_analyte_type` will have dynamic data depending on the dropdown selection (see #1917).  

When exporting the tables from Challenge boxes, an additional table should be exported with the following columns:

- protocol_id: _to be completed by the awardee_
- technology_platform: _to be completed by the awardee_
- vqa_box_id: the `uuid` of the box
- sample_group: the `purpose` of the box + `"-panel"`
- target_analyte_type fixed to `inactivated virus`
- target_organism_name: fixed to `SARS-CoV-2`
- target_organism_taxonomy_id: new field in batch entity, see #1863 
- pango_lineage: new field in batch entity, see #1863 
- who_label: new field in batch entity, see #1863 
- non_clinical_cutoff_min: the actual selected `threshold` in the report
- non_clinical_cutoff_unit: fixed to `copies/ml`
- true_positives: its actual value in the confusion matrix based on the selected `threshold`
- false_positives: its actual value in the confusion matrix based on the selected `threshold`
- true_negatives: its actual value in the confusion matrix based on the selected `threshold`
- false_negatives: its actual value in the confusion matrix based on the selected `threshold`
- non_clinical_sensitivity: actual value of TPR  based on the selected `threshold`
- non_clinical_sensitivity_ci_percent **+**
- non_clinical_sensitivity_ci_min **+**
- non_clinical_specificity: actual value of FPR  based on the selected `threshold`
- non_clinical_specificity_ci_percent **+**
- non_clinical_specificity_ci_min **+**
- area_under_the_roc_curve: value already computed in the report for the selected threshold.
- roc_x_values: list of the x axis values of the roc-curve plot
- roc_y_values: list of the y axis values of the roc-curve plot

**+** In all the `_ci` fields CI stands for Confidence Interval. This can be computed but, given the low number of replicates won't be really informative, to be defined

sardar-usman:
@leandroradusky It is working fine.

PR #1959, opened 2023-07-07
Specific purposes nih tables exporting

### LINKED ISSUE #1906: Specific LOD results table
The columns `target_organism_name` and `target_analyte_type` will have dynamic data depending on the dropdown selection (see #1917).  

When exporting the tables from LOD boxes, an additional table should be exported with the following columns:

- protocol_id: _to be completed by the awardee_
- technology_platform: _to be completed by the awardee_
- vqa_box_id: the `uuid` of the box
- sample_group: the `purpose` of the box + `"-panel"`
- target_analyte_type fixed to `inactivated virus`
- target_organism_name: fixed to `SARS-CoV-2`
- target_organism_taxonomy_id: new field in batch entity, see #1863 
- pango_lineage: new field in batch entity, see #1863 
- who_label: new field in batch entity, see #1863 
- limit_of_blank: new computed field in the report, see #1904 
- limit_of_blank_unit: fixed to `copies/ml`
- limit_of_detection: computed in the report
- limit_of_detection_reference_gene: new field in batch entity, see #1863  
- limit_of_detection_unit: fixed to `copies/ml`
- limit_of_detection_sn_ratio: **to be confirmed** fixed to a value based on SD (standard deviation) used in LOD calculation
- limit_of_detection_ci_percent: **to be confirmed** fixed to a value based on SD (standard deviation) used in LOD calculation
- limit_of_detection_detail: fixed to `linear regression`
- analytical_sensitivity: _to be completed by the awardee_
- analytical_sensitivity_unit: _to be completed by the awardee_

leandroradusky:
Close #1906 & #1907 

Export specific tables por `LOD` & `Challenge` purposes. The only remark to do is that the `Challenge` purpose rely on the client selected `threshold` and for that we need to make `async` the function `downloadNIHFiles` and get the params for the table as we do with other reports.


app/views/samples_reports/nih_challenge.csv.csvbuilder
```diff
@@ -0,0 +1,55 @@
...
+  'non_clinical_specificity',
+  'non_clinical_specificity_ci_percent',
+  'non_clinical_specificity_ci_min',
+  'area_under_the_roc_curve',
+  'roc_x_values',
+  'roc_y_values'
+]
+
+sample = @samples_report.samples[0]
+batch = Batch.find_by(batch_number: sample.batch_number)
```
ysbaddaden, line 29: You should load `@batch` in the controller directly, or better add a `SamplesReport#batch` method, so the logic to find the batch associated to a SamplesReport is abstracted in the model.

Cherry on the cake: using SQL to find the batch?

```ruby
Batch.joins(:samples).where(samples_report_id: id).take
```

app/views/samples_reports/show.haml
```diff
@@ -294,14 +294,39 @@
...
+          .then((ret) => {
+            var params = {};
+            params["true_positives"] = ret.confusion_matrix.true_positive || 0;
+            params["true_negatives"] = ret.confusion_matrix.true_negative || 0;
+            params["false_positives"] = ret.confusion_matrix.false_positive || 0;
+            params["tpr"] = (ret.confusion_matrix.true_positive || 0) / ((ret.confusion_matrix.true_positive || 0) + (ret.confusion_matrix.false_negative || 0));
+            params["fpr"] = (ret.confusion_matrix.false_positive || 0) / ((ret.confusion_matrix.false_positive || 0) + (ret.confusion_matrix.true_negative || 0));
+            params["auc"] = #{auc(roc_curve(@samples_report))};
+            params["roc_x_values"] = #{roc_curve(@samples_report).map{|x| x[0]}};
+            params["roc_y_values"] = #{roc_curve(@samples_report).map{|x| x[1]}};
```
leandroradusky, line 316: Didn't found a way to call rails `roc_cuve` and store it in javascript to don't repeat these calls.

app/views/samples_reports/show.haml
```diff
@@ -294,14 +294,39 @@
...
     hideConfirmModal();
-    
-    // Open the download_nih_tables path in a new tab
     var context = getUrlParameter('context');
-    var url = '/samples_reports/'+#{@samples_report.id}+'/nih_tables?context='+context+"";
-    window.open(url, '_blank');
-
+    
+    // if purpose is challenge, then we need the live threshold calculations
+    if ("#{@purpose}" == "Challenge") {
```
ysbaddaden, line 302: Maybe:

```suggestion
    if (#{@purpose.to_json} == "Challenge") {
```

app/views/samples_reports/show.haml
```diff
@@ -294,14 +294,39 @@
...
+      var threshold = document.getElementById("threshold").value;
+      var url = '/samples_reports/update_threshold?context='+context+"&samples_report_id="+#{@samples_report.id}+"&threshold="+threshold;
```
ysbaddaden, line 304: We can just inject the ID right into the String:

```suggestion
      var url = '/samples_reports/update_threshold?context='+context+"&samples_report_id=#{@samples_report.id}&threshold="+threshold;
```

app/views/samples_reports/show.haml
```diff
@@ -294,14 +294,39 @@
...
+            params["fpr"] = (ret.confusion_matrix.false_positive || 0) / ((ret.confusion_matrix.false_positive || 0) + (ret.confusion_matrix.true_negative || 0));
+            params["auc"] = #{auc(roc_curve(@samples_report))};
+            params["roc_x_values"] = #{roc_curve(@samples_report).map{|x| x[0]}};
+            params["roc_y_values"] = #{roc_curve(@samples_report).map{|x| x[1]}};
+
+            // create url params with the params dictionary
+            var urlParams = new URLSearchParams(Object.entries(params));
+
+            // Open the download_nih_tables path in a new tab
+            var url = '/samples_reports/'+#{@samples_report.id}+'/nih_tables?context='+context+"&"+urlParams.toString();
```
ysbaddaden, line 322:
```suggestion
            var url = '/samples_reports/#{@samples_report.id}/nih_tables?context='+context+"&"+urlParams.toString();
```

app/views/samples_reports/show.haml
ysbaddaden, line 316: nitpick: You can directly set each object attributes:

```js
params.true_negatives = ...
```

Or directly build the object with attributes:

```js
var params = {
  true_positives: ...,
}
```

ysbaddaden (commented):
Good job! I just left a few details & nitpicks :)

app/views/samples_reports/nih_challenge.csv.csvbuilder
leandroradusky: Indeed we wanted the non-distractor batch to be taken as the target batch... Thanks to this correction the error is solved :)

leandroradusky: Corrected by rubocop:

`Rails/FindBy: Use find_by instead of where.take.
    samples_report_samples.joins(:sample).where("samples.distractor IS NULL OR samples.distractor = 'false'").take.sample.batch`

app/models/samples_report.rb
```diff
@@ -68,6 +68,11 @@ def calculate_lod_and_lob
     self.save
   end
 
+  def target_batch
+    # The target batch for this box is the batch of any sample which distractor is false or null
+    samples_report_samples.joins(:sample).where("samples.distractor IS NULL OR samples.distractor = false").first.sample.batch
```
ysbaddaden: FYI: this is acceptable as we don't call the method much, but this will make a number of SQL requests:

```sql
select * from samples_reports_samples limit 1
select * from samples joins limit 1
select * from batches limit 1
```

The ideal would be to select from batch and use joins to find it from samples reports. I just tried, and the following should work...

```ruby
Batch
  .joins(:samples => :samples_report_sample)
  .where("samples_report_samples.samples_report_id = ?", id)
  .where("samples.distractor <> ?", true)
  .take
```

... if you must add the missing association on `Sample`:

```ruby
class Sample
  has_one :samples_report_sample
end

This is a problem of the active-record pattern, it's easy to use the object relations, and we can easily lose track of how many requests it will actually make in the background.

ysbaddaden: Also you should memoize the object, for example `@target_batch ||= ...`, otherwise each call to the method will run the SQL request(s).

ysbaddaden (commented):
A couple comments to optimize then LGTM

PR #1979, opened 2023-07-26
Button allowing to copy uuid for each sample in box detail

### LINKED ISSUE #1967: Allow to copy sample ID and link to the sample detail, all from the box detail
From the Box detail, it would be awesome to be able to easily copy the sample ID, and to open their sample detail.  It will help a lot when testing and trying to verify-double check the system is creating the samples as expected.

image

leandroradusky:
Closes #1967 

Now, in the box detail, there is a button allowing the copy for each sample uuid composing the box.

image

### DIFF
--- a/app/assets/stylesheets/_buttons.scss
+++ b/app/assets/stylesheets/_buttons.scss
@@ -229,4 +229,9 @@
   &:hover {
     border: 1px solid #999;
   }
+
+  &.v-aligned-box-detail{
+    position: absolute;
+    left: 300px;
+  }
 }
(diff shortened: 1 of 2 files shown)


ysbaddaden (commented):
This is fine, yet from a design point of view I'd like the buttons to be vertically aligned.

leandroradusky:
Now it looks like this :) 

image

ysbaddaden (approved):
A horizontal flex would be more flexible, but that will do.

PR #1981, opened 2023-07-31
Fix: ensure that samples' measured signals are floats

### LINKED ISSUE #1966: Accessing the report crashes when the samples values are changed
### Steps to reproduce the error
1. Create a box, transfer it to another institution, confirm the transfer, upload the results, create a report.
2. You'll see that you can access the report normally.
3. Open any sample from the box.   Edit any value, for example the Technician name.
4. Go back to the report and try to open it.  
5. It will crash.

image

### LINKED ISSUE #1968: Downloading NIH table crashes
I wasn't able to detect what am I doing in particular to make it crash, but every time I followed the flow for creating the report, it crashed when downloading the NIH report.

Update: it crashed even for a new institution created from scratch

ysbaddaden:
I believe normalizing the value should fix samples reports related issues in staging, and it won't hurt to run them in production, to make sure that the values, when present, are floats, not a blank string or a float string.

closes #1966 
closes #1968

### DIFF
--- a/db/migrate/20230731140544_fix_samples_measured_signals.rb
+++ b/db/migrate/20230731140544_fix_samples_measured_signals.rb
@@ -1,0 +1,17 @@
+class FixSamplesMeasuredSignals < ActiveRecord::Migration[5.0]
+  def up
+    # make sure measured signals are always a float (the entity framework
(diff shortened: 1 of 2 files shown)


leandroradusky (approved):
LGTM! :)

PR #1982, opened 2023-08-02
Samples reports generated in the backend

### LINKED ISSUE #1843: Consider the Samples Report PDF on the backend
The samples report PDF in #1840 relies on JavaScript libraries on the frontend, which is different from how we generate PDF elsewhere on the system. Ideally we shouldn't introduce many ways to do the same thing.

The only bothering thing is that the charts are generated on the frontend using d3 (which we already used to generate other charts). Ideally we'd generate them on the backend, but I can't find any good Ruby library to generate SVG. Running d3 on the backend feels weird, and it doesn't work with MiniRacer (I get some missing JS references for d3).

Follow up of #1840

leandroradusky:
Closes #1843 and #1900 

Now, the samples reports are generated in the backend. 

There are some slight changes in the look. The first page looks now like this:

image

Then the other two pages just contain the SVG files as they are downloaded.

image

For the sake of completeness, the ROC plot is also rendered OK, here is how it looks.

image

Some commentaries:

- There are no mockups for the reports and they were not used yet by users, said that the first table with the summary was modified to look like this since it looks better than before.
- In the second section, the confusion matrix, is not looking exactly the same as the webpage due to some limitations of the `prawn-table` library: round borders are not supported and spaces between cells are managed with the borders, and painting borders will fill the full space between cells.
- There was a correction of the error bars, since before the middle bar was a little short, not connecting the horizontal lines properly.
- There was a correction to have all the axes capitalized on the plots.
- The error bars were black with opacity, then they looked gray. This was corrected since they were looking black in the PDF but also they were looking black when opening the SVGs with some software like InkScape. Now, they look the same everywhere.
- The rendering code can be improved, but I want your input on the look of the stuff, if you give me the ok I move forward in improving this code (cc @ysbaddaden @diegoliberman)

Important (cc @julien, I need help in solving this):
**I'm posting the data in a newly created form, but I systematically get an `Invalid Authenticity Token`  error which I don't know how to solve. I'm sure is a silly error but I didn't manage to solve it. Then I had to add the line `skip_before_action :verify_authenticity_token` which is only a transitory solution to make the feature to work.**

I've tested with a lot of different conditions:
- All purposes
- Containing samples without results and not
- Long report names
- Different produced plots. 
- Different browsers. 
- Moving the threshold for the Challenge purpose (which trigger changes in the plots)

All of them worked correctly for me, but this issue needs in-depth testing.

CC @diegoliberman


diegoliberman:
> There are no mockups for the reports and they were not used yet by users, said that the first table with the summary was modified to look like this since it looks better than before.

I don't quite get this.  The design (better than mockups) for the table and reports is [here](https://projects.invisionapp.com/d/main#/console/4103589/470820372/preview) and it had been already implemented.  I'm not sure if it looks better... Anyhow, I say ship it as is, my observation is no blocker at all.

leandroradusky:
Hi @diegoliberman. Sorry because the name was misleading, here we're talking about the PDF exported for the samples report.

The second and third page looks quite the same, the first page now looks like this 

image

As I mentioned above, what I decided to change is the summary, the first section. The Confussion Matrix section now looks slightly different than in the webpage because of the library that we use to generate the PDF report on the backend.

Let's chat if I'm not clarifying with this message :)

ysbaddaden (commented):
That looks really nice! I especially like the matrix for styling the tables. Just a couple comments:

1. We could fix the CSRF error by having a hidden form created by Rails then update the input values when they need to, and have the link submit that hidden form. It would also simplify the JS.

2. I recall we can manually create a table cell for prawn-table so we should be able to put a Bounding Box inside a cell and style it independently from the table (borders, corner radius, background color). Maybe, that's to be verified.

leandroradusky:
Made the fixes to display the confusion matrix exactly as on the website (as for the SVGs, the grays are not handled with transparencies ─as in the project's CSSs─ but with grey colors, which is a limitation of prawn but looks super similar). The styling now is made in subclasses of `Prawn::Table::Cell`.

image

Also, the authenticity token problem was fixed with the `form_tag` tag thanks to the guidance of @ysbaddaden (Thanks!! :bowing_man:)

app/views/samples_reports/show.haml
```diff
@@ -331,5 +330,62 @@
...
+    var specific_svg;
+    if ("#{@purpose}" == "Challenge")
+      specific_svg = encodeURIComponent(getSvg("rocchart", "roc_curve"));
+    else
+      specific_svg = encodeURIComponent(getSvg("linechart", "limit_of_detection"));
+
+    // if purpose is challenge, then we need the live threshold calculations
+    if (#{@purpose.to_json} == "Challenge") {
+      var threshold = document.getElementById("threshold").value;
+      var url = '/samples_reports/update_threshold?context='+context+"&samples_report_id=#{@samples_report.id}&threshold="+threshold;
```
ysbaddaden, line 353: Why do you need to make the calculation on the backend now, just before sending the charts to the backend for rendering? Didn't you already make that request to live update the charts?

app/views/samples_reports/show.haml
```diff
@@ -331,5 +330,62 @@
...
+    else {
+      var params = {
+        id: #{@samples_report.id},
+        measured_signal_svg: measured_signal_svg,
+        specific_svg: specific_svg
+      };
+      var print_url = '/samples_reports/print'; 
+      postObjectToUri(print_url, params, context);
+    }
+  }
```
ysbaddaden, line 379: Couldn't we get all these params right from the actual SamplesReport on the backend? Then we'd only have to send the SVG charts to the backend, and that's it.

If it's the result of the confusion matrix calculation, then it could be re-calculated on the backend from the selected threshold in the `print` action, which avoids the above `update_threshold` request.

app/views/samples_reports/show.haml
```diff
@@ -131,6 +130,10 @@
                       showCloseButton: false,
                       message: "After downloading these files, you will need to fill specific columns. <br />Make sure you read the instructions to be aware of these considerations <br>contained in the file called Instructions.txt")
 
+= form_tag print_samples_reports_path, method: "post", id: "print-pdf-form", target: "_blank" do
+  %input{type: "hidden", name: "json", id: "print-pdf-json"}
+  %input{type: "hidden", name: "context", id: "print-pdf-context"}
```
ysbaddaden, line 135: You can set the current context directly as the input's value (changing context will re-render the page, and update the form accordingly).

app/controllers/samples_reports_controller.rb
```diff
@@ -66,26 +66,48 @@ def create
     end
   end
 
+  def print
+    # convert json param to hash
+    json_params = JSON.parse(params[:json])
```
ysbaddaden, line 71: Can't we send inputs directly instead of going through a `json` param?

app/documents/nih_report.rb
```diff
@@ -0,0 +1,244 @@
...
+        GreyCell.new(@document, [0, 0], content: data[2][2]), 
+        TotalCell.new(@document, [0, 0], content: data[2][3])
+      ],
+      [
+        RotatedCell.new(@document, [0, 0], content: data[3][0]), 
+        TotalCell.new(@document, [0, 0], content: data[3][1]), 
+        TotalCell.new(@document, [0, 0], content: data[3][2]), 
+        TotalCell.new(@document, [0, 0], content: data[3][3])
+      ]
+    ]
```
ysbaddaden, line 149: Oh, I really like those specific cell classes! They really help to simplify.

Hum, maybe we can go even further, and merge `data` into `cells`, and move the `<font>` declaration into `GreyCell` and `TotalCell` that would take distinct kwargs?

```ruby
TotalCell.new(@document, title: "XYZ", value: 12)
```

Note: `point` is also always the same `[0, 0]`, maybe it could be right into `CustomCell`?

ysbaddaden (commented):
Great job @leandroradusky!

Jumping into `prawn` is quite challenging, as it mixes both the semantic + design and it's hard to distinguish them, and you did very nicely.

As always, I'd love to reduce the custom JavaScript to a minimum, and maybe avoiding the `json` param to have just a few inputs (only `threshold` + SVG charts) could help do that. _The dumbest the frontend is, the better._

leandroradusky:
I think I've addressed all the commentaries, if not, now let's sync :) 

Thanks a lot for the guidance !!

app/views/samples_reports/show.haml
```diff
@@ -340,49 +347,25 @@
...
+    };
+
+    document.getElementById("print-pdf-measured-signal-svg").value = encodeURIComponent(getSvg("barchart", "measured_signal"));
+    
     if ("#{@purpose}" == "Challenge")
-      specific_svg = encodeURIComponent(getSvg("rocchart", "roc_curve"));
+      document.getElementById("print-pdf-specific-svg").value = encodeURIComponent(getSvg("rocchart", "roc_curve"));
     else
-      specific_svg = encodeURIComponent(getSvg("linechart", "limit_of_detection"));
+      document.getElementById("print-pdf-specific-svg").value = encodeURIComponent(getSvg("linechart", "roc_curve"));
```
ysbaddaden, line 360:
```suggestion
      document.getElementById("print-pdf-specific-svg").value = encodeURIComponent(getSvg("linechart", "limit_of_detection"));
```

ysbaddaden: I'm also pretty sure you don't need to `encodeURIcomponent` the SVG. The browser will do that by itself (same on the backend: rails will decode automatically).

PR #1983, opened 2023-08-03
Added column replicates in CSV box creation

### LINKED ISSUE #1972: When uploading samples with a CSV, the replicate number should be taken as an input
If we don't allow to enter the replicate number, and if they input a CSV with samples from the same batch and the same concentration, all of them remain with the same replicate number --> 1

I think that a column in the CSV should be "replicates".  If the user enters "4", then the system will create 4 replicates for that row.  
An alternative would be that the replicate column has the actual number so the user has full control.  I'm not sure what would be more comfortable for them.  Maybe we could ask AlexC?

leandroradusky:
Closes #1972.

When creating a box from a CSV file, a new column asking for the number of `Replicates` was added, and the samples will be created with the according replicate number (before all the samples had `replicate=1`).

Two errors were fixed: 

- Error messages were not correctly shown in the `rowContent`, this was reported by Muhammad but not corrected, not sure why but no issue was created. It was due to an inversed condition. 
- The first row of the uploaded CSV was always ignored. The specs were written but the number of expected samples was one less than the ones in the CSV files used for testing.

### DIFF
--- a/app/controllers/boxes_controller.rb
+++ b/app/controllers/boxes_controller.rb
@@ -80,7 +80,6 @@ class BoxesController < ApplicationController
 
     @found_batches = @navigation_context.institution.batches.where(batch_number: batch_numbers.to_a).pluck(:batch_number)
     @not_found_batches = (batch_numbers - @found_batches)
-
   end
 
   def create
--- a/app/controllers/boxes_controller.rb
+++ b/app/controllers/boxes_controller.rb
@@ -57,25 +57,30 @@ class BoxesController < ApplicationController
 
     # TODO: should be handled by BoxForm (& duplicates BoxForm#parse_csv)
     CSV.open(params[:csv_box].path, headers: true) do |csv|
-      while csv.headers == true
-        csv.readline
-      end
-
-      unless csv.headers == ["Batch", "Concentration", "Distractor", "Instructions"]
-        @error_message = "Invalid columns"
-        return # rubocop:disable Lint/NonLocalExitFromIterator
-      end
-
+      unique_rows = Set.new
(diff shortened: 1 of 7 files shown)


app/controllers/boxes_controller.rb
```diff
@@ -57,19 +57,23 @@ def validate
 
     # TODO: should be handled by BoxForm (& duplicates BoxForm#parse_csv)
     CSV.open(params[:csv_box].path, headers: true) do |csv|
-      while csv.headers == true
-        csv.readline
-      end
```
ysbaddaden, line 62: I understand this is related to the fix, but in my tests `CSV.open` would merely open the file and `csv.headers` would be `true` until I read at least one line, in which case it became `["Batch", ...]` :confused:

app/models/box_form.rb
```diff
@@ -217,7 +217,7 @@ def parse_csv(path)
           instruction: row["Instructions"],
           concentrations: {
             "0" => {
-              replicate: 1,
+              replicate: row["Replicates"],
```
ysbaddaden, line 220:
```suggestion
              replicate: row["Replicates"].to_i,
```

ysbaddaden (approved):

app/controllers/boxes_controller.rb
leandroradusky: The `csc.each do ...` Actually reads a line and solve the problem... It's failing in your local?

ysbaddaden: No no, but that's very weird :/

ysbaddaden (approved):

PR #1992, opened 2023-09-14
Sample boxes in inventory & labels arranged by id

leandroradusky:
In response to client requirements, the ordering of the samples used in the inventory and the labels is now by sample's id.

### DIFF
--- a/app/controllers/boxes_controller.rb
+++ b/app/controllers/boxes_controller.rb
@@ -146,7 +146,7 @@ class BoxesController < ApplicationController
     samples = if @box.blinded? && !params[:unblind]
       samples.scrambled
     else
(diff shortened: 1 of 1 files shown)


app/controllers/boxes_controller.rb
```diff
@@ -146,7 +146,7 @@ def load_box_samples
     samples = if @box.blinded? && !params[:unblind]
       samples.scrambled
     else
-      samples.sort_by{ |sample|  [ sample.batch_number.to_s, sample.concentration , sample.replicate ] }
+      samples.sort_by{ |sample|  [ sample.id ] }
```
ysbaddaden, line 149: We don't need to load the samples to sort them anymore (no more entity framework field):

```suggestion
      samples.order(:id)
```

Issue #1990, opened 2023-08-28
New table to export for NIH


leandroradusky:
A fourth table has to be included within the zip file when `Export NIH tables` button is pressed. 

The fields are those included in [this](https://docs.google.com/spreadsheets/d/13eew7mJOp0fbh_hs8SbUGHtnZvZ7giqp/edit#gid=1065749321) spreadsheet, but only for the batches included in the box/panel that is being reported in the report where the NIH tables are exported.

Issue #1988, opened 2023-08-28
Migrate new fields for batches entities (v2)


leandroradusky:
After creating the new fields for batches entities (#1987), make a one time migration taking the new fields' values for batches from the given spreadsheet https://docs.google.com/spreadsheets/d/13eew7mJOp0fbh_hs8SbUGHtnZvZ7giqp/edit#gid=1065749321.

leandroradusky:
This could include migrations for the new fields added in v0.23.0 which aren't deployed up to this date.

leandroradusky:
Also, update the `original_batch_id` field of sampled based on the `old_batch_number` previously defined core field.

PR #1996, opened 2023-10-23
New fields for batches entities, new table for NIH

### LINKED ISSUE #1987: New fields for batches entities (v2)
The fields in batches entities should include all missing fields in [this](https://docs.google.com/spreadsheets/d/13eew7mJOp0fbh_hs8SbUGHtnZvZ7giqp/edit#gid=1065749321) table, since these properties will be exported in the NIH tables.

In my recollection these fields are:

- `gisaid_clade`
- `virus_sample_source`
- `virus_sample_source_url`
- `virus_source`
- `virus_location`
- `virus_sample_type`
- `virus_sample_formulation`
- `virus_sample_concentration`
- `virus_sample_concentration_unit`
- `virus_sample_genome_equivalents`
- `virus_sample_genome_equivalents_unit`
- `virus_sample_genome_equivalents_reference_gene`
- `virus_preinactivation_tcid50`
- `virus_preinactivation_tcid50_unit`
- `virus_sample_grow_cell_line`

Please note that these are the field numbers expressed in the table, when added into the batches entity, they shouldn't have the `virus_` prefix on their name.

leandroradusky:
Closes #1987 and #1990.

- Creates the new fields required for `Batches` entity.
- Remove the new fields previously copied to samples in the `Samples` entity (#1863).
- Create a new `original_batch_id` field for `Samples` to recover information of the original batch upon transfer.
- Modifies the existing NIH tables accordingly and creates the new required batches table for the NIH export feature.
- Specs were modified accordingly (let me know if something is missing with respect to that)

**QA notes:**
- For the correct migration of data, I will modify accordingly (#1988), then, while testing this PR, I recommend to use fresh batches/boxes/samples, existing ones could give raise to inconsistencies.


leandroradusky:
It's ready for your review @ysbaddaden :)

There is one spec systematically failing which is completely unrelated to this development :disappointed: : the `blender_spec` expects two institutions but there are seven, which makes no sense since in the setup there are only two. I will investigate what is happening but that shouldn't block the review :)

app/models/box.rb
```diff
@@ -108,4 +108,12 @@ def unblind!
...
     update_columns(blinded: true)
   end
+
+  def original_batches
+    Batch.where(id: samples.map(&:original_batch_id).uniq)
+  end
+
+  def batches
+    batches + original_batches
+  end
```
ysbaddaden, line 118: **question**: isn't this calling the `batches` method recursively (instead of calling the association)?

You might want to name it `all_batches` or maybe this isn't needed anymore?

app/models/sample.rb
```diff
@@ -187,4 +183,8 @@ def self.sort_columns
   def self.sort_column?(attr_name)
     sort_columns.include?(attr_name)
   end
+
+  def original_batch
+    Batch.find_by_id(original_batch_id)
+  end
```
ysbaddaden, line 189: **suggestion**: this is a relation, so you could declare it as such in the model. then we can leverage activerecord to generate efficient SQL queries (e.g. `report.samples.preload(:original_batch)`):

```ruby
belongs_to :original_batch, class_name: "Batch", optional: true
```

app/views/samples/_form.haml
```diff
@@ -65,22 +65,6 @@
...
-          = autocomplete_field(f, "sample[target_organism_taxonomy_id]", :target_organism_taxonomy_id, @navigation_context.institution, class: "input-x-large")
-
-      = f.form_field :pango_lineage do
-        - @sample_form.blinded_attribute(:pango_lineage) do
-          = autocomplete_field(f, "sample[pango_lineage]", :pango_lineage, @navigation_context.institution, class: "input-x-large")
-
-      = f.form_field :who_label do
-        - @sample_form.blinded_attribute(:who_label) do
-          = autocomplete_field(f, "sample[who_label]", :who_label, @navigation_context.institution, class: "input-x-large")
-
```
ysbaddaden, line 83: **question**: institutions that received samples won't be able to view these fields anymore (they can't access the batches), except from the NIH tables. Is it expected?

db/schema.rb
```diff
@@ -128,23 +128,38 @@
...
+  create_table "autocomplete_values", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_520_ci" do |t|
+    t.string   "field_name"
+    t.string   "value"
+    t.integer  "institution_id", null: false
+    t.datetime "created_at",     null: false
+    t.datetime "updated_at",     null: false
+    t.index ["field_name", "value", "institution_id"], name: "autocomplete_index", unique: true, using: :btree
+    t.index ["institution_id"], name: "index_autocomplete_values_on_institution_id", using: :btree
+  end
+
```
ysbaddaden, line 140: **note**: I think this table doesn't exist anymore (but we might want to reintroduce it to speed up the autocomplete queries).

```suggestion
```

spec/controllers/transfer_packages_controller_spec.rb
```diff
@@ -331,6 +331,31 @@
...
+                            },
+                          },
+                        },
+                      }
+      end.to change { TransferPackage.count }.by(1)
+
+      sample.reload
+      expect(sample.batch_id).to be_nil
+      expect(sample.original_batch_id).to eq(batch.id)
+    end
```
ysbaddaden, line 357: **praise**: nice!

ysbaddaden (commented):
Hey Leandro, I left a bunch of suggestions to speed things up, by adding actual relations between models, to reduce the number of SQL queries.

I'm also concerned about a change in the autocompleter.

app/controllers/autocompletes_controller.rb
```diff
@@ -22,11 +47,6 @@ def unique_values_for(field_name, query)
       values << value if value&.match?(matcher)
     end
 
-    check_access(@institution.samples, READ_SAMPLE).select(:id, :core_fields, :custom_fields).find_each do |sample|
-      value = sample.send(field_name)
-      values << value if value&.match?(matcher)
-    end
```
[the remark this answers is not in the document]
leandroradusky: Respect to this comment: the only thing that I'm doing here is removing the query on samples and just keeping the query on batches because now (after "reverting" the copy of attributes in samples in #1863) all the autocomplete values belong only to batches. Then this security issue was there before. 

Said that, I will restrict the autocomplete to the institution anyway, mostly because of the performance issue. Leaking data about other institution samples is not super sensitive but is also a plus, so let's go with it :)

app/models/box.rb
leandroradusky: This is dead code, we're using now only `Sample#original_batch` and `SamplesReport#original_batches` after applying the previous comment :)

app/models/box.rb
```diff
@@ -108,4 +108,12 @@ def unblind!
   def blind!
     update_columns(blinded: true)
   end
+
+  def original_batches
+    Batch.where(id: samples.map(&:original_batch_id).uniq)
```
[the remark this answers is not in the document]
leandroradusky: Don't needed anymore.

app/models/sample.rb
leandroradusky: Solved in a previous comment :)

app/views/samples/_form.haml
leandroradusky: These fields for samples never arrived to production, so there will be no impact in removing this.

db/schema.rb
leandroradusky: Given that now autocomplete apply only to batches fields and the number of batches that each institution will have (in the order of few dozens, not like samples that there are a lot, hundreds or thousands) we don't want to reintroduce this table.

app/controllers/autocompletes_controller.rb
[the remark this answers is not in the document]
ysbaddaden: Oh, right. Forget I said anything.

ysbaddaden (approved):

PR #1997, opened 2023-10-30
Migrations needed for modifications on the new batches NIH table

### LINKED ISSUE #1988: Migrate new fields for batches entities (v2)

leandroradusky:
Close #1988 .

Migrations needed for the new NIH table:
- Updating of the `original_batch_id` field. Since the migration is not applied nor in staging neither in production I just modified the existing migration from the previous PR (which just added the field). Is that correct?
- Update of the rake task that was there to upload the information for the four previously added fields into a task that accepts a CSV with all the new fields. This will have to be run once, and the client owe us the good file to make the migration.


db/migrate/20231030102203_add_original_batch_id_to_samples.rb
```diff
@@ -1,5 +1,13 @@
 class AddOriginalBatchIdToSamples < ActiveRecord::Migration
   def change
     add_column :samples, :original_batch_id, :integer
+
+    Sample.preload(:batch).find_each do |sample|
+      batch = sample.batch || Batch.find_by(batch_number: sample.old_batch_number)
```
ysbaddaden, line 6: **suggestion**: maybe search the old batch number through the box? so we'd be 100% sure that the batch is really the original batch, just in case.

ysbaddaden (approved):

